Issue #8, opened 2017-02-04
Implement __FILE__ and __LINE__ macros


MikePopoloski:
The intrinsic __FILE__ and __LINE__ macros are currently recognized but generate empty tokens ("" and 0 respectively). The __LINE__ macro in particular will require making use of the source manager infrastructure for figuring out which line we're looking at.

Note that this plays into the `line directive as well, which changes the "apparent" line number (like #line in C++).

drsam94:
This issue was closed by commit 0649d61

PR #24, opened 2017-02-06
Added support for __LINE__ and __FILE__ preprocessor directives

drsam94:
This addes support for the __FILE__ and __LINE__ preprocessor directives, and unittests testing that they are supported properly. There is a case this does not support, which I mentioned in a TODO, which is definitely supported by the C preprocessor, but I don't know whether we need to support it here:

In something like
~~~~~~~~~~~
`define PASTE(x, y) x `` y
`define PASTELINE(x) PASTE(x, `__LINE__)
PASTELINE(foo)
~~~~~~~~~~~

We encounter an error because we don't look to expand macro arguments before expanding the outer macro, and pasting with __LINE__ and __FILE__ are not supported. Other than that, these seem to be implemented fine now, and I added corresponding helper methods to SourceManager which make it easy to compute the relevant values for these. This commit does not add support for `line, but it should be a reasonably easy next step from here. This pull request resolves Issue #8


build/linux/Makefile
```diff
@@ -1,5 +1,5 @@
 TOP:=../..
-CXX?=clang++-3.9
+CXX=clang++-3.8
```
Baranowski: Could we keep this unchanged or create build/mac/Makefile?

include/SourceManager.h
```diff
@@ -53,9 +53,11 @@ class SourceManager {
     void addUserDirectory(StringRef path);
 
     /// Gets the source line number for a given source location.
-    /// @a location must be a file location.
     uint32_t getLineNumber(SourceLocation location);
 
+    /// Gets the source file number for a given source location
```
MikePopoloski: Fix comment here.

source/SourceManager.cpp
```diff
@@ -32,7 +32,16 @@ void SourceManager::addUserDirectory(StringRef path) {
     userDirectories.push_back(Path::makeAbsolute(path));
 }
 
+SourceLocation SourceManager::getFirstFileLocation(SourceLocation location) {
+    while (location.buffer() && std::get_if<FileInfo>(&bufferEntries[location.buffer().id]) == nullptr) {
```
MikePopoloski: could use IsMacroLoc() here to make it more obvious what is happening.

source/Preprocessor.cpp
```diff
@@ -523,18 +525,19 @@ MacroActualArgumentListSyntax* Preprocessor::handleTopLevelMacro(Token directive
...
-            info->extra = "";
+            info->extra = sourceManager.getFileName(token.location());
             token = Token(TokenKind::StringLiteral, info);
-        }
-        else if (token.kind == TokenKind::IntrinsicLineMacro) {
+        } else if (token.kind == TokenKind::IntrinsicLineMacro) {
             auto info = alloc.emplace<Token::Info>(token.trivia(), "", token.location(), 0);
-            info->setNumInfo(SVInt(0));
+            uint64_t lineNum = sourceManager.getLineNumber(token.location());
+            // sourceManager gives a 0-indexed line number
```
MikePopoloski: Is this comment correct? If so then DiagnosticWriter is probably broken.

build/linux/Makefile
MikePopoloski: The ?= means that we can set CXX as a command line argument for the make process... maybe that's preferable to trying to get one value that works everywhere.

source/Preprocessor.cpp
drsam94: I am pretty sure it is correct, i.e I very consistently got correct values for __LINE__ that were one less than the correct value, including a value of 0 if it was on the first line. I didn't dig into DiagnosticWriter to see if that made sense, but it is likely that everything would be better if the "+1" were in getLineNumber

source/SourceManager.cpp
drsam94: Sounds good, didn't see that helper there and got a bit creative

include/SourceManager.h
drsam94: will do

build/linux/Makefile
drsam94: Ah, so that's what the ? means...I just found that the build didn't work if I left the ? there. If it will work with me just manually passing a flag instead of changing this for everyone, that seems like the clearly better way to go.

drsam94:
To record the results of experiments, In the code snippet I mention in the description, other SV preprocessors all make the `__LINE__` disappear instead of actually pasting it. Also, other SV preprocessors don't like the space around the pasting ``, but whatever

PR #25, opened 2017-02-06
Parser: throw exception when recursing too deep

Baranowski:
Parser::setDepthGuard() creates a DepthGuard object. Parser::depth keeps
count of all the DepthGuard objects currently alive. Throws a
runtime_error whne it exceeds MaxDepth.


include/Parser.h
```diff
@@ -20,6 +20,8 @@ class Preprocessor;
 /// Implements a full syntax parser for SystemVerilog.
 class Parser : ParserBase {
 public:
+    static constexpr size_t MaxDepth=100;
```
MikePopoloski: Could you add a TODO that we should pull this value from an Options system at some point (which doesn't currently exist)? That way things like this can be configured from the command line.

PR #28, opened 2017-02-07
Make with lldb

denisrr:
tested with:
CC=clang-4.0
CXX=clang++-4.0
AR=llvm-ar-4.0

### DIFF
--- a/build/linux/Makefile
+++ b/build/linux/Makefile
@@ -1,8 +1,10 @@
 TOP:=../..
+
 CXX?=clang++-3.9
-CXXFLAGS:=-std=c++1z -MMD -MP -g
+CXXFLAGS?=-std=c++1z -MMD -MP -g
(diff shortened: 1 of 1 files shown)


build/linux/Makefile
```diff
@@ -1,8 +1,10 @@
 TOP:=../..
+
 CXX?=clang++-3.9
-CXXFLAGS:=-std=c++1z -MMD -MP -g
+CXXFLAGS?=-std=c++1z -MMD -MP -g
 INCLUDES:=-I$(TOP)/external -I$(TOP)/include -I./
-ARFLAGS:=-rs
+AR?=llvm-ar-3.9
```
MikePopoloski: Rather than rely on a specific version of AR for everybody, maybe just set that in your environment and just make AR?=ar here.

denisrr: did not intend to remove ARFLAGS

Issue #4, opened 2017-02-04
Handle remaining preprocessor directives


MikePopoloski:
There are a few preprocessor directives that aren't yet handled. Implement them in Preprocessor::next().

drsam94:
This is done other than handling of any pragmas

PR #30, opened 2017-02-07
Added support for line directive

drsam94:
This change adds support for the 'line directive, and tests it's compatibility with respect to the `__LINE__` and `__FILE__` macros, as well as in diagnostic reporting

### DIFF
--- a/include/SourceManager.h
+++ b/include/SourceManager.h
@@ -117,7 +117,7 @@ private:
         std::string name;                   // name of the file
         std::vector<uint32_t> lineOffsets;  // char offset for each line
         struct LineDirectiveInfo {
-            uint32_t lineInFile; // Actual file line where direcitve occurred
+            uint32_t lineInFile; // Actual file line where directive occurred
             uint32_t lineOfDirective; // Line number set by directive
             std::string name; // File name set by directive
             uint8_t level; // level of directive
--- a/include/SourceManager.h
+++ b/include/SourceManager.h
@@ -62,10 +62,6 @@ public:
     /// @a location must be a file location.
     uint32_t getColumnNumber(SourceLocation location);
 
-    /// Gets the file name of the given buffer.
-    /// If @a buffer does not represent a file, the result will be an empty string.
-    StringRef getBufferName(BufferID buffer);
-
     /// Gets a location that indicates from where the given buffer was included.
     /// @a location must be a file location.
     SourceLocation getIncludedFrom(BufferID buffer);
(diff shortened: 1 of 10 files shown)


drsam94:
This is work towards issue #4

include/SourceManager.h
```diff
@@ -117,6 +116,28 @@ class SourceManager {
         Vector<char> mem;                   // file contents
         std::string name;                   // name of the file
         std::vector<uint32_t> lineOffsets;  // char offset for each line
+        struct LineDirectiveInfo {
+            uint32_t lineInFile; // Actual file line where direcitve occurred
```
MikePopoloski: "directive" typo in comment

source/Diagnostics.cpp
```diff
@@ -100,6 +100,7 @@ DiagnosticWriter::DiagnosticWriter(SourceManager& sourceManager) :
     descriptors[DiagCode::InvalidMacroName] = { "can't redefine compiler directive as a macro", DiagnosticSeverity::Error };
     descriptors[DiagCode::TooManyActualMacroArgs] = { "too many arguments provided to function-like macro", DiagnosticSeverity::Error };
     descriptors[DiagCode::NotEnoughMacroArgs] = { "not enough arguments provided to function-like macro", DiagnosticSeverity::Error };
+    descriptors[DiagCode::InvalidLineDirectiveLevel] = { "Invalid level for `line directive, must be 0, 1, or 2", DiagnosticSeverity::Error };
```
MikePopoloski: Error messages should start with lower case character.

source/Preprocessor.cpp
```diff
@@ -417,6 +418,23 @@ Trivia Preprocessor::handleTimescaleDirective(Token directive) {
     return Trivia(TriviaKind::Directive, timescale);
 }
 
+Trivia Preprocessor::handleLineDirective(Token directive) {
+    Token lineNumber = expect(TokenKind::IntegerLiteral);
+    Token fileName = expect(TokenKind::StringLiteral);
+    Token level = expect(TokenKind::IntegerLiteral);
+    const SVInt& levNum = std::get<SVInt>(level.numericValue());
+    if (!(levNum == 0 || levNum == 1 || levNum == 2)) {
```
MikePopoloski: If any of lineNumber, fileName, or level are missing we should probably return instead of calling addLineDirective (we should still create the LineDirectiveSyntax though). The expect() call will generate a token with the Missing flag set if it couldn't find the thing we wanted.

drsam94: Good point, otherwise you could probably end up with some really weird error messages if you had errors after a malformed `line. Will make both this change and the other minor changes.

PR #31, opened 2017-02-07
added mykind to symbol, replaced parameters with scope in module

denisrr:


tests/file_tests/corpus/include/module1.sv
```diff
@@ -1,3 +1,17 @@
+module helper
```
drsam94: I think it would make sense to make this "module2.sv" and leave the old "module1.sv" as it was -- in general with test cases for a compiler, I think "the more the merrier" is always the case

include/Scope.h
```diff
@@ -1,36 +1,42 @@
...
+    const Symbol* getNth(const SymbolKind& kind, size_t index) const;
 
     const Scope* parent() const { return parentScope; }
 
     static const Scope* const Empty;
 
 private:
     std::unordered_map<StringRef, const Symbol*> table;
+    std::vector<const Symbol*> list;
     Scope* parentScope = nullptr;
```
drsam94: shouldn't parentScope default to Scope::Empty, not nullptr?

include/Symbol.h
```diff
@@ -37,7 +38,11 @@ class Symbol {
         kind(kind), name(name), location(location) {}
 
     template<typename T>
-    const T& as() const { return *static_cast<const T*>(this); }
+    const T& as() const {
+        // TODO: if (T::mykind != this->kind) return nullptr;
```
drsam94: It looks like the main issue with this todo is that this returns a const& instead of a pointer, so you can't return something like nullptr. Do we have a reasonable solution in mind?

source/Scope.cpp
```diff
@@ -4,18 +4,32 @@
...
+        auto it = table.find(name);
+        if (it != table.end()) return it->second;
+        if (local) return nullptr;
+    }
+    return nullptr;
 }
 
-}
\ No newline at end of file
+const Symbol* Scope::getNth(const SymbolKind& kind, size_t index) const {
```
drsam94: Is there some sort of guarantee of ordering so we can actually know what this will do? I understand the normal lookups, I don't really understand the idea behind this.

tests/file_tests/corpus/include/module1.sv
denisrr: good point

include/Scope.h
denisrr: yes

include/Symbol.h
denisrr: we can refactor as() to return a pointer

source/Scope.cpp
denisrr: yes, I added a list to Scope to preserve the order of declarations

include/Scope.h
```diff
@@ -1,36 +1,44 @@
...
 public:
-    Scope() {};
+    Scope();
 
-    void add(const Symbol* symbol);
+    bool add(const Symbol* symbol);
 
     template<typename TContainer>
-    void addRange(const TContainer& container) {
+    bool addRange(const TContainer& container) {
```
MikePopoloski: This is fine for now, but it's hard to know what to do if we get a false return from this function because we don't know which symbols were added and which weren't.

include/SemanticModel.h
```diff
@@ -91,7 +88,7 @@ class SemanticModel {
     void makeAttributes(SmallVector<const AttributeSymbol*>& results, const SyntaxList<AttributeInstanceSyntax>& attributes);
 
     const ModuleSymbol* makeModule(const ModuleDeclarationSyntax* syntax, ArrayRef<const ParameterSymbol*> parameters);
-    void handleInstantiation(const HierarchyInstantiationSyntax* syntax, SmallVector<const Symbol*>& results);
+    void handleInstantiation(const Scope* instantiationScope, const HierarchyInstantiationSyntax* syntax, SmallVector<const Symbol*>& results);
```
MikePopoloski: Could you put the scope parameter at the end like all the other functions? Just a nitpick on consistency.

source/Scope.cpp
```diff
@@ -4,18 +4,42 @@
 namespace slang {
 
 static const Scope emptyScope;
-const Scope* const Scope::Empty = &emptyScope;
+const Scope *const Scope::Empty = &emptyScope;
```
MikePopoloski: Not sure why you moved the pointer symbol. The style in the entire codebase is to put the * and & symbols on the typename, since they're logically part of the type.

source/Scope.cpp
```diff
@@ -4,18 +4,42 @@
...
 
-void Scope::add(const Symbol* symbol) {
-    ASSERT(symbol);
-    table.emplace(symbol->name, symbol);
+Scope::Scope() : parentScope{Scope::Empty} {}
+
+bool Scope::add(const Symbol *symbol) {
+        ASSERT(symbol);
```
MikePopoloski: Weird indentation here.

source/Scope.cpp
```diff
@@ -4,18 +4,42 @@
...
+    list.emplace_back(symbol);
+    return table.emplace(symbol->name, symbol).second;
+}
+
+const Symbol *Scope::lookup(StringRef name, bool local) const {
+    for (auto scope = this; scope && scope != Scope::Empty; scope = scope->parent()) {
```
MikePopoloski: I think we shouldn't require all scopes to point to Scope::Empty... it should be fine for the parent pointer to just be null (which is the default when a scope is first constructed), in which case we don't need any special knowledge of Scope::Empty.

source/Scope.cpp
```diff
@@ -4,18 +4,42 @@
...
+    }
+    return nullptr;
 }
 
-}
\ No newline at end of file
+std::string Scope::toString() const {
+    std::string result;
+    for (const Symbol* sym : list)
+        result += std::string("[") + sym->name.toString() + std::string("[");
```
MikePopoloski: The second "[" should be "]" right?

tests/file_tests/main.cpp
```diff
@@ -31,11 +34,25 @@ int main() {
         files++;
 
         trees.emplace_back(std::move(tree));
+        table.addSyntaxTree(&tree);
```
MikePopoloski: Be careful here; you're adding a pointer to a temporary variable. It's only luck that it's working.

tests/unit_tests/ModuleBindingTests.cpp
```diff
@@ -118,7 +118,9 @@ endmodule
 
     // Check that the tree of children has been instantiated correctly
     const auto& leaf = instance->getChild<InstanceSymbol>(0).getChild<InstanceSymbol>(0);
-    CHECK(std::get<SVInt>(leaf.module->parameters[0]->value) == 4);
+    const auto& foo = leaf.module->scope->lookup(StringRef(std::string("foo")))->as<ParameterSymbol>();
+
+    CHECK(std::get<SVInt>(foo.value) == 4);
```
MikePopoloski: You don't need the get<SVInt> here, you can just do foo.value.integer().

tests/unit_tests/ModuleBindingTests.cpp
```diff
@@ -118,7 +118,9 @@ endmodule
 
     // Check that the tree of children has been instantiated correctly
     const auto& leaf = instance->getChild<InstanceSymbol>(0).getChild<InstanceSymbol>(0);
-    CHECK(std::get<SVInt>(leaf.module->parameters[0]->value) == 4);
+    const auto& foo = leaf.module->scope->lookup(StringRef(std::string("foo")))->as<ParameterSymbol>();
```
MikePopoloski: No need to wrap "foo" in std::string, it should work if you just pass that to StringRef. In fact it's an implicit conversion, so you should be able to just do lookup("foo",

tests/unit_tests/ModuleBindingTests.cpp
```diff
@@ -157,8 +159,9 @@ endmodule
     const auto& leaf = instance->getChild<InstanceSymbol>(0)
         .getChild<GenerateBlock>(0)
         .getChild<InstanceSymbol>(0);
+    const auto& foo = leaf.module->scope->lookup(StringRef(std::string("foo")))->as<ParameterSymbol>();
```
MikePopoloski: Same here.

tests/unit_tests/ModuleBindingTests.cpp
```diff
@@ -157,8 +159,9 @@ endmodule
...
 
-    CHECK(std::get<SVInt>(leaf.module->parameters[0]->value) == 1);
+    CHECK(std::get<SVInt>(foo.value) == 1);
```
MikePopoloski: std::get not needed here either.

include/Scope.h
denisrr: ok i can make it check for dupes before it does any changes, so that none would be added

include/SemanticModel.h
denisrr: ok

source/Scope.cpp
denisrr: IDE ate my pointer

source/Scope.cpp
denisrr: fixed

source/Scope.cpp
denisrr: so Scope::Empty is just an optimization instead of default-constructed Scope() ?

source/Scope.cpp
denisrr: yes

tests/file_tests/main.cpp
denisrr: fixed

tests/unit_tests/ModuleBindingTests.cpp
denisrr: fixed

tests/unit_tests/ModuleBindingTests.cpp
denisrr: fixed

PR #32, opened 2017-02-07
Implemented remaining preprocessor directives

drsam94:
This implements the remaining preprocessor directives, undef, undefineall, begin_keywords, and end_keywords. The only totally unsupported directive at this point is pragma.


include/Token.h
```diff
@@ -182,7 +182,18 @@ class Token {
 
 TokenKind getSystemKeywordKind(StringRef text);
 StringRef getTokenKindText(TokenKind kind);
-const StringTable<TokenKind>* getKeywordTable();
+// Different restricted sets of keywords that can b
```
MikePopoloski: Comment seems truncated.

source/Preprocessor.cpp
```diff
@@ -472,6 +474,68 @@ Trivia Preprocessor::handleDefaultNetTypeDirective(Token directive) {
     return Trivia(TriviaKind::Directive, result);
 }
 
+Trivia Preprocessor::handleUndefDirective(Token directive) {
+    Token nameToken = expect(TokenKind::Identifier);
+    StringRef name = nameToken.valueText();
```
MikePopoloski: If name is empty (because nameToken was missing) we shouldn't also error about not finding the directive.

source/Preprocessor.cpp
```diff
@@ -472,6 +474,68 @@ Trivia Preprocessor::handleDefaultNetTypeDirective(Token directive) {
...
+Trivia Preprocessor::handleUndefineAllDirective(Token directive) {
+    macros.clear();
+    macros["__FILE__"] = &fileDirective;
+    macros["__LINE__"] = &lineDirective;
+    return createSimpleDirective(directive);
+}
+
+Trivia Preprocessor::handleBeginKeywordsDirective(Token directive) {
+    Token versionToken = expect(TokenKind::StringLiteral);
+    StringRef verStr = versionToken.valueText();
```
MikePopoloski: Same thing here.

source/Preprocessor.cpp
drsam94: good point , that would lead to confusing error messages. Will fix in both places.

PR #34, opened 2017-02-08
Make SyntaxTree easier to use

MikePopoloski:
Make SyntaxTree (and to a lesser extent SemanticModel) easier to use when you don't care about doing a full compilation of multiple full source files.

SyntaxTree now has fromText overloads that let you parse some specific kind of syntax directly.


MikePopoloski:
We can ignore the build failure; it's Windows only. I'll take a look at it when I get home.

include/SemanticModel.h
```diff
@@ -110,6 +114,10 @@ class SemanticModel {
 
     // cache of simple integral types keyed by {width, signedness, 4-state, isReg}
     std::unordered_map<uint64_t, const TypeSymbol*> integralTypeCache;
+
+    // An empty declaration table to make it easier to create SemanticModels
+    // where we don't care about the full compilation process.
```
drsam94: It seems like slang is in general super inconsistent about using doxygen comments, but the fact that you added two comments in this change, one // and one ///, makes it especially noticeable.

include/SyntaxTree.h
```diff
@@ -30,23 +34,81 @@ class SyntaxTree {
...
+        return fromFile(path, getDefaultSourceManager());
+    }
 
-    /// Gets the root of the syntax tree.
-    const CompilationUnitSyntax* root() const { return rootNode; }
+    static SyntaxTree fromText(StringRef text) {
+        return fromText(text, getDefaultSourceManager());
+    }
+
+    static SyntaxTree fromFile(StringRef path, SourceManager& sourceManager) {
```
drsam94: factory methods that return values rather than pointers seem a bit strange to me, but if it's working fine so far, there is nothing wrong with it.

include/SemanticModel.h
MikePopoloski: I typically use doxygen comments for public-facing things. Private variables and functions are just normal comments. You're right though that I'm not the most consistent about it.

include/SyntaxTree.h
MikePopoloski: No need to heap allocate if it's not necessary; I'm a big fan of move semantics.

Issue #14, opened 2017-02-04
Start of line check for directives


MikePopoloski:
Most preprocessor directives require that they start on their own line. Currently this isn't checked or enforced.

PR #35, opened 2017-02-08
Enforced that compiler directives begin on their own line

drsam94:
This addresses Issue #14

### DIFF
--- a/source/Lexer.cpp
+++ b/source/Lexer.cpp
@@ -648,11 +648,13 @@ TokenKind Lexer::lexDollarSign(Token::Info* info) {
 }
 
 TokenKind Lexer::lexDirective(Token::Info* info) {
+    // store the offset before scanning in order to easily report error locations
+    uint32_t startingOffset = currentOffset();
     scanIdentifier();
 
     // if length is 1, we just have a grave character on its own, which is an error
     if (lexemeLength() == 1) {
-        addError(DiagCode::MisplacedDirectiveChar, currentOffset() - 1);
+        addError(DiagCode::MisplacedDirectiveChar, startingOffset);
         info->extra = SyntaxKind::Unknown;
         return TokenKind::Directive;
     }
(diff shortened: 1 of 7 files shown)


source/Preprocessor.cpp
```diff
@@ -82,9 +82,12 @@ Token Preprocessor::next(LexerMode mode) {
             case SyntaxKind::NoUnconnectedDriveDirective:
             case SyntaxKind::CellDefineDirective:
             case SyntaxKind::EndCellDefineDirective:
-            default:
                 trivia.append(createSimpleDirective(token));
                 break;
+            default:
```
MikePopoloski: There is a DEFAULT_UNREACHABLE macro to assert when you hit the default case, if this really shouldn't ever be possible.

source/Lexer.cpp
```diff
@@ -656,6 +658,10 @@ TokenKind Lexer::lexDirective(Token::Info* info) {
     }
 
     info->extra = getDirectiveKind(lexeme().subString(1));
+    if (!onNewLine && std::get<SyntaxKind>(info->extra) != SyntaxKind::MacroUsage) {
+        // All directives other than a macro usage must be on their own line
+        addError(DiagCode::DirectiveNotFirstOnLine, currentOffset() - 1);
```
MikePopoloski: currentOffset() here points to wherever the lexer is looking, so it's at the end of the directive. We probably would like to point the error at the beginning of the directive token (so just store the result of currentOffset() before we advance the pointer above in scanIdentifier()).

drsam94: Ah ok. I used this value because we used it in the stray grave case, but that one got to use the knowledge that currentOffset() - 1 = start. Will change.

source/Preprocessor.cpp
drsam94: awesome, that seems a lot better than ASSERT(false) ing

PR #38, opened 2017-02-09
Allow macros to use keywords as formal argument names

evanbryantmiller:


source/Preprocessor.cpp
```diff
@@ -844,7 +844,23 @@ MacroActualArgumentSyntax* Preprocessor::MacroParser::parseActualArgument() {
 }
 
 MacroFormalArgumentSyntax* Preprocessor::MacroParser::parseFormalArgument() {
-    auto arg = expect(TokenKind::Identifier);
+    Token arg;
```
MikePopoloski: This seems like a complicated way to do this. Could we instead peek() the current token; if it's a keyword or an identifier, consume() it. Otherwise call expect(TokenKind::Identifier) to have it fabricate a placeholder for us.

include/Token.h
```diff
@@ -165,6 +165,8 @@ class Token {
     /// Determines whether the token has the given trivia.
     bool hasTrivia(TriviaKind triviaKind) const;
 
+    bool isKeywordOrIdentifier() const;
```
MikePopoloski: Just a stylistic thing: we can make this a free function (see LexerFacts.cpp) instead of a member?

source/Token.cpp
```diff
@@ -180,6 +180,12 @@ bool Token::hasTrivia(TriviaKind triviaKind) const {
     return false;
 }
 
+bool Token::isKeywordOrIdentifier() const {
```
MikePopoloski: Rather than checking the text, let's just compare against the set of keywords that are actually known. Just copy the switch / case blocks from getTokenKindText() and return true if it's any of the keywords and false if it's not. That means we need only one switch lookup (probably turned into a jump table by the compiler) to do the check.

See SyntaxFacts.cpp for a large collection of existing things that do this.

include/Token.h
```diff
@@ -197,6 +197,7 @@ enum class KeywordVersion : uint8_t {
 };
 
 const StringTable<TokenKind>* getKeywordTable(KeywordVersion version);
+bool isKeywordOrIdentifier(TokenKind kind);
```
drsam94: I think the meaning of the return value would be clearer if this were just, say "isKeyword" or something -- having the or is a bit confusing.

Also, maybe add a comment that this actual checks all keywords, so should only be used conservatively where you are ok letting them be interhchangeable (like Macro arguments) it will produce incorrect results if you really need to know whether or not something is a keyword and the keyword table is not the default due to the begin_keywords directive.

MikePopoloski: Maybe even better would be isPossibleMacroArg or something like that. I'm fine with having a simple isKeyword though; the lexer won't actually produce tokens of those types unless the preprocessor is set properly anyway.

source/Preprocessor.cpp
```diff
@@ -938,7 +938,12 @@ MacroActualArgumentSyntax* Preprocessor::MacroParser::parseActualArgument() {
 }
 
 MacroFormalArgumentSyntax* Preprocessor::MacroParser::parseFormalArgument() {
-    auto arg = expect(TokenKind::Identifier);
+    Token arg = pp.peek(currentMode);
```
MikePopoloski: You should probably use peek() here instead of pp.peek()... see the function definition below this. It does some other stuff like looking in the local macro buffer.

include/Token.h
drsam94: good point about not needing to care about the other kinds, I stand corrected

MikePopoloski:
It'd be nice if we had a unit test for this case too.

source/Preprocessor.cpp
MikePopoloski: Just this one change and then you're good to merge.

evanbryantmiller: Oops. Thought I changed that.

MikePopoloski:
Feel free to push the Rebase and Merge button.

PR #39, opened 2017-02-09
Add support for loop generate constructs

MikePopoloski:
This evaluates the initial, step, and stop expressions to generate blocks in a loop. Test shows it working for a simple straightforward loop, but it should work for any kind of constant expression we can evaluate.

### DIFF
--- a/tests/unit_tests/ModuleBindingTests.cpp
+++ b/tests/unit_tests/ModuleBindingTests.cpp
@@ -158,4 +158,37 @@ endmodule
     CHECK(foo.value.integer() == 1);
 }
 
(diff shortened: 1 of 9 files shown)


tests/unit_tests/ModuleBindingTests.cpp
```diff
@@ -158,4 +158,37 @@ endmodule
     CHECK(foo.value.integer() == 1);
 }
 
+TEST_CASE("Module children (loop generate)", "[binding:modules]") {
```
drsam94: Do we also have / we should have tests to make sure that diagnostics do reasonable things when you have a loop generate and the expressions aren't actually constant and fail to evaluate?

Change looks good to me in general

PR #40, opened 2017-02-09
Added Robust error checking for timescale directive

drsam94:
There are a lot of annoying rules and corner cases for this, but this should handle all of them.

### DIFF
--- a/source/Preprocessor.cpp
+++ b/source/Preprocessor.cpp
@@ -397,35 +397,36 @@ Trivia Preprocessor::handleEndIfDirective(Token directive) {
 }
 
 bool Preprocessor::expectTimescaleSpecifier(Token& value, Token& unit) {
-    auto firstToken = peek();
-    if (firstToken.kind == TokenKind::IntegerLiteral) {
+    auto token = peek();
+    if (token.kind == TokenKind::IntegerLiteral) {
         value = consume();
         unit = expect(TokenKind::Identifier);
-    } else if (firstToken.kind == TokenKind::TimeLiteral) {
-        auto token = consume();
+        return true;
+    } else if (token.kind == TokenKind::TimeLiteral) {
+        bool success = true;
+        // So long as we are dealing with a time literal, we should consume and
+        // output the split tokens, even though those split tokens might be
+        // invalid if the data is poorly formated, i.e with a number other than
+        // 1,10,100
 
         // get_if necessary to check for the case where there is a double TimeLiteral
         const SVInt *val = std::get_if<SVInt>(&token.numericValue());
-        if (val == nullptr) {
-            addError(DiagCode::InvalidTimescaleSpecifier, token.location());
-            return false;
-        }
-
-        // check for the only allowed values: 1, 10, 100
         StringRef numText;
-        if      (*val == 1) numText = "1";
+        if (val == nullptr) {
+            // create a dummy value to place in the generated token
+            val = alloc.emplace<SVInt>(16, 0, true);
+            success = false;
+        }
+        else if (*val == 1) numText = "1";
         else if (*val == 10) numText = "10";
         else if (*val == 100) numText = "100";
-        else {
-            addError(DiagCode::InvalidTimescaleSpecifier, token.location());
-            return false;
-        }
+        else success = false;
 
-        // Split the TimeLiteral into a IntegerLiteral and Identifer token
+        // generate the tokens that come from splitting the TimeLiteral
         Token::Info* valueInfo = alloc.emplace<Token::Info>(token.trivia(),
             numText, token.location(), token.getInfo()->flags);
         value = *alloc.emplace<Token>(TokenKind::IntegerLiteral, valueInfo);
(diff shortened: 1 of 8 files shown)


source/Token.cpp
```diff
@@ -242,4 +242,35 @@ Token Token::createExpected(BumpAllocator& alloc, Diagnostics& diagnostics, Toke
...
+
+StringRef timeUnitToSuffix(TimeUnit unit) {
+    switch (unit) {
+        case TimeUnit::Seconds: return "s";
+        case TimeUnit::Milliseconds: return "ms";
+        case TimeUnit::Microseconds: return "us";
+        case TimeUnit::Nanoseconds: return "ns";
+        case TimeUnit::Picoseconds: return "ps";
+        case TimeUnit::Femtoseconds: return "fs";
+    }
```
MikePopoloski: Need a default case here to avoid warnings.

source/Preprocessor.cpp
```diff
@@ -395,28 +396,70 @@ Trivia Preprocessor::handleEndIfDirective(Token directive) {
...
-void Preprocessor::expectTimescaleSpecifier(Token& value, Token& unit) {
-    // TODO: check for allowed values
-    auto token = peek();
-    if (token.kind == TokenKind::IntegerLiteral) {
+bool Preprocessor::expectTimescaleSpecifier(Token& value, Token& unit) {
+    if (peek().kind == TokenKind::IntegerLiteral) {
         value = consume();
         unit = expect(TokenKind::Identifier);
+    } else {
+        auto token = expect(TokenKind::TimeLiteral);
```
MikePopoloski: Rather than expect() here, maybe just use peek() and look for the TimeLiteral token. That way we don't potentially create a new missing token with some valid trivia (like if there was whitespace there or something) and then throw it away.

Basically if peek() doesn't give us an integer literal or a time literal, just expect(IntegerLiteral) expect(Identifier) and we're done.

source/Preprocessor.cpp
```diff
@@ -395,28 +396,70 @@ Trivia Preprocessor::handleEndIfDirective(Token directive) {
...
+        if (token.isMissing()) {
+            value = Token::createMissing(alloc, TokenKind::IntegerLiteral, token.location());
+            unit = Token::createMissing(alloc, TokenKind::Identifier, token.location());
+            return false;
+        }
+
+        const SVInt *val = std::get_if<SVInt>(&token.numericValue());
```
MikePopoloski: If you modify to use peek() above then this won't be an issue, but right now we're consuming that time unit token and then losing it if doesn't have a numeric value. We want to always be able to round trip back to the original source even when weird errors like this happen.

source/Preprocessor.cpp
```diff
@@ -395,28 +396,70 @@ Trivia Preprocessor::handleEndIfDirective(Token directive) {
...
+        else if (*val == 10) numText = "10";
+        else if (*val == 100) numText = "100";
+        else {
+            addError(DiagCode::InvalidTimescaleSpecifier, token.location());
+            return false;
+        }
+
+        // Split the TimeLiteral into a IntegerLiteral and Identifer token
+        Token::Info* valueInfo = alloc.emplace<Token::Info>(token.trivia(),
+            numText, token.location(), token.getInfo()->flags | TokenFlags::IsFromPreprocessor);
```
MikePopoloski: Hmm, we probably don't want to add the IsFromPreprocessor flag. The point of that is so that we can accurately reproduce the source that the user typed; anything from the preprocessor (such as from includes or macro expansion) wasn't actually there, so we hide it, but these tokens are actually there in the source.

source/Preprocessor.cpp
drsam94: Even with peek, this check is necessary to make sure it isn't a double, as then it would be a valid TimeLiteral but not valid for a Timescale specifier

source/Preprocessor.cpp
drsam94: alright. I added it because it is literally true, but that makes sense. These are a reinterpretation of an existing token, rather than actual newly added tokens.

source/Preprocessor.cpp
```diff
@@ -424,16 +423,20 @@ bool Preprocessor::expectTimescaleSpecifier(Token& value, Token& unit) {
...
 
         StringRef timeUnitSuffix = timeUnitToSuffix(token.numericFlags().unit);
         Token::Info* unitInfo = alloc.emplace<Token::Info>(token.trivia(),
             timeUnitSuffix,
-            token.location() + numText.length(), token.getInfo()->flags | TokenFlags::IsFromPreprocessor);
+            token.location() + numText.length(), token.getInfo()->flags);
         unit = *alloc.emplace<Token>(TokenKind::Identifier, unitInfo);
         unitInfo->extra = IdentifierType::Normal;
+    } else {
+        value = Token::createExpected(alloc, diagnostics, firstToken, TokenKind::IntegerLiteral, lastConsumed);
```
MikePopoloski: Probably easier to just call expect() here, but this works too.

PR #43, opened 2017-02-09
SemanticModel: handle ProceduralBlock and DataDeclaration

Baranowski:
Create Symbols for ProceduralBlock and Data.
Let GenerateBlock have it's own scope
Test the above

### DIFF
--- a/include/Symbol.h
+++ b/include/Symbol.h
@@ -27,7 +27,7 @@ enum class SymbolKind {
     GenerateBlock,
     LocalVariable,
     ProceduralBlock,
-    Data
+    Variable
 };
 
 class Symbol {
@@ -204,9 +204,9 @@ public:
         Symbol(SymbolKind::Genvar, name, location) {}
 };
 
-class DataSymbol : public Symbol {
+class VariableSymbol : public Symbol {
 public:
-    class DataSymbolModifiers {
+    class VariableSymbolModifiers {
     public:
         unsigned int constM : 1;
         unsigned int varM : 1;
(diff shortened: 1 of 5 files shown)


include/Symbol.h
```diff
@@ -25,7 +25,9 @@ enum class SymbolKind {
     Attribute,
     Genvar,
     GenerateBlock,
-    LocalVariable
+    LocalVariable,
+    ProceduralBlock,
+    Data
```
MikePopoloski: I think I would just call this is a Variable symbol instead of data, and LocalVariable will probably be removed in favor of that.

source/SemanticModel.cpp
```diff
@@ -359,20 +366,81 @@ void SemanticModel::handleLoopGenerate(const LoopGenerateSyntax* syntax, SmallVe
...
+    Scope *scope = alloc.emplace<Scope>(_scope);
+    SmallVectorSized<const Symbol*, 2> children;
+    //TODO handleStatement(syntax->statement, children, scope);
+    results.append(alloc.emplace<ProceduralBlock>(
+        children.copy(alloc), proceduralBlockKindFromKeyword(syntax->keyword), scope));
+}
+
+void SemanticModel::handleDataDeclaration(const DataDeclarationSyntax *syntax, SmallVector<const Symbol *>& results, Scope* scope) {
+    DataSymbol::DataSymbolModifiers modifiers;
+    modifiers.constM = std::find_if(syntax->modifiers.begin(), syntax->modifiers.end(), [](auto tk) {return tk.kind == TokenKind::ConstKeyword;}) != syntax->modifiers.end();
```
MikePopoloski: This section is a little wasteful since we loop over every modifier for every modifier type. Not a huge deal since there won't ever be a lot of them, but to do the minimum work we'd just loop over syntax->modifiers once, switch() on the kind and then set the modifier member appropriately.

Baranowski:
Rebased and pushed to master

PR #45, opened 2017-02-09
Clean up function parsing and bind function symbols

MikePopoloski:
This binds function and task definitions (though not their bodies yet; that'll be my next step).

The parser code for function ports was being lazy and reusing module port parsing, but they're actually fairly different so I added proper parsing support there.


MikePopoloski:
You're right. I added a test for it.

PR #47, opened 2017-02-10
Start adding statement binding, merge LocalVariable and Variable symbols

MikePopoloski:
Currently only supports return statements. I should probably rename ExpressionBinder now, since it also binds statements.

This also merges the LocalVariable and Variable symbols, since they're basically the same thing, and adds an error when symbol lookup fails to find them.

### DIFF
--- a/include/Symbol.h
+++ b/include/Symbol.h
@@ -9,6 +9,10 @@
 
 namespace slang {
 
+class BoundExpression;
+class BoundStatement;
+class BoundStatementList;
+
 enum class SymbolKind {
     Unknown,
     IntegralType,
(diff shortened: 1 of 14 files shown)


include/Symbol.h
```diff
@@ -270,27 +271,29 @@ class ProceduralBlock : public Symbol {
         : children(children), kind(kind), scope(scope) {}
 };
 
-class FormalArgumentSymbol : public Symbol {
+class FormalArgumentSymbol : public VariableSymbol {
```
Baranowski: If FAS has no modifiers, shouldn't it have a common base with VariableSymbol, instead of inheriting from VariableSymbol?

MikePopoloski: Yeah, possibly we could add another base type for both, but I think an argument can actually have some of the modifiers (such as const and var).

Baranowski:
lgtm

Issue #6, opened 2017-02-04
Support macro-replaced include file names


MikePopoloski:
Include file names can actually be specified as a macro that should be expanded; this is fairly straightforward but pretty annoying since a bracketed include name will show up as a whole bunch of tokens in the stream. C++ supports this as well, so we can take inspiration from GCC or Clang here.

PR #48, opened 2017-02-10
added support for macro-expanded include filenames

drsam94:
This addresses Issue #6

### DIFF
--- a/source/Preprocessor.cpp
+++ b/source/Preprocessor.cpp
@@ -183,8 +183,8 @@ Trivia Preprocessor::handleIncludeDirective(Token directive) {
             // up to the '>'' in order to get the file name
             SourceLocation rootExpansionLoc = sourceManager.getExpansionLoc(fileName.location());
             Token concatenatedFileName = fileName;
-            std::vector<Token> tokens;
-            tokens.push_back(fileName);
+            SmallVectorSized<Token, 8> tokens;
+            tokens.append(fileName);
             Token nextToken = peek();
             while (nextToken.kind != TokenKind::GreaterThan) {
                 // TODO: we should probably stop at something like a >>, >>= token,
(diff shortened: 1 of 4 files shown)


source/Preprocessor.cpp
```diff
@@ -160,13 +161,72 @@ Token Preprocessor::nextRaw(LexerMode mode) {
...
+        } else {
+            addError(DiagCode::ExpectedIncludeFileName, fileName.location());
+        }
+    }
+    Token end = parseEndOfDirective(suppressEODError);
 
     // path should be at least three chars: "a" or <a>
     StringRef path = fileName.valueText();
+
+    // An includeFileName's value
```
MikePopoloski: Not sure what this comment is saying.

source/Preprocessor.cpp
```diff
@@ -160,13 +161,72 @@ Token Preprocessor::nextRaw(LexerMode mode) {
...
+            // In this case, we know that the first token is LessThan, and
+            // if things are proper, the last token is >, but there can be an
+            // arbitrary number of tokens that the macro expanded to in-between
+            // (as file names have no restrictions like identifiers do)
+            // so let us now concatenate all of the tokens from the macro expansion
+            // up to the '>'' in order to get the file name
+            SourceLocation rootExpansionLoc = sourceManager.getExpansionLoc(fileName.location());
+            Token concatenatedFileName = fileName;
+            std::vector<Token> tokens;
+            tokens.push_back(fileName);
```
MikePopoloski: Prefer a SmallVectorSized<> here instead of an std::vector. in the common case where there aren't many entries, we avoid a heap allocation.

source/Preprocessor.cpp
```diff
@@ -160,13 +161,72 @@ Token Preprocessor::nextRaw(LexerMode mode) {
...
+                }
+                consume();
+                tokens.push_back(nextToken);
+                nextToken = peek();
+                if (nextToken.kind == TokenKind::GreaterThan) {
+                    tokens.push_back(nextToken);
+                    // do stringification
+                    fileName = Lexer::stringify(alloc, fileName.location(), fileName.getInfo()->trivia,
+                        &tokens.front(), &tokens.back() + 1, true);
+                    fileName = *alloc.emplace<Token>(TokenKind::IncludeFileName, fileName.getInfo());
```
MikePopoloski: Not sure why you're heap allocating and then immediately copying back here.

drsam94: ah, oops -- throughout most of my changes I thought I had to mutate info in some way, but you are right that this is redundant

source/Preprocessor.cpp
drsam94: Alright

source/Preprocessor.cpp
drsam94: pretty sure I cut myself off

PR #49, opened 2017-02-10
Handle MacroPastes properly in several strange cases

evanbryantmiller:
Handle MacroPaste properly in these cases:

- The MacroPaste is next to whitespace or the end of the file
- Between two identifiers which would form a keyword
- Between two tokens that cannot be merged.

### DIFF
--- a/include/Token.h
+++ b/include/Token.h
@@ -571,6 +571,7 @@ enum class TokenKind : uint16_t {
     MacroQuote,
     MacroEscapedQuote,
     MacroPaste,
+    DummyMacroPaste,
 
     // special placeholders for intrinsic macros
     IntrinsicFileMacro,
--- a/include/Token.h
+++ b/include/Token.h
@@ -571,7 +571,6 @@ enum class TokenKind : uint16_t {
     MacroQuote,
     MacroEscapedQuote,
     MacroPaste,
-    DummyMacroPaste,
 
     // special placeholders for intrinsic macros
     IntrinsicFileMacro,
--- a/source/Lexer.cpp
+++ b/source/Lexer.cpp
@@ -109,7 +109,7 @@ Lexer::Lexer(BufferID bufferId, StringRef source, BumpAllocator& alloc, Diagnost
     }
 }
 
-Token Lexer::concatenateTokens(BumpAllocator& alloc, Token left, Token right) {
+Token Lexer::concatenateTokens(BumpAllocator& alloc, Token left, Token right, bool& error) {
     // TODO: think about what happens if there is trivia in right token
     // TODO: think about what we should set the resulting location to be in order to capture expansion info
     auto location = left.location();
@@ -118,8 +118,10 @@ Token Lexer::concatenateTokens(BumpAllocator& alloc, Token left, Token right) {
     // if either side is empty, we have an error; the user tried to concatenate some weird kind of token
     auto leftText = left.rawText();
     auto rightText = right.rawText();
-    if (!leftText || !rightText)
+    if (!leftText || !rightText) {
+        error = true;
         return Token();
+    }
 
     // combine the text for both sides; make sure to include room for a null
     uint32_t newLength = leftText.length() + rightText.length() + 1;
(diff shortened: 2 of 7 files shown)


tests/unit_tests/PreprocessorTests.cpp
```diff
@@ -194,7 +194,7 @@ TEST_CASE("Function macro (keyword as formal argument)", "[preprocessor]") {
 }
 
 TEST_CASE("Macro pasting (identifiers)", "[preprocessor]") {
-    auto& text = "`define FOO(x,y) x   ``   _blah``y\n`FOO(   bar,    _BAZ)";
+    auto& text = "`define FOO(x,y) x``_blah``y\n`FOO(   bar,    _BAZ)";
```
drsam94: In most prior conversation I think we had decided that we wanted to maintain the macro-paste-ignoring-whitespace feature, what is the reason for getting rid of it? does HRT code actually use the spacing around a macro paste operator?

include/Token.h
```diff
@@ -571,6 +571,7 @@ enum class TokenKind : uint16_t {
     MacroQuote,
     MacroEscapedQuote,
     MacroPaste,
+    DummyMacroPaste,
```
drsam94: Reading through this I figured out that "DummyMacroPaste" indicates a macropaste with spacing so that it doesn't really do any useful pasting; I feel like somewhere in this change a commend should spell that out.

tests/unit_tests/PreprocessorTests.cpp
evanbryantmiller: My original reason was I realized that it the compilers I tried on EDA playground did not have this feature.  HRT code does rely on whitespace existing next to a macro paste operator (e.g. hrt_prologue.svh:182).  The main reason I prefer this way is that it means you don't have to think of the operator as a paste between two neighboring tokens but instead you can just surround your argument with this operator regardless of its context to treat it more like an escaping mechanism.  This idea of putting the operator on both sides of the argument even when it isn't necessary is heavily used in our codebase.

include/Token.h
evanbryantmiller: I'll add a comment in Preprocessor.cpp

tests/unit_tests/PreprocessorTests.cpp
MikePopoloski: So after looking at the spec again, I agree that the paste operator should work this way. It's less of a paste operator like in C and more just some placeholder token to separate other tokens.

With that said, I think this change is more complicated than it needs to be. In the loop to handle macro expansion, you can check if the paste operator token and the rhs token both have .trivia().empty(). If not, then you just ignore the operator (maybe we'll issue a warning or something). That way we don't need a new dummy token type.

evanbryantmiller: Thanks. I hadn't seen trivia yet.

source/Lexer.cpp
```diff
@@ -129,32 +131,29 @@ Token Lexer::concatenateTokens(BumpAllocator& alloc, Token left, Token right) {
     mem[newLength - 1] = '\0';
     StringRef combined{ (char*)mem, newLength };
 
-    // common case: identifier + identifier
```
MikePopoloski: Why did you get rid of this case? It's an optimization that will probably be true in most cases.

tests/file_tests/main.cpp
```diff
@@ -1,58 +1,37 @@
...
 #include <string>
-#include <vector>
 
 #include "slang.h"
 
 using namespace slang;
 using namespace std::literals;
 
-static const char RelativeTestPath[] = "tests/file_tests/corpus/include";
+static const char RelativeTestPath[] = "/Volumes/hrt/surge/testing";
```
MikePopoloski: Probably shouldn't commit this change, since other people don't necessarily check out to the same location.

evanbryantmiller: True. Didn't mean to add this file.

source/Lexer.cpp
evanbryantmiller: I got rid of this since two identifiers don't always form an identifier.  They can instead form a keyword.  I tested this on eda playground and those compilers do recognize keywords formed in this way.

MikePopoloski: OK, that's a good point.

Issue #13, opened 2017-02-04
Stop lexing if we encounter too many errors


MikePopoloski:
The lexer tries to handle non-ASCII characters embedded in the source gracefully. If there are too many of them however, maybe the user accidentally pointed us at a binary file or something. Keep a counter and if we hit some threshold just swallow the rest of the text instead of producing a billion errors.

PR #50, opened 2017-02-10
Made Lexer stop parsing after encountering too many errors

drsam94:
This addresses Issue #13

### DIFF
--- a/include/Diagnostics.h
+++ b/include/Diagnostics.h
@@ -48,6 +48,7 @@ enum class DiagCode : uint8_t {
     BadDecimalDigit,
     BadHexDigit,
     DirectiveNotFirstOnLine,
+    TooManyErrors,
 
     // numeric
     LiteralSizeIsZero,
--- a/include/Diagnostics.h
+++ b/include/Diagnostics.h
@@ -48,7 +48,7 @@ enum class DiagCode : uint8_t {
     BadDecimalDigit,
     BadHexDigit,
     DirectiveNotFirstOnLine,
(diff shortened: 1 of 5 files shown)


include/Diagnostics.h
```diff
@@ -48,6 +48,7 @@ enum class DiagCode : uint8_t {
     BadDecimalDigit,
     BadHexDigit,
     DirectiveNotFirstOnLine,
+    TooManyErrors,
```
MikePopoloski: Might want to call this TooManyLexerErrors so that it's obvious that it only applies to the lexer.

source/Lexer.cpp
```diff
@@ -220,6 +223,12 @@ Token Lexer::lex(LexerMode mode, KeywordVersion keywordVersion) {
 
     onNewLine = false;
     info->rawText = lexeme();
+    if (diagnostics.count() >= MAX_LEXER_ERRORS) {
+        // Stop any further lexing by claiming to at the end of the buffeer;
+        addError(DiagCode::TooManyErrors, currentOffset());
+        // Ensure we don't look at the source any more
+        sourceBuffer = nullptr;
```
MikePopoloski: Rather than clearing out the sourceBuffer, instead I think we should skip from the current location to the end of the buffer and turn all that skipped into a Trivia object on the EoF token we create. That way we 1) preserve the text if the caller wants to look at it for some reason and 2) we don't need an additional branch above on sourceBuffer.

PR #51, opened 2017-02-10
Parse immediate assertion members

evanbryantmiller:
Though I couldn't find an explicit place in the spec where it said immediate assertions can be members, there is an example in the spec at the bottom of page 337 that has such a member.


source/Parser.cpp
```diff
@@ -339,13 +339,22 @@ MemberSyntax* Parser::parseMember() {
             // TODO: Don't assume we have an assert here; this could be an accidental label or something
             auto name = consume();
             auto label = alloc.emplace<NamedLabelSyntax>(name, expect(TokenKind::Colon));
-            return alloc.emplace<ConcurrentAssertionMemberSyntax>(attributes, parseConcurrentAssertion(label, nullptr));
+            auto statement = parseAssertionStatement(label, nullptr);
+            if (dynamic_cast<ImmediateAssertionStatementSyntax*>(statement)) {
```
MikePopoloski: Prefer statement->kind over dynamic casting.

evanbryantmiller:
Ignore this for now.  I've made some mistakes in syntax.txt

evanbryantmiller:
I just realized I was supposed to update syntax.txt instead of AllSyntax.h.  I made changes in this last commit to make that change for ImmediateAssertionMemberSyntax and EmptyIdentifierNameSyntax.  I had to change syntax_gen.py slightly to get the latter to work.

tools/syntax_gen.py
```diff
@@ -118,7 +118,7 @@ def generate(outf, name, tags, members, alltypes):
 	outf.write('        {}({}{}){}\n'.format(base, kindValue, baseInitializers, initializers))
 	outf.write('    {\n')
 
-	if len(members) == 0:
+	if len(members) == 0 and not forceEmpty:
```
MikePopoloski: I think instead of adding 'nothing' you could just make a small tweak and reuse empty. Right now everything that uses 'empty' also sets final=false. You can just check here if final=True (which is the default) and then output the functions in that case.

source/Parser.cpp
```diff
@@ -341,13 +341,22 @@ MemberSyntax* Parser::parseMember() {
             // TODO: Don't assume we have an assert here; this could be an accidental label or something
             auto name = consume();
             auto label = alloc.emplace<NamedLabelSyntax>(name, expect(TokenKind::Colon));
-            return alloc.emplace<ConcurrentAssertionMemberSyntax>(attributes, parseConcurrentAssertion(label, nullptr));
+            auto statement = parseAssertionStatement(label, nullptr);
+            if (statement->kind == SyntaxKind::ImmediateAssertStatement) {
```
MikePopoloski: There are a few kinds of things covered here right? Covers and Assumes in addition to Asserts.

source/Parser.cpp
```diff
@@ -341,13 +341,22 @@ MemberSyntax* Parser::parseMember() {
...
+            return alloc.emplace<ConcurrentAssertionMemberSyntax>(attributes, (ConcurrentAssertionStatementSyntax*) statement);
         }
         case TokenKind::AssertKeyword:
         case TokenKind::AssumeKeyword:
         case TokenKind::CoverKeyword:
-        case TokenKind::RestrictKeyword:
-            return alloc.emplace<ConcurrentAssertionMemberSyntax>(attributes, parseConcurrentAssertion(nullptr, nullptr));
+        case TokenKind::RestrictKeyword: {
+            auto statement = parseAssertionStatement(nullptr, nullptr);
+            if (statement->kind == SyntaxKind::ImmediateAssertStatement) {
```
MikePopoloski: Same here.

source/Parser.cpp
evanbryantmiller: Yes, those are both covered here as well.  The keyword token in the ImmediateAssertionStatementSyntax identifies which thing it is.

source/Parser.cpp
evanbryantmiller: Same here.

tools/syntax_gen.py
evanbryantmiller: PrimaryExpressionSyntax currently doesn't use final=false.  I assume I can set that to false?

PR #53, opened 2017-02-10
Add ScriptSession to make evaluating snippets easy

MikePopoloski:
This makes it easy to test evaluating arbitrary code. I also added initializer support to variable symbols in this change, and a function to the parser that tries to guess what it's looking at.


tests/unit_tests/EvalTests.cpp
```diff
@@ -0,0 +1,18 @@
+#include "Catch/catch.hpp"
+#include "slang.h"
+
+using namespace slang;
+
+namespace {
+
+TEST_CASE("Simple eval", "[eval") {
```
drsam94: This looks like an awesome thing to have, this should be very useful for testing stuff.

PR #55, opened 2017-02-10
ParameterSymbol always has meaningful type

Baranowski:


MikePopoloski:
I don't think this is the right way to fix this. Parameters with empty DataTypeSyntax entries need to do type inference, which is why they're filled in later. I made a change in a previous commit that makes makeTypeSymbol() return an error type instead of nullptr, so I think that'll fix the crash you were running into.

PR #57, opened 2017-02-10
Parse signed casting

evanbryantmiller:
Parse switching the signedness of a variable a la `signed'(x)` or `unsigned'(x)`

### DIFF
--- a/source/Parser_expressions.cpp
+++ b/source/Parser_expressions.cpp
@@ -214,6 +214,16 @@ ExpressionSyntax* Parser::parsePrimaryExpression() {
             }
             break;
         }
+        case TokenKind::SignedKeyword:
+        case TokenKind::UnsignedKeyword:
+            auto signing = consume();
+            auto apostrophe = expect(TokenKind::Apostrophe);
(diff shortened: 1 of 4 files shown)


evanbryantmiller:
All other casts assume the lefthand side is an expression or a type.  The keywords signed and unsigned correspond to neither.

source/Parser_expressions.cpp
```diff
@@ -221,8 +221,17 @@ ExpressionSyntax* Parser::parsePrimaryExpression() {
             // 3. implicit class handles
             // 4. any of [1-3] with an assignment pattern
             // 5. any of [1-3] with a cast expression
-            // 6. error
-            if (isPossibleDataType(kind) && kind != TokenKind::Identifier && kind != TokenKind::UnitSystemName) {
+            // 6. signing cast expression
+            // 7. error
+            if (kind == TokenKind::SignedKeyword || kind == TokenKind::UnsignedKeyword) {
```
MikePopoloski: This looks good, but could you make these case labels above? That avoids the extra branch.

PR #58, opened 2017-02-10
Package import

denisrr:
process package declarations
create global package namespace
handle explicit and wildcard imports in module header and body


source/ExpressionBinder.cpp
```diff
@@ -16,7 +16,7 @@ namespace slang {
 ExpressionBinder::ExpressionBinder(SemanticModel& sem, const Scope* scope) :
     sem(sem), alloc(sem.getAllocator()), scope(scope)
 {
-    ASSERT(scope);
+        ASSERT(scope);
```
MikePopoloski: Weird indentation change.

source/SemanticModel.cpp
```diff
@@ -69,13 +71,65 @@ InstanceSymbol* SemanticModel::makeImplicitInstance(const ModuleDeclarationSynta
     return alloc.emplace<InstanceSymbol>(module, true);
 }
 
+void SemanticModel::makePackages() {
+    HashMap<StringRef, SourceLocation> nameDupMap;
+    for (auto pkg : declTable.getPackages()) {
+        auto name = pkg->header->name.valueText();
+        Scope *scope = alloc.emplace<Scope>();
+        ASSERT(packages.add(alloc.emplace<ModuleSymbol>(pkg, scope, ArrayRef<const Symbol *>())));
```
MikePopoloski: Don't put code that must execute inside an ASSERT()... if it gets compiled out (which ASSERTs often due in optimized builds) then your code won't run anymore.

PR #59, opened 2017-02-13
Support call expressions

MikePopoloski:
- Bind and evaluate function call expressions and function statement bodies.
- Allocate storage for local variable definitions
- Handle built-in system function calls
- Add tests for the above functionality

### DIFF
--- a/tests/unit_tests/EvalTests.cpp
+++ b/tests/unit_tests/EvalTests.cpp
@@ -15,4 +15,12 @@ TEST_CASE("Simple eval", "[eval") {
     CHECK(value.integer() == 13);
 }
 
+TEST_CASE("Function call", "[eval") {
+    ScriptSession session;
+    session.eval("function logic [15:0] foo(int a, int b); return a + b; endfunction");
+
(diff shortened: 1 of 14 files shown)


tests/unit_tests/EvalTests.cpp
```diff
@@ -15,4 +15,12 @@ TEST_CASE("Simple eval", "[eval") {
     CHECK(value.integer() == 13);
 }
 
+TEST_CASE("Function call", "[eval") {
```
drsam94: typo, [eval]

also should also probably have a test for statement lists ?

MikePopoloski: Yeah, but right now I only have support for return statements so there's not really anything to test there yet.

drsam94: well, there is at least one list of statements you can test that I am not sure if is being handled properly now...

~~~
return 2;
return 1;
~~~

MikePopoloski: True, I don't test for that and it actually doesn't work properly right now.

PR #63, opened 2017-02-13
Added support for evaluating conditional and shift operators

drsam94:


source/SVInt.cpp
```diff
@@ -476,6 +476,71 @@ SVInt SVInt::lshr(uint32_t amount) const {
...
+    if (!signFlag)
+        return lshr(rhs);
+    if (rhs.hasUnknown())
+        return createFillX(bitWidth, signFlag);
+    if (rhs >= bitWidth)
+        return SVInt(bitWidth, *this >= 0 ? 0 : -1, signFlag);
+
+    uint16_t contractedWidth = bitWidth - rhs.getAssertUInt64();
+    SVInt tmp = lshr(rhs);
+    tmp.bitWidth = contractedWidth;
```
MikePopoloski: I think this is a potentially dangerous operation; if bitWidth was > 64 and now is less, functions will assume the memory is in a different format (pointer to the heap opposed to a local 64-bit value) and crash.

source/SVInt.cpp
```diff
@@ -476,6 +476,71 @@ SVInt SVInt::lshr(uint32_t amount) const {
...
+    if (amount >= bitWidth)
+        return SVInt(bitWidth, *this >= 0 ? 0 : -1, signFlag);
+
+    uint16_t contractedWidth = bitWidth - amount;
+    SVInt tmp = lshr(amount);
+    tmp.bitWidth = contractedWidth;
+    return signExtend(tmp, bitWidth);
+}
+
+SVInt SVInt::ambiguousConditionalCombination(const SVInt& rhs) const {
```
MikePopoloski: I feel like this should probably live as a helper method in the evaluator; it's too specific for anyone else to want to use it.

source/SVInt.cpp
drsam94: That is a good point, that case needs to be handled differently

source/SVInt.cpp
drsam94: Sure, that makes sense. this is definitely not the world's most popular binary operator

drsam94: now that I think about it, it would be hard to make this a helper method in the evaluator, as it has to access private fields of SVInt (val/pVal) -- we could expose those to friends, or this can stay part of SVInt

source/SVInt.cpp
```diff
@@ -476,6 +476,71 @@ SVInt SVInt::lshr(uint32_t amount) const {
...
+
+    if (isSingleWord()) {
+        // If the inputs are single word, as the inputs are unequal, we need to
+        // combine into a 2-word output with a word for unknowns
+        SVInt tmp = createFillX(bitWidth, bothSigned);
+        tmp.pVal[1] = val ^ rhs.val;
+        tmp.pVal[0] = ~tmp.pVal[1] & val & rhs.val;
+        return tmp;
+    }
+    SVInt tmp(*this);
```
MikePopoloski: Creating this temporary is wasteful since you immediately delete the memory and allocate your own. There's a separate constructor that lets you just pass in allocated memory to use.

source/SVInt.cpp
```diff
@@ -476,6 +476,71 @@ SVInt SVInt::lshr(uint32_t amount) const {
...
+    delete[] tmp.pVal;
+    uint32_t words = getNumWords(bitWidth, false) * 2;
+    tmp.pVal = new uint64_t[words];
+
+    // Unkown if either bit is unknown or bits differ
+    for (uint32_t i = 0; i < words; i++)
```
MikePopoloski: You should be able to combine the two loops to avoid looping twice.

source/SVInt.cpp
MikePopoloski: Ok, fair enough.

source/SVInt.cpp
```diff
@@ -476,6 +476,85 @@ SVInt SVInt::lshr(uint32_t amount) const {
...
+        return lshr(amount);
+    if (amount == 0)
+        return *this;
+    if (amount >= bitWidth)
+        return SVInt(bitWidth, *this >= 0 ? 0 : -1, signFlag);
+
+    uint16_t contractedWidth = bitWidth - amount;
+    SVInt tmp = lshr(amount);
+    tmp.bitWidth = contractedWidth;
+    return signExtend(tmp, bitWidth);
```
MikePopoloski: I think this still has the same problem. Probably you should follow the pattern for lsh() and make the SVInt version call into the uint32_t version.

drsam94: oops, I fixed the SVInt version and in so doing totally forgot that the uint32_t version existed. I will make one call the other.

PR #68, opened 2017-02-14
Add Binding and evaluation for Unsized Unbased Literals 

drsam94:
('0, '1, 'x, 'z)

### DIFF
--- a/source/ConstantEvaluator.cpp
+++ b/source/ConstantEvaluator.cpp
@@ -54,7 +54,24 @@ ConstantValue ConstantEvaluator::evaluate(const BoundNode* tree) {
 }
 
 ConstantValue ConstantEvaluator::evaluateLiteral(const BoundLiteral* expr) {
-    return expr->value;
+    switch (expr->syntax->kind) {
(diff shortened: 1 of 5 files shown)


source/ConstantEvaluator.cpp
```diff
@@ -54,7 +54,24 @@ ConstantValue ConstantEvaluator::evaluate(const BoundNode* tree) {
...
+    switch (expr->syntax->kind) {
+        case SyntaxKind::UnbasedUnsizedLiteralExpression: {
+            // In this case, the value depends on the final size, so we evaluate
+            // the right value here
+            logic_t digit = (logic_t)expr->value.integer();
+            uint16_t width = expr->type->width();
+            bool isSigned = expr->type->isSigned();
+            switch (digit.value) {
+                case 0: return SVInt(width, 0, isSigned);
+                case 1: return SVInt(width, (1 << width) - 1, isSigned);
```
MikePopoloski: This is probably not right for very large integers; it will overflow a 64-bit value.

drsam94: ah, you are right. I forgot about that case.

PR #70, opened 2017-02-14
Map all SyntaxKinds to concrete types

MikePopoloski:
This is useful for a bunch of reasons. First, I discovered some unused kinds in SyntaxKind which didn't map to anything at all. More usefully, it lets us verify as<> casts and also lets us write tree visitors without needing double virtual dispatch.


evanbryantmiller:
Thanks for doing this. I was just wanting something like this earlier today.  It looks good to me.  The only thing that I think could be improved would be if we could guarantee in syntax_gen.py that every type has a kind, has a kind map, or sets final=false.  That would help enforce that this stays up to date.

MikePopoloski:
Good point. I added that check to syntax_gen.py and it caught an unused type.

PR #72, opened 2017-02-14
Add support for concatenation, replication, and wildcard equality operators

drsam94:
Support added to ExpressionBinder and ConstantEvaluator

### DIFF
--- a/source/SVInt.cpp
+++ b/source/SVInt.cpp
@@ -552,6 +552,14 @@ SVInt SVInt::ambiguousConditionalCombination(const SVInt& rhs) const {
     return SVInt(data, bitWidth, bothSigned, true);
 }
 
+SVInt SVInt::replicate(const SVInt& times) const {
+    uint32_t n = times.getAssertUInt32();
+    SmallVectorSized<const SVInt*, 8> buffer;
+    for (size_t i = 0; i < n; ++i)
+        buffer.append(this);
+    return concatenate(ArrayRef<const SVInt*>(buffer.begin(), buffer.end()));
+}
+
 size_t SVInt::hash(size_t seed) const {
     return xxhash(getRawData(), getNumWords() * WORD_SIZE, seed);
 }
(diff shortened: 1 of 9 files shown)


source/SVInt.cpp
```diff
@@ -1690,4 +1699,107 @@ bool exactlyEqual(const SVInt& lhs, const SVInt &rhs) {
...
+    if (lhs.bitWidth != rhs.bitWidth) {
+        bool bothSigned = lhs.signFlag && rhs.signFlag;
+        if (lhs.bitWidth < rhs.bitWidth)
+            return wildcardEqual(extend(lhs, rhs.bitWidth, bothSigned), rhs);
+        else
+            return wildcardEqual(lhs, extend(rhs, lhs.bitWidth, bothSigned));
+    }
+
+    size_t words = SVInt::getNumWords(lhs.bitWidth, false);
+    // Do word-for-word wildcard comparison, being careful that one argument
```
MikePopoloski: "not have unknown"

source/SVInt.cpp
```diff
@@ -1690,4 +1699,107 @@ bool exactlyEqual(const SVInt& lhs, const SVInt &rhs) {
...
+    // might nothave unknwon flags
+    for (size_t i = 0; i < words; ++i) {
+        // bitmask of bits that actually matter in the comparison
+        uint64_t mask = ~((lhs.unknownFlag ? lhs.pVal[i + words] : 0) |
```
MikePopoloski: X and Z are only wildcards in the rhs; in the lhs side they either match a wildcard, or they cause the result to be X (so this function can't just return a bool).

source/SVInt.cpp
```diff
@@ -1690,4 +1699,107 @@ bool exactlyEqual(const SVInt& lhs, const SVInt &rhs) {
...
+    return true;
+}
+
+
+void copyBits(uint8_t* dest, uint16_t destBitOffset, uint8_t* src, uint16_t bitLength) {
+    // Like memcpy, but at the bit level instead of bytes
+    // i.e if destBitOffset % 8 == bitLength % 8 == 0, this is
+    // equivalent to memcpy(dest + destBitOffset / 8, src, bitLength / 8)
+
+    // Get the first byte we want to write to, and the reamining bits are a bit offset
```
MikePopoloski: "remaining"

source/SVInt.cpp
```diff
@@ -1690,4 +1699,107 @@ bool exactlyEqual(const SVInt& lhs, const SVInt &rhs) {
...
+    while (bitLength > 0) {
+        // Number of bits we are writing to this byte
+        uint16_t bitsToWrite = std::min<uint8_t>(bitLength, 8 - destBitOffset);
+        // get the next 8 bits of src, probably not byte aligned
+        uint8_t srcByte = (*src >> srcBitOffset) + (src[1] << (8 - srcBitOffset));
+
+        *dest = (*dest   & ((1 << destBitOffset) - 1)) + // preserved bits
+                ((srcByte & ((1 << bitsToWrite) - 1)) << destBitOffset); // new bits
+        // next write will begin at the next byte boundry
+        destBitOffset = 0;
```
MikePopoloski: It's probably worth manually unrolling this first iteration, since iterations 1+ through the loop are very different since destBitOffset will always be 0.

source/SVInt.cpp
```diff
@@ -1690,4 +1699,107 @@ bool exactlyEqual(const SVInt& lhs, const SVInt &rhs) {
...
+        uint64_t lhsWord = lhs.isSingleWord() ? lhs.val : lhs.pVal[i];
+        uint64_t rhsWord = rhs.isSingleWord() ? rhs.val : rhs.pVal[i];
+        if ((lhsWord & mask) != (rhsWord & mask))
+            return false;
+    }
+    return true;
+}
+
+
+void copyBits(uint8_t* dest, uint16_t destBitOffset, uint8_t* src, uint16_t bitLength) {
```
MikePopoloski: Maybe put this function in SVIntHelpers.h and mark it static.

source/SVInt.cpp
```diff
@@ -1690,4 +1699,107 @@ bool exactlyEqual(const SVInt& lhs, const SVInt &rhs) {
...
+    uint16_t bits = 0;
+    bool unknownFlag = false;
+
+    for (const auto& op : operands) {
+        bits += op.bitWidth;
+        unknownFlag |= op.unknownFlag;
+    }
+
+    // words is the count of not unknown words
+    uint16_t words = SVInt::getNumWords(bits, false);
```
MikePopoloski: It seems like everywhere you use words you're including the effect of unknownFlag. getNumWords(bits, unknownFlag) will just give you that value.

source/SVInt.cpp
```diff
@@ -1690,4 +1699,107 @@ bool exactlyEqual(const SVInt& lhs, const SVInt &rhs) {
...
+    if (words == 1 && !unknownFlag) {
+        // The concatenation still fits into a single word
+        uint16_t offset = 0;
+        uint64_t val = 0;
+        // The first operand writs the msb, therefore, we must operate in reverse
```
MikePopoloski: "writes"

source/SVInt.cpp
```diff
@@ -1690,4 +1699,107 @@ bool exactlyEqual(const SVInt& lhs, const SVInt &rhs) {
...
+        for (const SVInt* op = operands.end() - 1; op >= operands.begin(); --op) {
+            copyBits((uint8_t*)&val, offset, (uint8_t*)&op->val, op->bitWidth);
+            offset += op->bitWidth;
+        }
+        return SVInt(val);
+    }
+    uint64_t *data = new uint64_t[words * (unknownFlag + 1)]();
```
MikePopoloski: Note that the () on the end here causes the data to be zeroed, so no need for the memset.

source/SVInt.cpp
```diff
@@ -1690,4 +1699,107 @@ bool exactlyEqual(const SVInt& lhs, const SVInt &rhs) {
...
+
+    uint16_t offset = 0;
+    // zero out the whole thing ahead of time instead of having to be more careful
+    // and having to set zero bits for missing data
+    memset((void*)data, 0, sizeof(uint64_t) * words * (unknownFlag + 1));
+    for (const SVInt* op = operands.end() - 1; op >= operands.begin(); --op) {
+
+        copyBits((uint8_t*)data, offset,
+            op->isSingleWord() ? (uint8_t*)&op->val
```
MikePopoloski: getRawData() on the SVInt returns a pointer to either val or pVal, depending on whether there is multiple words.

source/SVInt.cpp
drsam94: ah, I didn't read the definition too closely and made too many assumptions that this operator would be something that intuitively made sense.

source/SVInt.cpp
drsam94: Reasonable; that probably would clean up the code and make a little more clear what's going on

source/SVInt.cpp
drsam94: I use it almost everywhere, except in the bounds of the array. Maybe I should just get both values explicitly using getNumWords instead of getting one and inferring the other.

source/SVInt.cpp
drsam94: ah, I see. That's a useful tidbit to know, and kind of makes sense.

source/SVInt.cpp
drsam94: that is very useful

PR #73, opened 2017-02-14
Parse property declarations

evanbryantmiller:


tools/syntax.txt
```diff
@@ -615,6 +619,16 @@ list<VariableDimensionSyntax> dimensions
 UntypedSyntax base=DataTypeSyntax kind=Untyped
 token untyped
 
+PropertyTypeSyntax base=DataTypeSyntax kind=PropertyType
```
MikePopoloski: Rather than have separate types for these, you could just reuse KeywordTypeSyntax, which represents all types that are a single keyword.

tools/syntax.txt
```diff
@@ -1096,6 +1110,40 @@ LetPortListSyntax portList
 EqualsValueClauseSyntax initializer
 token semi
 
+AssertionVariableDeclarationSyntax kind=AssertionVariableDeclaration
```
MikePopoloski: I think you should be able to reuse DataDeclarationSyntax here; they have the same meaning and layout.

tools/syntax.txt
```diff
@@ -1096,6 +1110,40 @@ LetPortListSyntax portList
...
+DataTypeSyntax type
+separated_list<VariableDeclaratorSyntax> assignments
+token semi
+
+PropertyLvarPortDirectionSyntax kind=PropertyLvarPortDirection
+token input
+
+PropertyLocalPortSyntax kind=PropertyLocalPort
+token local
+PropertyLvarPortDirectionSyntax direction
```
MikePopoloski: You don't need a whole separate syntax node type for optional tokens; a default constructed Token() represents one that is empty.

tools/syntax.txt
```diff
@@ -1096,6 +1110,40 @@ LetPortListSyntax portList
...
+token closeParen
+
+PropertyDeclarationSyntax kind=PropertyDeclaration base=MemberSyntax
+token property
+token name
+PropertyPortListSyntax portList
+token semi
+list<AssertionVariableDeclarationSyntax> assertionVariables
+PropertySpecSyntax spec
+EmptyStatementSyntax optionalSemicolon
```
MikePopoloski: Same here.

source/Parser_expressions.cpp
```diff
@@ -33,7 +33,7 @@ ExpressionSyntax* Parser::parseSubExpression(ExpressionOptions::Enum options, in
     int newPrecedence = 0;
 
     auto current = peek();
-    if (current.kind == TokenKind::NewKeyword)
+    if (current.kind == TokenKind::NewKeyword || isClassScope(TokenKind::NewKeyword))
```
MikePopoloski: Rather than having to scan here for class scope, I think we should also check for new expressions in the postfix phase, at which point the class scope will already have been parsed into a NameSyntax.

include/SyntaxNode.h
```diff
@@ -407,6 +411,12 @@ enum class SyntaxKind : uint16_t {
     LetPort,
     LetPortList,
     LetDeclaration,
+    AssertionVariableDeclaration,
```
MikePopoloski: Remember to remove the unused SyntaxKinds. I believe there are warnings in the build about it.

PR #75, opened 2017-02-15
Parse clocking declarations

evanbryantmiller:


tools/syntax.txt
```diff
@@ -542,6 +542,38 @@ token keyword
...
+token globalOrDefault
+token clocking
+token blockName
+token at
+ParenthesizedEventExpressionSyntax event
+token eventIdentifier
+token semi
+list<ClockingItemSyntax> items
+token endClocking
+token colon
```
MikePopoloski: Should use NamedBlockClauseSyntax so that analysis code for that can be unified.

source/Parser.cpp
```diff
@@ -442,6 +442,13 @@ MemberSyntax* Parser::parseMember() {
             return alloc.emplace<EmptyMemberSyntax>(attributes, nullptr, consume());
         case TokenKind::PropertyKeyword:
             return parsePropertyDeclaration(attributes);
+        case TokenKind::GlobalKeyword:
+        case TokenKind::DefaultKeyword:
+            if (peek(1).kind != TokenKind::ClockingKeyword) {
+                break;
+            }
```
MikePopoloski: Don't rely on fallthrough, it's confusing.

source/Parser.cpp
```diff
@@ -1907,6 +1914,125 @@ ParameterDeclarationSyntax* Parser::parseParameterPort() {
     return alloc.emplace<ParameterDeclarationSyntax>(Token(), type, parseOneVariableDeclarator());
 }
 
+ClockingSkewSyntax* Parser::parseClockingSkew() {
+    Token edge, hash;
+    ExpressionSyntax* value = nullptr;
+    if (peek().kind == TokenKind::EdgeKeyword || peek().kind == TokenKind::PosEdgeKeyword || peek().kind == TokenKind::NegEdgeKeyword) {
```
MikePopoloski: There's a peek() method that takes a TokenKind and returns a bool which would make this easier.

source/Parser.cpp
```diff
@@ -1907,6 +1914,125 @@ ParameterDeclarationSyntax* Parser::parseParameterPort() {
...
+        edge = consume();
+    }
+    if (peek().kind == TokenKind::Hash) {
+        hash = consume();
+        switch(peek().kind) {
+            case TokenKind::OpenParenthesis:
+            {
```
MikePopoloski: Braces on previous line.

source/Parser.cpp
```diff
@@ -1907,6 +1914,125 @@ ParameterDeclarationSyntax* Parser::parseParameterPort() {
...
+    Token semi = expect(TokenKind::Semicolon);
+    SmallVectorSized<ClockingItemSyntax*, 4> buffer;
+    if (globalOrDefault.kind != TokenKind::GlobalKeyword) {
+        while(peek().kind != TokenKind::EndClockingKeyword && peek().kind != TokenKind::EndOfFile) {
+            Token defaultKeyword, inputKeyword, outputKeyword;
+            ClockingDirectionSyntax* direction;
+            ClockingSkewSyntax* inputSkew = nullptr, *outputSkew = nullptr;
+            MemberSyntax* declaration;
+            switch(peek().kind) {
+                case TokenKind::DefaultKeyword:
```
MikePopoloski: The use of fallthrough in this switch statement is pretty confusing.

source/Parser.cpp
```diff
@@ -1907,6 +1914,125 @@ ParameterDeclarationSyntax* Parser::parseParameterPort() {
...
+            auto closeParen = expect(TokenKind::CloseParenthesis);
+            event = alloc.emplace<ParenthesizedEventExpressionSyntax>(openParen, innerExpr, closeParen);
+        } else {
+            eventIdentifier = expect(TokenKind::Identifier);
+        }
+    }
+    Token semi = expect(TokenKind::Semicolon);
+    SmallVectorSized<ClockingItemSyntax*, 4> buffer;
+    if (globalOrDefault.kind != TokenKind::GlobalKeyword) {
+        while(peek().kind != TokenKind::EndClockingKeyword && peek().kind != TokenKind::EndOfFile) {
```
MikePopoloski: You have to be careful with this kind of loop; if the endclocking keyword is missing then it will loop forever.

evanbryantmiller: What should I do about this then? Should I include other tokens to check against here?

MikePopoloski: You can look at functions like parseMemberList (in fact, you might be able to just use that directly). It does a lot of work to gracefully handle errors.

MikePopoloski: parseBlockItems might be a better example in that it also checks isEndKeyword(), which means if the user forgot the endclocking but has some other end token in their file (like endmodule or whatever) we won't just suck up every token.

PR #79, opened 2017-02-15
CXXFLAGS: turn on sanitization, turn off optimisation

Baranowski:


Baranowski:
Nah, it seems impossible to do sanitisation without memleak detection, and we leak a lot which causes endless error reports.

MikePopoloski:
Where do we leak memory? As far as I know everything should be cleaned up when our allocators are destructed.

Baranowski:
See the build report in this pull request

PR #80, opened 2017-02-15
Parse up to 64 bit integral literals with no length given

evanbryantmiller:


MikePopoloski:
What is the reason for this change? This won't actually work correctly for all cases I don't think, and I'm pretty sure the language rules say that a plain integer is 32-bit.

drsam94:
Yeah, the spec says

"The number of bits that make up an unsized number (which is a simple decimal number or a number with a base specifier but no size specification) shall be at least 32."

i.e I think the proper size is max(32, minNumberOfBitsForLiteral)

evanbryantmiller:
I made this change since we couldn't parse the integers in ARPHandler_tb.sv.  I didn't bother checking with the spec to see if it was valid.

drsam94:
So yeah we shouldn't treat it as errors. The size of a constant <= 2**32 - 1 should be stored in 32 bits, and other things should be sized as tightly as possible, i.e log2(bits)

evanbryantmiller:
I left decimal integers alone at 32 integers.  Now I've only extended vector literals to necessarily be large enough.  Do we want a hard cap for these vector literals that don't explicitly specify a size?

source/VectorBuilder.cpp
```diff
@@ -166,11 +166,15 @@ SVInt VectorBuilder::finish() {
             bits += clog2(digits[0].value + 1);
 
         if (bits > sizeBits) {
-            diagnostics.add(DiagCode::VectorLiteralOverflow, firstLocation);
-            return SVInt(0);
+            if (sizeBits || bits > SVInt::MAX_BITS) {
+                diagnostics.add(DiagCode::VectorLiteralOverflow, firstLocation);
+                return SVInt(0);
+            }
+            return SVInt(bits <= 32 ? 32 : bits, literalBase, signFlag, hasUnknown, ArrayRef<logic_t>(digits.begin(), digits.count()));
```
drsam94: just a nit that I think that this would be clearer as `std::max(32, bits)`

PR #81, opened 2017-02-15
Added support for bit select expressions

drsam94:
This adds support for all kinds of bit select expressions (notably not part select expressions and other things that require us to have a good implementation of unpacked array types). While writing this, it was also necessary to fix some bugs in the SVInt(StringRef) constructor, in concatenation, and slightly tweak the syntax for ItemSelectExpressions, which I think potentially may need to be tweaked further later.


source/SVInt.cpp
```diff
@@ -561,6 +570,68 @@ SVInt SVInt::replicate(const SVInt& times) const {
     return concatenate(ArrayRef<SVInt>(buffer.begin(), buffer.end()));
 }
 
+SVInt SVInt::bitSelect(int16_t lsb, int16_t msb) const {
+    if (msb < lsb) {
```
MikePopoloski: Should probably just assert here, no?

drsam94: I think so: I think that the only cases where this could happen we should (but obviously don't now) detect it before it gets here.

PR #82, opened 2017-02-15
make macro stringification more robust

drsam94:
This change makes us take all whitespace in a macro definition that we are stringifying literally, which is what all other preprocessors I have tested do. This includes having to include any whitespace before an actual argument of no tokens.

This addresses Issue #3

### DIFF
--- a/include/Trivia.h
+++ b/include/Trivia.h
@@ -49,7 +49,9 @@ public:
     /// If this trivia is tracking a skipped syntax node, return that now.
     SyntaxNode* syntax() const;
 
-    StringRef getRawText() const {
+    /// Get the raw whitespace of some trivia
+    StringRef getRawWhitespace() const {
+        ASSERT(kind == TriviaKind::Whitespace);
         return rawText;
     }
 private:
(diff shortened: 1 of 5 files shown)


include/Trivia.h
```diff
@@ -49,6 +49,9 @@ class Trivia {
     /// If this trivia is tracking a skipped syntax node, return that now.
     SyntaxNode* syntax() const;
 
+    StringRef getRawText() const {
+        return rawText;
```
MikePopoloski: Should ASSERT on the kind here.

source/Preprocessor.cpp
```diff
@@ -920,6 +923,11 @@ bool Preprocessor::expandMacro(DefineDirectiveSyntax* macro, Token usageSite, Ma
                 if (begin != end) {
                     dest.append(begin->withTrivia(alloc, token.trivia()));
                     dest.appendRange(++begin, end);
+                } else {
+                    // the macro argument contained no tokens.
+                    // we still need to supply an empty token here to ensure
+                    // that the trivia of the formal paremter is passed on
+                    dest.append(*alloc.emplace<Token>(TokenKind::EmptyMacroArgument, token.getInfo()));
```
MikePopoloski: I don't understand the point of heap allocating and then immediately dereferencing.

drsam94: you shouldn't, this was dumb.

PR #88, opened 2017-02-16
Fixed literal parsing corner cases

drsam94:
In particular this change:

- Combines a lot of the SVInt(StringRef) and SVInt(...., ArrayRef<logic_t>) code for simplicity/ code reuse
- Due to ^, propagates my previous fix for literals with unknown bits to the other constructor
- Adds a fast path for constructing SVInts from vectors that fit in one word
- Adds support for "unknown padding" of literals as described in the spec, i.e `3'bx1 == 3'bxx1`
- Adds support for literal truncation, and downgrades overflow from an error to a warning
- Cleans up some related code

### DIFF
--- a/include/SVInt.h
+++ b/include/SVInt.h
@@ -24,6 +24,8 @@ enum class LiteralBase : uint8_t {
     Hex
 };
 
+bool literalBaseFromChar(char base, LiteralBase& result);
+
 /// Represents a single 4-state bit. The usual bit values of 0 and 1 are
 /// augmented with X (unknown) and Z (high impedance). Both X and Z are
 /// considered "unknown" for most computation purposes.
--- a/source/SVInt.cpp
+++ b/source/SVInt.cpp
@@ -18,6 +18,29 @@ namespace slang {
 const logic_t logic_t::x = logic_t(logic_t::X_VALUE);
 const logic_t logic_t::z = logic_t(logic_t::Z_VALUE);
 
+bool literalBaseFromChar(char base, LiteralBase& result) {
+    switch (base) {
+        case 'd':
+        case 'D':
+            result = LiteralBase::Decimal;
+            return true;
+        case 'b':
+        case 'B':
+            result = LiteralBase::Binary;
+            return true;
+        case 'o':
+        case 'O':
+            result = LiteralBase::Octal;
+            return true;
+        case 'h':
+        case 'H':
+            result = LiteralBase::Hex;
+            return true;
+        default:
+            return false;
+    }
+}
+
 SVInt::SVInt(StringRef str) {
     ASSERT(str);
 
@@ -26,6 +49,14 @@ SVInt::SVInt(StringRef str) {
     if (*c == '-' || *c == '+') {
         c++;    // heh
         ASSERT(c != str.end(), "String only has a sign?");
+    }
+
+    unknownFlag = false;
+    for (const char* tmp = c; tmp != str.end(); tmp++) {
+        if (isLogicDigit(*tmp)) {
+            unknownFlag = true;
+            break;
+        }
     }
 
     // look for a base specifier (optional)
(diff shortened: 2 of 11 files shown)


source/SVInt.cpp
```diff
@@ -86,146 +108,64 @@ SVInt::SVInt(StringRef str) {
...
-                digit.val = value;
-            else
-                digit.pVal[0] = value;
-            *this += digit;
-        }
+    if (!isSingleWord()) {
+        // We use the assignment operator to return the result of the other constructor,
+        // but the assignment operator will try to delete our pVal, so let's make sure
+        // there is something there
+        pVal = new uint64_t[0]();
```
MikePopoloski: Deleting a null pointer is fine, there's no reason to waste time allocating here.

source/SVInt.cpp
```diff
@@ -86,146 +108,64 @@ SVInt::SVInt(StringRef str) {
...
+    }
+    std::string s= "";
```
MikePopoloski: Debug code?

source/SVInt.cpp
```diff
@@ -237,39 +177,76 @@ SVInt::SVInt(uint16_t bits, LiteralBase base, bool isSigned, bool anyUnknown, Ar
...
-        default:
-            ASSERT(false, "Unreachable");
-            break;
+
+        DEFAULT_UNREACHABLE;
+    }
+
+    if (isSingleWord()) {
+        // Fast path for values that fit in one word:
+        // let's do the computation entirely in a normal uin64_t instead
```
MikePopoloski: "uint64_t"

source/SVInt.cpp
```diff
@@ -237,39 +177,76 @@ SVInt::SVInt(uint16_t bits, LiteralBase base, bool isSigned, bool anyUnknown, Ar
...
 
         // shift if we can, multiply if we must
+        // this operation can result in us losing our unknown flag, so let's
+        // make sure that doesn't happen
+        bool unknownFlagStore = unknownFlag;
+        int bitWidthStore = bitWidth;
         if (shift)
             *this = shl(shift, false);
         else
             *this *= radixSv;
```
MikePopoloski: I think it's probably worth in the future having an internal mul() method that, like shl(), takes a flag indicating whether the unknown flag should go away. That way we don't have to do this extra checking here.

source/SVInt.cpp
```diff
@@ -237,39 +177,76 @@ SVInt::SVInt(uint16_t bits, LiteralBase base, bool isSigned, bool anyUnknown, Ar
...
             unknown = ones;
+        } else {
+            ASSERT(value < radix, "Digit %d too large for radix %d", value, radix);
         }
 
         // shift if we can, multiply if we must
+        // this operation can result in us losing our unknown flag, so let's
+        // make sure that doesn't happen
+        bool unknownFlagStore = unknownFlag;
+        int bitWidthStore = bitWidth;
```
MikePopoloski: bitWidthStore is unused?

include/SVInt.h
```diff
@@ -24,6 +24,8 @@ enum class LiteralBase : uint8_t {
     Hex
 };
 
+bool literalBaseFromChar(char base, LiteralBase& result);
```
drsam94: hmm, wasn't aware of the existence of std::optional. I think I will just leave it as the best solution is probably to use a smart enum with INVALID

source/SVInt.cpp
drsam94: something was seg faulting somewhere...I will go back and see what the exact situation is now.

source/SVInt.cpp
```diff
@@ -28,6 +51,14 @@ SVInt::SVInt(StringRef str) {
         ASSERT(c != str.end(), "String only has a sign?");
     }
 
+    unknownFlag = false;
```
drsam94: good idea

source/SVInt.cpp
drsam94: yep...

source/SVInt.cpp
drsam94: yeah, that was me not cleaning up after a failed idea

source/SVInt.cpp
drsam94: I'll add a TODO

PR #90, opened 2017-02-16
Instance arrays

denisrr:


source/Diagnostics.cpp
```diff
@@ -155,6 +155,8 @@ DiagnosticWriter::DiagnosticWriter(SourceManager& sourceManager) :
     descriptors[DiagCode::UnknownModule] = { "unknown module named '{}'", DiagnosticSeverity::Error };
     descriptors[DiagCode::LocalParamNoInitializer] = { "local parameter is missing an initializer", DiagnosticSeverity::Error };
     descriptors[DiagCode::BodyParamNoInitializer] = { "parameters declaration is missing an initializer", DiagnosticSeverity::Error };
+    descriptors[DiagCode::UnpackedDimensionRequired] = { "unpacked dimension is required in array declaration", DiagnosticSeverity::Error };
+    descriptors[DiagCode::UnpackedDimensionRequiresConstRange] = { "unpacked dimension requires a constant range '{}'", DiagnosticSeverity::Error };
```
MikePopoloski: You don't need the {} in your error string since you aren't giving it any parameters to replace. The SourceRange you give it doesn't get put into the string, it gets used to underline the part of the source text that gets displayed underneath.

PR #91, opened 2017-02-16
Added support for $bits system function

drsam94:
This only works for $bits(expression), not $bits(data_type)


source/SemanticModel.cpp
```diff
@@ -67,6 +67,14 @@ SemanticModel::SemanticModel(BumpAllocator& alloc, Diagnostics& diagnostics, Dec
 
     args.append(alloc.emplace<FormalArgumentSymbol>(intType));
     systemScope.add(alloc.emplace<SubroutineSymbol>("$clog2", intType, args.copy(alloc), SystemFunction::clog2));
+
+    // Assume input type has no width, so that the argument's self-determined type won't be expanded due to the
```
MikePopoloski: Yeah, I was thinking we would probably add a special "AnyWidth" integral type for these system functions to use. Since we don't warn about widths right now it's no big deal.

PR #93, opened 2017-02-16
WiP - REPL

Baranowski:
This is still work in progress. Feel free to ignore or comment on the way this is going.


tools/shell.cpp
```diff
@@ -0,0 +1,61 @@
...
+    el_set(el, EL_HIST, history, cmdHistory);
+
+    int charsRead;
+    const char *line;
+
+    /*
+    BumpAllocator alloc;
+    Diagnostics diagnostics;
+    SourceManager& srcMgr = SyntaxTree::getDefaultSourceManager();
+    Preprocessor preprocessor(srcMgr, alloc, diagnostics);
```
drsam94: yeah, the preprocessor seems tricky to get working here: I think the ideal thing would be to add the ability for the ScriptSession to maintain preprocessor stain between evaluations, otherwise this would probably be hard.

tools/shell.cpp
```diff
@@ -0,0 +1,61 @@
...
+        /*
+        preprocessor.pushSource(srcMgr.assignText(line));
+        Parser parser(preprocessor);
+        */
+        try {
+            auto value = session.evalWithKind(line);
+            printf("Detected kind: %s\n", get<1>(value).c_str());
+            const SVInt *integer = get_if<SVInt>(&get<0>(value));
+            if (integer) {
+                printf("Integer value: %d\n", integer->getAssertUInt16());
```
drsam94: I think we should add for printing something like

~~~
if (!integer->isSingleWord())
  printf(... integer->toString(LiteralBase::Binary).c_str() ... );
~~~
so that you can see the results of more expressions

tools/shell.cpp
```diff
@@ -0,0 +1,61 @@
...
+            }
+        } catch (Diagnostics &diags) {
+            printf("Errors when parsing:\n");
```
drsam94: Just a meaning error, this can be an "Error when parsing" but can also be an error when evaluating or doing any number of other parts of the process, right?

tools/shell.cpp
drsam94: Also isSingleWord() doesn't catch the negative case: there is getAssertInt64() that will handle anything that fits in an int64_t, which is probably good enough for most things in the demo

tools/shell.cpp
Baranowski: I don't think it's the job of shell to do preprocessing. Usually preprocessor and compiler/interpreter are separate tools, right?

tools/shell.cpp
Baranowski: Those are only parsing errors

Baranowski:
This is ready, @MikePopoloski please review

build/linux/Makefile
```diff
@@ -12,23 +12,24 @@ SLANG_SOURCES:=$(wildcard $(TOP)/source/*.cpp) $(wildcard $(TOP)/external/ppk_as
...
 
 FMTLIB_OBJECTS:=$(patsubst $(TOP)/%.cc,obj/%.o,$(FMTLIB_SOURCES))
 SLANG_OBJECTS:=$(patsubst $(TOP)/%.cpp,obj/%.o,$(SLANG_SOURCES))
 UNITTEST_OBJECTS:=$(patsubst $(TOP)/%.cpp,obj/%.o,$(UNITTEST_SOURCES))
 FILETEST_OBJECTS:=$(patsubst $(TOP)/%.cpp,obj/%.o,$(FILETEST_SOURCES))
 REWRITER_OBJECTS:=$(patsubst $(TOP)/%.cpp,obj/%.o,$(REWRITER_SOURCES))
+SHELL_OBJECTS:=$(patsubst $(TOP)/%.cpp,obj/%.o,$(SHELL_SOURCES))
 
-ALL_CPP_OBJS:=$(SLANG_OBJECTS) $(UNITTEST_OBJECTS) $(FILETEST_OBJECTS) $(REWRITER_OBJECTS)
-ALL_OBJECTS:=$(FMTLIB_OBJECTS) $(ALL_CPP_OBJS)
```
MikePopoloski: Did the ALL_OBJECTS variable get deleted?

include/SyntaxTree.h
```diff
@@ -40,16 +40,16 @@ class SyntaxTree {
     }
 
     /// Creates a syntax tree by guessing at what might be in the given source snippet.
-    static SyntaxTree fromText(StringRef text) {
-        return fromText(text, getDefaultSourceManager());
+    static SyntaxTree fromText(StringRef text, bool throwDiagnostics = false) {
```
MikePopoloski: I don't think this should be added to this API... the caller can easily check if diagnostics().empty() is false and choose to throw if it wants.

tools/shell.cpp
```diff
@@ -0,0 +1,84 @@
+#include "slang.h"
+#include <histedit.h>
```
MikePopoloski: It looks like the travis build is failing because it can't find histedit.

Baranowski:
Looks like this only builds on mac

tools/shell.cpp
denisrr: apt-get install libedit-dev

MikePopoloski:
Closing this out... too much has changed at this point and it was mostly for our demo last year anyway.

PR #99, opened 2017-02-17
Parse DPI Import/Export declarations

evanbryantmiller:


include/SourceLocation.h
```diff
@@ -57,6 +58,9 @@ class SourceLocation {
     /// Note that there is no error checking to ensure that the location
     /// still points to a valid place in the source.
     SourceLocation operator+(int delta) const {
+        if (bufferID.getValue() == 6) {
```
MikePopoloski: What is this change for?

source/Parser.cpp
```diff
@@ -1837,6 +1842,27 @@ PackageImportItemSyntax* Parser::parsePackageImportItem() {
...
+    Token property, name, equals;
+    if (keyword.kind == TokenKind::ImportKeyword && (peek(TokenKind::ContextKeyword) || peek(TokenKind::PureKeyword))) {
+        property = consume();
+    }
+    if (peek(TokenKind::Identifier)) {
+        name = consume();
+        equals = expect(TokenKind::Equals);
+    }
+    if (property.kind == TokenKind::PureKeyword && !peek(TokenKind::FunctionKeyword)) {
+        addError(DiagCode::ExpectedFunctionKeyword, peek().location());
```
MikePopoloski: I don't think we should have a separate error for this. The expect() function can already build a good error for missing keywords, we should try to use that in some way. Also if there isn't a function or task here, parseFunctionPrototype() will just take whatever token happens to be there and pretend it's the keyword, which seems bad.

include/SourceLocation.h
evanbryantmiller: Oops. I was using this for debugging since this bug somehow caused the location to point to an invalid location in the source. I'll remove it.

source/Parser.cpp
evanbryantmiller: Good point.

PR #133, opened 2017-08-30
Add docker and gcc 7 instructions to README

martinda:
Adding docker instructions.


MikePopoloski:
Hi, thanks for the interest in the project. I have nothing against people doing Docker builds, but I don't think the README is the right place for putting build instructions for every possible build system or environment someone may want to use. The basic instructions of generating project files and running a build are there, and people can adapt them as necessary to their own environment.

PR #173, opened 2018-08-13
GCC Compile fix, Regressions tests and gitignore update

pleroux0:
I didn't see a contributor.txt or something similar so I will state here:

All my contributions under the slang project are also under the MIT license.

I found this project an hour ago. It looks very promising.

There was no regression test system in place so I just made a folder with a CMakeLists.txt

I tried a few module descriptions, but it seems to run into a few internal errors. I've simplified it as much as possible and put the two failing cases in regression tests.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 .vs/
 build/
+compile_commands.json
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -126,4 +126,8 @@ conan_cmake_run(CONANFILE conanfile.txt BASIC_SETUP CMAKE_TARGETS)
 
 add_subdirectory(source)
 add_subdirectory(tools)
+
+include(CTest)
+
 add_subdirectory(tests/unittests)
+add_subdirectory(tests/regression)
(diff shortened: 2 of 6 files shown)


pleroux0:
I see the GCC compile fix is fixed in the mean time. The pull request still contains the travis fix where it builds the correct branch instead of just master. As well as the regression tests. I could seperate them into seperate pull request if you are interested.

MikePopoloski:
Hi, thanks for the pull request! Splitting into two pieces would be good:
- I'm happy to take the travis and gitignore changes as-is.
- I had planned on having regression test system similar to what you've proposed here; I've held off so far because:
  1. It's still fairly early days; as you've seen there's lots of features I simply don't support yet.
  2. I expect I'll need some more comprehensive tool like the "lit" tool that LLVM uses to run the compiler and compare against expected results.
Maybe it's fine to start with what you've shown here and move to something more complicated later.

I'm currently focusing my efforts on fixing / improving more fundamental parts of elaborating the hierarchy. I can't promise I'll get to the features shown in the failing tests anytime soon.

pleroux0:
Superseded by #174 and #175 

For unimplemented features maybe saying "unimplemented" instead of "internal compiler error" might be more descriptive for people checking out the project.

I'm in no rush to see them fixed. I appreciate the effort and look forward to see where this project goes.

PR #177, opened 2018-12-05
Add IdentifierSelectName to avoid exception, cosmetic newline added

jrrk:
The tool throws an exception if it encounters IdentifierSelectName in DownwardLookupResult.


MikePopoloski:
Thanks for the pull request. I'm not going to merge as-is, because the reason that case was left out was as a reminder to eventually go add full select support there. This diff doesn't add that support, so it'd be much easier for it to be silently forgotten. Also a diff like this should typically include unit tests to exercise the change.

MikePopoloski:
Superseded by other changes.

Issue #257, opened 2019-06-16
Add a CMake flag to disable tests


Kuree:
Hello,

I'm using `slang` as a git submodule (using `add_subdirectory`). Is it possible to add a variable/flag in CMake to disable adding all the tests? Thanks.

MikePopoloski:
Sure, I can do that, or I'd welcome a pull request if you want to add it.

I haven't spent any time trying to consume slang as a library via cmake yet, so if you have changes that would make it easier please feel free to propose them.

PR #258, opened 2019-06-17
add option to turn off tests

Kuree:
See https://github.com/MikePopoloski/slang/issues/257.

This option is off by default.


Kuree:
No idea why `Install Conan (Linux)` failed on azure pipeline.

MikePopoloski:
Microsoft broke Python on their Azure images. Pretty annoying, but not related to your diff.

PR #260, opened 2019-06-27
add proper dependencies

Kuree:
When using debug/dev mode in cmake, `Version.cpp` somehow won't get generated. This patch fixes the dependency.


MikePopoloski:
Weird that this is necessary, it seems like cmake has all the info it needs to infer the dependency, but no harm in specifying it explicitly in this case.

PR #262, opened 2019-08-14
Source manager memory allocation fix

### LINKED ISSUE #261: Parsing failure
It faults when I try to parse the VQM (verilog subset) from titan_release_1.3.0/benchmarks/titan23/directrf/netlists/directrf_stratixiv.vqm

The source is at http://www.eecg.utoronto.ca/~kmurray/titan.html

I tried ../build/bin/rewriter .directrf_stratixiv.vqm
and seg-faults

tgorochowik:
Casting the total size to u32_t effectively limits the max files size. This can cause a SEGFAULT if the input file is larger than 2**32 B.
    
Fixes #261

Additionally this PR adds some sanity checks to the rewriter tool.

### DIFF
--- a/source/text/SourceManager.cpp
+++ b/source/text/SourceManager.cpp
@@ -432,7 +432,7 @@ bool SourceManager::readFile(const fs::path& path, std::vector<char>& buffer) {
         return false;
 
     // + 1 for null terminator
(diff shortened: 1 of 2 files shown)


MikePopoloski:
Nice, thanks for the fix.

PR #270, opened 2019-10-12
FindPython requires CMake 3.12

toddstrader:
FindPython appears to have been added in CMake 3.12:
https://cmake.org/cmake/help/v3.12/release/3.12.html?highlight=findpython#modules


MikePopoloski:
Indeed, thanks!

PR #272, opened 2019-10-28
Fix compilation with threads

tgorochowik:
This fix is needed for the code to compile in Travis which uses an old Ubuntu version (16.04).

Refer to https://github.com/SymbiFlow/conda-packages/pull/52 for more information about the issue.

This does not seem affect the compilation on a modern OS.

Please let me know if you think this should be put in a different place or if it should be solved differently.


MikePopoloski:
Seems straightforward to me, thanks for the PR!

PR #294, opened 2020-02-29
Add serialization for statements

Kuree:
This PR adds serialization (JSON) to statements. The `name` for each `ASTSerializer.write()` call is consistent with the class field name. 

Minor API changes:
- I added `startObject` and `endObject` to `ASTSerializer` since in the case statement, each item is a key-value pair.
- I refactored the enums in the case statement to the top level so that they can have `toString` generated automatically.

Please let me know if you have any code-review related questions.

### DIFF
--- a/source/binding/Statements.cpp
+++ b/source/binding/Statements.cpp
@@ -14,6 +14,7 @@
 #include "slang/diagnostics/NumericDiags.h"
 #include "slang/diagnostics/StatementsDiags.h"
 #include "slang/parsing/LexerFacts.h"
+#include "slang/symbols/ASTSerializer.h"
 #include "slang/symbols/ASTVisitor.h"
 #include "slang/syntax/AllSyntax.h"
 
@@ -54,6 +55,11 @@ struct VerifyVisitor {
 namespace slang {
 
 const InvalidStatement InvalidStatement::Instance(nullptr);
+
+void InvalidStatement::serializeTo(ASTSerializer& serializer) const {
+    if (child)
+        serializer.write("child", *child);
+}
 
 ER Statement::eval(EvalContext& context) const {
     EvalVisitor visitor;
@@ -528,8 +534,21 @@ bool StatementList::verifyConstantImpl(EvalContext& context) const {
     return true;
 }
 
+void StatementList::serializeTo(ASTSerializer& serializer) const {
+    serializer.startArray("list");
+    for (auto const& stmt : list) {
+        serializer.serialize(*stmt);
+    }
+    serializer.endArray();
+}
+
 BlockStatement::BlockStatement(const StatementBlockSymbol& block, SourceRange sourceRange) :
     Statement(StatementKind::Block, sourceRange), blockKind(block.blockKind), block(&block) {
+}
+
+void BlockStatement::serializeTo(ASTSerializer& serializer) const {
+    serializer.write("blockKind", toString(blockKind));
+    serializer.write("list", *list);
 }
 
 Statement& BlockStatement::fromSyntax(Compilation& compilation, const BlockStatementSyntax& syntax,
@@ -628,6 +647,10 @@ bool ReturnStatement::verifyConstantImpl(EvalContext& context) const {
     return !expr || expr->verifyConstant(context);
 }
 
+void ReturnStatement::serializeTo(ASTSerializer& serializer) const {
+    serializer.write("expr", *expr);
+}
+
 Statement& BreakStatement::fromSyntax(Compilation& compilation, const JumpStatementSyntax& syntax,
                                       const BindContext& context, StatementContext& stmtCtx) {
     auto result = compilation.emplace<BreakStatement>(syntax.sourceRange());
@@ -684,6 +707,10 @@ bool VariableDeclStatement::verifyConstantImpl(EvalContext& context) const {
             return false;
     }
     return true;
+}
+
+void VariableDeclStatement::serializeTo(ASTSerializer& serializer) const {
+    serializer.write("symbol", symbol);
 }
 
 Statement& ConditionalStatement::fromSyntax(Compilation& compilation,
@@ -757,6 +784,12 @@ bool ConditionalStatement::verifyConstantImpl(EvalContext& context) const {
     return true;
 }
 
+void ConditionalStatement::serializeTo(ASTSerializer& serializer) const {
+    serializer.write("cond", cond);
+    serializer.write("ifTrue", ifTrue);
+    serializer.write("ifFalse", ifFalse);
+}
+
 Statement& CaseStatement::fromSyntax(Compilation& compilation, const CaseStatementSyntax& syntax,
                                      const BindContext& context, StatementContext& stmtCtx) {
     bool bad = false;
@@ -765,34 +798,34 @@ Statement& CaseStatement::fromSyntax(Compilation& compilation, const CaseStateme
         bad = true;
     }
 
-    Condition condition;
+    CaseStatementCondition condition;
     switch (syntax.caseKeyword.kind) {
         case TokenKind::CaseKeyword:
-            condition = Condition::Normal;
+            condition = CaseStatementCondition::Normal;
             break;
         case TokenKind::CaseXKeyword:
-            condition = Condition::WildcardXOrZ;
+            condition = CaseStatementCondition::WildcardXOrZ;
             break;
         case TokenKind::CaseZKeyword:
-            condition = Condition::WildcardJustZ;
+            condition = CaseStatementCondition::WildcardJustZ;
             break;
         default:
             THROW_UNREACHABLE;
     }
 
-    Check check;
+    CaseStatementCheck check;
     switch (syntax.uniqueOrPriority.kind) {
         case TokenKind::Unknown:
-            check = Check::None;
+            check = CaseStatementCheck::None;
             break;
         case TokenKind::UniqueKeyword:
-            check = Check::Unique;
+            check = CaseStatementCheck::Unique;
             break;
         case TokenKind::Unique0Keyword:
-            check = Check::Unique0;
+            check = CaseStatementCheck::Unique0;
             break;
         case TokenKind::PriorityKeyword:
-            check = Check::Priority;
+            check = CaseStatementCheck::Priority;
             break;
         default:
             THROW_UNREACHABLE;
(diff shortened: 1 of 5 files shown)


source/binding/Statements.cpp
```diff
@@ -528,10 +534,23 @@ bool StatementList::verifyConstantImpl(EvalContext& context) const {
...
+    serializer.endArray();
+}
+
 BlockStatement::BlockStatement(const StatementBlockSymbol& block, SourceRange sourceRange) :
     Statement(StatementKind::Block, sourceRange), blockKind(block.blockKind), block(&block) {
 }
 
+void BlockStatement::serializeTo(ASTSerializer& serializer) const {
+    serializer.write("blockKind", toString(blockKind));
+    serializer.write("list", *list);
```
MikePopoloski: Only one of the block or list members will be non-null, so you need to check which one here. Instead, I would recommend calling this "body" and setting it to getStatements() which will do that check for you.

The reason for this split is because some block statements in Verilog create a new level of hierarchy, in which case there will be a BlockSymbol here, and some do not, in which case this will just be a list of statements.

source/binding/Statements.cpp
```diff
@@ -628,6 +647,10 @@ bool ReturnStatement::verifyConstantImpl(EvalContext& context) const {
     return !expr || expr->verifyConstant(context);
 }
 
+void ReturnStatement::serializeTo(ASTSerializer& serializer) const {
+    serializer.write("expr", *expr);
```
MikePopoloski: expr can be null, if this is a return statement in a void function.

source/binding/Statements.cpp
```diff
@@ -686,6 +709,10 @@ bool VariableDeclStatement::verifyConstantImpl(EvalContext& context) const {
     return true;
 }
 
+void VariableDeclStatement::serializeTo(ASTSerializer& serializer) const {
+    serializer.write("symbol", symbol);
```
MikePopoloski: I think you want to use writeLink() here instead, because this VariableDeclStatement is referencing a variable that is owned by the block symbol. The VariableDeclStatement is a way of saying "this variable gets initialized at this point relative to other statements in the block". Using write() instead of writeLink() will cause the full definition of the symbol to show up in the output twice.

source/symbols/ASTSerializer.cpp
```diff
@@ -91,6 +109,16 @@ void ASTSerializer::visit(const T& elem) {
             elem.serializeTo(*this);
         }
 
+        writer.endObject();
+    } else if constexpr(std::is_base_of_v<Statement, T>) {
+        (void)(elem);
```
MikePopoloski: You shouldn't need this void cast, you reference the element below.

source/binding/Statements.cpp
```diff
@@ -1083,6 +1137,24 @@ bool ForLoopStatement::verifyConstantImpl(EvalContext& context) const {
...
 }
 
+void ForLoopStatement::serializeTo(ASTSerializer &serializer) const {
+    serializer.startArray("initializers");
+    for (auto const &ini: initializers) {
+        serializer.serialize(*ini);
+    }
+    serializer.endArray();
+
+    serializer.write("stopExpr", *stopExpr);
```
MikePopoloski: stopExpr can be null.

source/binding/Statements.cpp
```diff
@@ -995,6 +1028,27 @@ bool CaseStatement::verifyConstantImpl(EvalContext& context) const {
...
+        serializer.startArray("expressions");
+        for (auto const &ex: item.expressions) {
+            serializer.serialize(*ex);
+        }
+        serializer.endArray();
+        serializer.startObject("stmt");
+        serializer.serialize(*item.stmt);
+        serializer.endObject();
+    }
+    // default case
```
MikePopoloski: This comment isn't saying much that isn't already evident in the code.

source/binding/Statements.cpp
```diff
@@ -995,6 +1028,27 @@ bool CaseStatement::verifyConstantImpl(EvalContext& context) const {
...
+    serializer.write("condition", toString(condition));
+    serializer.write("check", toString(check));
+    serializer.startArray("items");
+    for (auto const &item: items) {
+        serializer.startArray("expressions");
+        for (auto const &ex: item.expressions) {
+            serializer.serialize(*ex);
+        }
+        serializer.endArray();
+        serializer.startObject("stmt");
```
MikePopoloski: I don't think this will work, it's creating a property inside of an array. Instead I think you want each element in the item array to itself be an object, with one property being the expressions array and the other being the stmt.

source/binding/Statements.cpp
```diff
@@ -757,6 +784,12 @@ bool ConditionalStatement::verifyConstantImpl(EvalContext& context) const {
     return true;
 }
 
+void ConditionalStatement::serializeTo(ASTSerializer& serializer) const {
+    serializer.write("cond", cond);
+    serializer.write("ifTrue", ifTrue);
+    serializer.write("ifFalse", ifFalse);
```
MikePopoloski: Hmm, I'm a little surprised this compiles, ifFalse is a pointer.

Also you'll have to check ifFalse for null. The general practice throughout the library is that if it's a pointer that's not in a list and not wrapped in `not_null<>`, it can potentially be null.

source/symbols/ASTSerializer.cpp
Kuree: Sorry that was used to allow compilation without implementing the `serializedTo`. I forgot to remove it.

source/binding/Statements.cpp
```diff
@@ -995,6 +1033,32 @@ bool CaseStatement::verifyConstantImpl(EvalContext& context) const {
     return true;
 }
 
+void CaseStatement::serializeTo(ASTSerializer& serializer) const {
+    serializer.write("condition", toString(condition));
+    serializer.write("check", toString(check));
+    serializer.startArray("items");
+    for (auto const& item : items) {
+        serializer.startObject("item");
```
MikePopoloski: This is still not quite correct. If you run this code you'll see that it produces JSON like items=[item={}], but that's not valid JSON. Array elements shouldn't have property names. Instead for each item you just want to start an object, with no property name.

source/binding/Statements.cpp
```diff
@@ -528,10 +534,26 @@ bool StatementList::verifyConstantImpl(EvalContext& context) const {
     return true;
 }
 
+void StatementList::serializeTo(ASTSerializer& serializer) const {
+    serializer.startArray("list");
+    for (auto const& stmt : list) {
+        if (stmt)
```
MikePopoloski: You don't need to check list items for null, we basically don't ever put null items into lists. Also consider just using plain `auto` when the item is a pointer. There's no reason to take a reference to the pointer.

source/binding/Statements.cpp
```diff
@@ -995,6 +1033,32 @@ bool CaseStatement::verifyConstantImpl(EvalContext& context) const {
...
+void CaseStatement::serializeTo(ASTSerializer& serializer) const {
+    serializer.write("condition", toString(condition));
+    serializer.write("check", toString(check));
+    serializer.startArray("items");
+    for (auto const& item : items) {
+        serializer.startObject("item");
+
+        serializer.startArray("expressions");
+        for (auto const& ex : item.expressions) {
+            if (ex)
```
MikePopoloski: Same comment about null list items.

source/binding/Statements.cpp
```diff
@@ -995,6 +1033,32 @@ bool CaseStatement::verifyConstantImpl(EvalContext& context) const {
...
+                serializer.serialize(*ex);
+        }
+        serializer.endArray();
+
+        serializer.startObject("stmt");
```
MikePopoloski: This is now wrong in the opposite way; you're putting the plain statement as an element of the object, but objects have properties, not undecorated elements (basically that's the difference in JSON between properties and arrays). You probably just want `serializer.write("stmt", *item.stmt);`

source/binding/Statements.cpp
```diff
@@ -1083,6 +1147,27 @@ bool ForLoopStatement::verifyConstantImpl(EvalContext& context) const {
     return body.verifyConstant(context);
 }
 
+void ForLoopStatement::serializeTo(ASTSerializer& serializer) const {
+    serializer.startArray("initializers");
+    for (auto const& ini : initializers) {
+        if (ini)
```
MikePopoloski: Same comment about null list item.

source/binding/Statements.cpp
```diff
@@ -1083,6 +1147,27 @@ bool ForLoopStatement::verifyConstantImpl(EvalContext& context) const {
...
+            serializer.serialize(*ini);
+    }
+    serializer.endArray();
+
+    if (stopExpr)
+        serializer.write("stopExpr", *stopExpr);
+
+    serializer.startArray("steps");
+    for (auto const& step : steps) {
+        if (step)
```
MikePopoloski: Same comment about null list item.

source/binding/Statements.cpp
```diff
@@ -1239,6 +1329,25 @@ ER ForeachLoopStatement::evalRecursive(EvalContext& context, span<const Constant
     return ER::Success;
 }
 
+void ForeachLoopStatement::serializeTo(ASTSerializer& serializer) const {
+    serializer.write("arrayRef", arrayRef);
+
+    serializer.startArray("loopRanges");
+    for (auto const& r : loopRanges) {
+        serializer.write("range", r.toString());
```
MikePopoloski: Seems like the same problem here, adding properties to an array. It seems my ASTSerializer API is too confusing :(

source/binding/Statements.cpp
Kuree: What I had in mind is something like this:
```JSON
"items": [
    {
        "expressions": [],
        "stmt": stmt
    },
    {
        "expressions": [],
        "stmt": stmt
    }
]
```

I will double check with the output.

source/binding/Statements.cpp
Kuree: Sorry haven't touched raw pointers for a while. I will refactor the code.

source/binding/Statements.cpp
Kuree: How can I generate a list of strings? Can I just add a function like `serializer.serialize("str")` and use it inside the loop?

source/binding/Statements.cpp
MikePopoloski: Yeah, in that case you need a startObject() call that doesn't write the property name.

source/binding/Statements.cpp
MikePopoloski: Yeah, that's how I would do it.

MikePopoloski: Perhaps the existing write() methods should have been named writeProperty() to make it clear what they're doing. I might make that change in a different diff, I'll think about it some more.

source/binding/Statements.cpp
Kuree: Thanks. I've validated the JSON output from case statement and it seems to be correct.

MikePopoloski:
Great, thanks!

PR #298, opened 2020-03-20
Add missing serializations

Kuree:
- Add `expr` to `CaseStatement`
- Add `body` and `arguments` to `SubroutineSymbol`
- Adjust `arguments` to `CallExpression` so that arguments will get properly serialized in both cases.

Once thing I'm not sure is whether to check if the `arguments` in `SubroutineSymbol` are empty and serialize accordingly.


MikePopoloski:
I think it's fine to have an empty arguments array if the function takes no arguments.

PR #303, opened 2020-07-24
Add option to enable static linking

tgorochowik:
This is what could be used to avoid issues with the slang conda-packages (and sv-tests).

This would help us avoid maintaining the patch mentioned in https://github.com/SymbiFlow/sv-tests/issues/955

Please let me know if this is acceptable or if you want this renamed or moved.


MikePopoloski:
Thanks, works for me!

PR #306, opened 2020-08-09
Built and run on macOS successfully

thingkingland:
I changed 4 files (totally 5 lines) to make slang source code built and run on macOS with clang.

1. Makes 2 headers file self-contained to have
 ```cpp
 #include <unordered_map>
 #include <vector>
 #include <string>
```
    for STL containers they use.

2. In source/CMakeLists.txt do not link stdc++fs like MSVC
```cmake
if(NOT CMAKE_CXX_COMPILER_ID MATCHES "MSVC" AND NOT APPLE)
```
3. Resolves ambiguity of size_t between uint64_t and uint32_t by adding static_cast
 ```cpp
uintToStr(text, static_cast<uint64_t>(lineNum));
```

### DIFF
--- a/include/slang/text/SourceManager.h
+++ b/include/slang/text/SourceManager.h
@@ -15,6 +15,8 @@
 #include <set>
 #include <shared_mutex>
 #include <variant>
+#include <unordered_map>
+#include <vector>
 
 #include "slang/text/SourceLocation.h"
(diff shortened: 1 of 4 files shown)


MikePopoloski:
Nice, thank you!

PR #308, opened 2020-08-11
A string literal can be assigned to an unpacked array of bytes.

thingkingland:
I am investigating slang failures on [SystemVerilog Tester](https://symbiflow.github.io/sv-tests/). This is an attempt to support test case [5.9-string-word-assignment.sv](https://github.com/SymbiFlow/sv-tests/blob/master/tests/chapter-5/5.9-string-word-assignment.sv). SV-2017 standard §5.9 on page 77 says:

> A string literal can be assigned to an unpacked array of bytes. If the size differs, it is left justified.
byte c3 [0:12] = "hello world\n" ;

Major C++ code changes include

1. bool Type::isUnpackedArrayOfByte() detects if a type is unpacked arrays of bytes.
2. ConstantValue ConstantValue::convertToByteArray(bitwidth_t size, bool isSigned) converts constant string literals to byte arrays.
3. Utilizes above two functions in Expression::convertAssignment and ConversionExpression::convert.

A sample SystemVerilog test case:
```sv
module top;
  byte b[3 : 0] = "hello world\n";
  logic [7:0]  c[3 : 0] = "hi2";
  // test ConversionExpression::convert
  localparam byte d[3 : 0] = "hi2";
  parameter bit [7:0]  e[3 : 0] = "hello world\n";
endmodule
```

### DIFF
--- a/include/slang/numeric/ConstantValue.h
+++ b/include/slang/numeric/ConstantValue.h
@@ -144,7 +144,8 @@ public:
     ConstantValue convertToReal() const;
     ConstantValue convertToShortReal() const;
     ConstantValue convertToStr() const;
-    ConstantValue convertToByteArray(bitwidth_t size, bool isSigned = false) const;
+    ConstantValue convertToByteArray(bitwidth_t size, bool isSigned) const;
+    ConstantValue convertToByteQueue(bool isSigned) const;
 
     static const ConstantValue Invalid;
 
--- a/source/binding/AssignmentExpressions.cpp
+++ b/source/binding/AssignmentExpressions.cpp
@@ -105,7 +105,7 @@ namespace slang {
 Expression& Expression::implicitConversion(const BindContext& context, const Type& targetType,
                                            Expression& expr) {
     ASSERT(targetType.isAssignmentCompatible(*expr.type) ||
-           (targetType.canBeStringLike() && expr.isImplicitString()) ||
+           ((targetType.isString() || targetType.isByteArray()) && expr.isImplicitString()) ||
            (targetType.isEnum() && isSameEnum(expr, targetType)));
 
     Expression* result = &expr;
@@ -292,7 +292,7 @@ Expression& Expression::convertAssignment(const BindContext& context, const Type
     if (!type.isAssignmentCompatible(*rt)) {
         // String literals have a type of integer, but are allowed to implicitly convert to the
         // string type. See comments on isSameEnum for why that's here as well.
-        if ((type.canBeStringLike() && expr.isImplicitString()) ||
+        if (((type.isString() || type.isByteArray()) && expr.isImplicitString()) ||
             (type.isEnum() && isSameEnum(expr, type))) {
 
             result = &implicitConversion(context, type, *result);
--- a/source/binding/AssignmentExpressions.cpp
+++ b/source/binding/AssignmentExpressions.cpp
@@ -105,8 +105,7 @@ namespace slang {
 Expression& Expression::implicitConversion(const BindContext& context, const Type& targetType,
                                            Expression& expr) {
     ASSERT(targetType.isAssignmentCompatible(*expr.type) ||
-           (expr.isImplicitString() &&
-            (targetType.isString() || targetType.isUnpackedArrayOfByte())) ||
+           (targetType.canBeStringLike() && expr.isImplicitString()) ||
            (targetType.isEnum() && isSameEnum(expr, targetType)));
 
     Expression* result = &expr;
@@ -293,7 +292,7 @@ Expression& Expression::convertAssignment(const BindContext& context, const Type
     if (!type.isAssignmentCompatible(*rt)) {
         // String literals have a type of integer, but are allowed to implicitly convert to the
         // string type. See comments on isSameEnum for why that's here as well.
-        if ((expr.isImplicitString() && (type.isString() || type.isUnpackedArrayOfByte())) ||
+        if ((type.canBeStringLike() && expr.isImplicitString()) ||
             (type.isEnum() && isSameEnum(expr, type))) {
 
             result = &implicitConversion(context, type, *result);
@@ -605,9 +604,20 @@ ConstantValue ConversionExpression::convert(EvalContext& context, const Type& fr
         return std::move(value);
     }
 
-    if (to.isUnpackedArrayOfByte()) {
-        const FixedSizeUnpackedArrayType* ct = &to.as<FixedSizeUnpackedArrayType>();
-        return value.convertToByteArray(ct->range.width(), to.getArrayElementType()->isSigned());
+    if (to.isByteArray()) {
+        const auto& ct = to.getCanonicalType();
+        const auto isSigned = ct.getArrayElementType()->isSigned();
+        if (ct.isQueue()) {
+            return value.convertToByteQueue(isSigned);
+        }
+        else {
+            bitwidth_t size;
+            if (ct.hasFixedRange())
+                size = ct.as<FixedSizeUnpackedArrayType>().range.width();
+            else
+                size = 0; // dynamic array use string size
+            return value.convertToByteArray(size, isSigned);
+        }
     }
 
     // TODO: other types
--- a/source/numeric/ConstantValue.cpp
+++ b/source/numeric/ConstantValue.cpp
@@ -343,7 +343,7 @@ ConstantValue ConstantValue::convertToStr() const {
     int32_t extraBits = int32_t(val.getBitWidth() % 8);
 
     std::string result;
-    result.reserve(msb / 8 + 1);
+    result.reserve((val.getBitWidth() + 7) / 8);
     if (extraBits) {
         auto c = val.slice(msb, msb - extraBits + 1).as<uint8_t>();
         if (c && *c)
@@ -369,7 +369,7 @@ ConstantValue ConstantValue::convertToByteArray(bitwidth_t size, bool isSigned) 
         const SVInt& val = integer();
         int32_t msb = int32_t(val.getBitWidth() - 1);
         int32_t extraBits = int32_t(val.getBitWidth() % 8);
-        result.reserve(msb / 8 + 1);
+        result.reserve((val.getBitWidth() + 7) / 8);
         if (extraBits) {
             auto c = val.slice(msb, msb - extraBits + 1).as<uint8_t>();
             if (c && *c)
@@ -392,7 +392,7 @@ ConstantValue ConstantValue::convertToByteArray(bitwidth_t size, bool isSigned) 
     for (auto ch: result) {
         if (array.size() >= size)
             break;
-        array.emplace_back(SVInt(8, ch, isSigned));
+        array.emplace_back(SVInt(8, static_cast<uint64_t>(ch), isSigned));
     }
     while (array.size() < size) {
         array.emplace_back(SVInt(8, 0, isSigned));
--- a/source/numeric/ConstantValue.cpp
+++ b/source/numeric/ConstantValue.cpp
@@ -365,39 +365,40 @@ ConstantValue ConstantValue::convertToByteArray(bitwidth_t size, bool isSigned) 
     if (isUnpacked())
         return *this;
     std::string result;
-    if (isInteger()) {
-        const SVInt& val = integer();
-        int32_t msb = int32_t(val.getBitWidth() - 1);
-        int32_t extraBits = int32_t(val.getBitWidth() % 8);
-        result.reserve((val.getBitWidth() + 7) / 8);
-        if (extraBits) {
-            auto c = val.slice(msb, msb - extraBits + 1).as<uint8_t>();
-            if (c && *c)
-                result.push_back(char(*c));
-            msb -= extraBits;
-        }
-        while (msb >= 7) {
-            auto c = val.slice(msb, msb - 7).as<uint8_t>();
-            if (c && *c)
-                result.push_back(char(*c));
-            msb -= 8;
-        }
-    } else if (isString()) {
+    if (isInteger())
+        result = convertToStr().str();
+    else if (isString())
         result = str();
-    } else {
+    else
         return nullptr;
-    }
     Elements array;
+    if (!size) // dynamic array use string size
+        size = static_cast<bitwidth_t>(result.size());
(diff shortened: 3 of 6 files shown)


source/symbols/Type.cpp
```diff
@@ -239,6 +239,13 @@ bool Type::isUnpackedArray() const {
     }
 }
 
+bool Type::isUnpackedArrayOfByte() const {
```
MikePopoloski, line 242: The Type::isByteArray method already does much the same thing, though it only allows actual byte arrays but does allow dynamic arrays. The LRM isn't very clear on this point, but I tend to think it should only allow the byte type, but should allow dynamic arrays and queue as well. Thoughts?

include/slang/numeric/ConstantValue.h
```diff
@@ -144,6 +144,7 @@ class ConstantValue {
     ConstantValue convertToReal() const;
     ConstantValue convertToShortReal() const;
     ConstantValue convertToStr() const;
+    ConstantValue convertToByteArray(bitwidth_t size, bool isSigned = false) const;
```
MikePopoloski, line 147: I'd prefer not to default the isSigned... it really should always be provided (or it should always be true, if we only allow this conversion to actual "byte" arrays, since byte is signed in SystemVerilog.

source/binding/AssignmentExpressions.cpp
```diff
@@ -105,7 +105,8 @@ namespace slang {
 Expression& Expression::implicitConversion(const BindContext& context, const Type& targetType,
                                            Expression& expr) {
     ASSERT(targetType.isAssignmentCompatible(*expr.type) ||
-           (targetType.isString() && expr.isImplicitString()) ||
+           (expr.isImplicitString() &&
+            (targetType.isString() || targetType.isUnpackedArrayOfByte())) ||
```
MikePopoloski, line 109: Could simplify this a bit by using the Type::canBeStringLike() method here.

source/binding/AssignmentExpressions.cpp
```diff
@@ -604,6 +605,11 @@ ConstantValue ConversionExpression::convert(EvalContext& context, const Type& fr
         return std::move(value);
     }
 
+    if (to.isUnpackedArrayOfByte()) {
+        const FixedSizeUnpackedArrayType* ct = &to.as<FixedSizeUnpackedArrayType>();
```
MikePopoloski, line 609: When doing as<> casting on types you need to be careful to get the canonical type first, in case there are typedefs involved.

source/numeric/ConstantValue.cpp
```diff
@@ -360,6 +361,45 @@ ConstantValue ConstantValue::convertToStr() const {
     return result;
 }
 
+ConstantValue ConstantValue::convertToByteArray(bitwidth_t size, bool isSigned) const {
+    if (isUnpacked())
+        return *this;
+    std::string result;
+    if (isInteger()) {
```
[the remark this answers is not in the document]
thingkingland: Yes, inside the new convertToByteArray I will replace copied code with one line
```c++
        result = convertToStr().str();
```

source/symbols/Type.cpp
thingkingland: I plan to remove isUnpackedArrayOfByte and replace it with
```c++
type.isByteArray() && type.getCanonicalType().hasFixedRange()
```
To be safe, I will limit the enhancement to fixed size packed array for now. Only using isByteArray will accept SystemVerilog code like
```sv
localparam byte d[] = "hi2";
```
and it is a question how to determine at compile time the size of a dynamic array whose size can be changed at runtime. Please let me know if you think it is better to have a new function than to use existing isByteArray() && hasFixedRange().

include/slang/numeric/ConstantValue.h
thingkingland: I agree. I will remove the default value for isSigned.

source/binding/AssignmentExpressions.cpp
thingkingland: canBeStringLike() accepts dynamic arrays. As mentioned above, for safety I will limit the enhancement to fixed-size unpacked array for now, so canBeStringLike() does not look good for now.

source/binding/AssignmentExpressions.cpp
thingkingland: Yes, I will change the code into
```c++
    if (to.isByteArray()) {
        const auto& ct = to.getCanonicalType();
        if (ct.hasFixedRange()) {
            const FixedSizeUnpackedArrayType* cf = &ct.as<FixedSizeUnpackedArrayType>();
            return value.convertToByteArray(cf->range.width(),
                                            to.getArrayElementType()->isSigned());
        }
    }
```

thingkingland:
Thanks for the feedback. For every file I changed in a pull request, I will run
```sh
clang-format -style=file -i <file>
```
For files I do not change, I will not touch them with clang-format.

I will work on adding unit test cases.

source/numeric/ConstantValue.cpp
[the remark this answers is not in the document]
MikePopoloski: You have to check validity first; if evaluation fails it will return a ConstantValue that is "bad" and you need to just return a bad value in that case (return nullptr). But otherwise yeah, that works.

source/symbols/Type.cpp
MikePopoloski: Hmm, I think we should either make this new code support dynamic arrays as well, or we should remove the dynamic array support from isByteArray(). It's ok for a parameter to generally be a dynamic array, the code can already resolve it at compile time based on the assignment, so it should be possible to make it work for string literals as well (it should be sized according to the number of characters in the string). The question is whether SystemVerilog allows that according to the LRM, and I think it does because it uses the term "unpacked array" to mean all forms of unpacked arrays (fixed size, dynamic, and queues). 

I do think isByteArray() needs to be changed to disallow associative arrays, because those clearly don't work. You could make that change or I can do it.

I think you don't have to do much extra work to support dynamic arrays, it should really just work as-is. For queues, your convert function will need to check whether to.isQueue() and create an SVQueue instead of Elements vector.

MikePopoloski:
Thanks, clang-format and unit tests sound good. I think the main thing to resolve is supporting dynamic arrays and queues, which I think shouldn't be too hard and would let you reduce the number of special cases (no need to check for fixed size, for example).

source/numeric/ConstantValue.cpp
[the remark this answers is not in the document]
thingkingland: My code checks isInteger() before calling convertToStr()
```c++
    if (isInteger()) {
        result = convertToStr().str();
    }
```
and the current convertToStr() code cannot fail if isInteger() is true. Do you still think another check on convertToStr() return value is necessary?

source/symbols/Type.cpp
thingkingland: What you said makes sense. My solution is

1. In isByteArray(), reject associative arrays by adding
```c++
    if (ct.kind == SymbolKind::AssociativeArrayType)
        return false;
```
2. Convert string to byte queue by adding  ConstantValue::convertToByteQueue(bool isSigned) const;
```c++
ConstantValue ConstantValue::convertToByteQueue(bool isSigned) const {
    if (isQueue())
        return *this;
    std::string result;
    if (isInteger())
        result = convertToStr().str();
    else if (isString())
        result = str();
    else
        return nullptr;
    SVQueue queue;
    for (auto ch : result)
        queue.emplace_back(SVInt(8, static_cast<uint64_t>(ch), isSigned));
    return queue;
}
```
3. In ConversionExpression::convert, pass size=0 to convertToByteArray for dynamic arrays, and call convertToByteQueue for queues.
```c++
    if (to.isByteArray()) {
        const auto& ct = to.getCanonicalType();
        const auto isSigned = ct.getArrayElementType()->isSigned();
        if (ct.isQueue()) {
            return value.convertToByteQueue(isSigned);
        }
        else {
            bitwidth_t size;
            if (ct.hasFixedRange())
                size = ct.as<FixedSizeUnpackedArrayType>().range.width();
            else
                size = 0;  // dynamic array use string size
            return value.convertToByteArray(size, isSigned);
        }
    }
```
4. Inside convertToByteArray, if size==0 from dynamic arrays, set size to string size:
```c++
ConstantValue ConstantValue::convertToByteArray(bitwidth_t size, bool isSigned) const {
    if (isUnpacked())
        return *this;
    std::string result;
    if (isInteger())
        result = convertToStr().str();
    else if (isString())
        result = str();
    else
        return nullptr;
    Elements array;
    if (!size)  // dynamic array use string size
        size = static_cast<bitwidth_t>(result.size());
    array.reserve(size);
    for (auto ch : result) {
        if (array.size() >= size)
            break;
        array.emplace_back(SVInt(8, static_cast<uint64_t>(ch), isSigned));
    }
    while (array.size() < size)
        array.emplace_back(SVInt(8, 0, isSigned));
    return array;
}
```
5. With these, I can use type.canBeStringLike() to replace type.isString() to accept such assignments.

A SV test case I created to test these C++ code is
```sv
module test;
  localparam byte d[] = "dynamic array";
  localparam byte q[$] = "queue";
  localparam dynamicSize = d.size;
  localparam queueSize = q.size;
  // Asscoiative array should fail
  // localparam byte a[string] = "associative";
endmodule
```

source/numeric/ConstantValue.cpp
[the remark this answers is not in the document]
MikePopoloski: Ah yeah, if you're already checking isInteger() then it's fine.

source/symbols/Type.cpp
MikePopoloski: Yeah, this all sounds good to me. Thanks!

thingkingland:
I append a unit test to ExpressionTests.cpp:
```c++
std::string testStringLiteralsToByteArray(const std::string& text) {
    const auto& fullText = "module Top; " + text + " endmodule";
    auto tree = SyntaxTree::fromText(string_view(fullText));

    Compilation compilation;
    compilation.addSyntaxTree(tree);
    NO_COMPILATION_ERRORS;

    const auto& module = *compilation.getRoot().topInstances[0];
    if (!tree->diagnostics().empty())
        WARN(report(tree->diagnostics()));

    const ParameterSymbol& param = module.body.memberAt<ParameterSymbol>(0);
    const auto& value = param.getValue();
    std::string result;
    if (value.isUnpacked()) {
        for (const auto& svInt : value.elements()) {
            REQUIRE(svInt.isInteger());
            REQUIRE(svInt.integer().getBitWidth() == 8);
            result.push_back(static_cast<char>(*svInt.integer().getRawPtr()));
        }
    }
    else {
        REQUIRE(value.isQueue());
        for (const auto& svInt : value.queue()) {
            REQUIRE(svInt.isInteger());
            REQUIRE(svInt.integer().getBitWidth() == 8);
            result.push_back(static_cast<char>(*svInt.integer().getRawPtr()));
        }
    }
    return result;
}

TEST_CASE("String literal assigned to unpacked array of bytes") {
    // Unpacked array size shorter than string literals
    CHECK(testStringLiteralsToByteArray("localparam byte a[3:0] = \"hello world\";") == "hell");
    // Unpacked array size longer than string literals
    CHECK(testStringLiteralsToByteArray("localparam byte a[1:4] = \"hi\";") == "hi");
    // Dynamaic array
    CHECK(testStringLiteralsToByteArray("localparam byte d[] = \"hello world\";") == "hello world");
    // Queue
    CHECK(testStringLiteralsToByteArray("localparam byte q[$] = \"hello world\";") ==
          "hello world");
}
```

tests/unittests/ExpressionTests.cpp
```diff
@@ -1239,4 +1239,69 @@ endmodule
...
+std::string testStringLiteralsToByteArray(const std::string& text) {
+    const auto& fullText = "module Top; " + text + " endmodule";
+    auto tree = SyntaxTree::fromText(string_view(fullText));
+
+    Compilation compilation;
+    compilation.addSyntaxTree(tree);
+    NO_COMPILATION_ERRORS;
+
+    const auto& module = *compilation.getRoot().topInstances[0];
+    if (!tree->diagnostics().empty())
```
MikePopoloski, line 1253: The NO_COMPILATION_ERRORS macro will already check that there are no diagnostics, so this bit of code here is unnecessary.

tests/unittests/ExpressionTests.cpp
```diff
@@ -1239,4 +1239,69 @@ endmodule
...
+        WARN(report(tree->diagnostics()));
+
+    const ParameterSymbol& param = module.body.memberAt<ParameterSymbol>(0);
+    const auto& value = param.getValue();
+    std::string result;
+    if (value.isUnpacked()) {
+        for (const auto& svInt : value.elements()) {
+            REQUIRE(svInt.isInteger());
```
MikePopoloski, line 1261: This is fine, but .integer() will already assert so you don't actually have to REQUIRE() isInteger here. Doesn't hurt though.

tests/unittests/ExpressionTests.cpp
```diff
@@ -1239,4 +1239,69 @@ endmodule
...
+            REQUIRE(svInt.integer().getBitWidth() == 8);
+            auto ch = static_cast<char>(*svInt.integer().getRawPtr());
```
MikePopoloski, line 1263: getRawPtr() works in this case, but the more straightforward way to get a concrete integer out of an SVInt is to use as<>, like:
`*svInt.integer().as<char>()`

It returns an optional<> but in this case you already checked the bitwidth so you know it can't fail to convert.

tests/unittests/ExpressionTests.cpp
thingkingland: I removed the redundant code as suggested.

tests/unittests/ExpressionTests.cpp
thingkingland: I remove the redundant code as suggested.

tests/unittests/ExpressionTests.cpp
thingkingland: I replace getRawPtr() with *svInt.integer().as<char>() as suggested.

MikePopoloski:
Looks great, merged!

PR #309, opened 2020-08-13
Fix endless loop and growing memory for an unsupported statement

thingkingland:
On [SystemVerilog Tester](https://symbiflow.github.io/sv-tests/), slang has 7.5GB maximum ram usage, worst among the 14 Verilog/SystemVerilog parsers under test. The maximum ram usage occurred for a [test case](https://github.com/Nic30/hdlConvertor/blob/f7344ddfc8826b71ab63a8b77dbad4197de554c4/tests/sv_test/std2017/p550.sv) with only 15 lines but with an unsupported statement **randsequence** (SV2017 §18.7).

From Parser::parseBlockItems() inside a while loop at line 594, the parser calls Parser::parseStatement() and reaches the end
```c++
    addDiag(diag::ExpectedStatement, peek().location());
    return factory.emptyStatement(label, attributes,
                                  missingToken(TokenKind::Semicolon, peek().location()));
}
```
Because the token is not advanced by "consume" or "skipToken", the while loop in Parser::parseBlockItems() sticks at the same token, and keeps creating empty statements to grow memory usage.
```c++
       else if (isPossibleStatement(kind)) {
            newNode = &parseStatement();
            isStmt = true;
            sawStatement = true;
        }
```
The solution detects after the return of Parser::parseStatement() whether the return statement is
1. An empty statement.
2. With a missing token.
3. No change in token location.

and call "skipToken" to advance token location. With that, the unsupported test case finishes quickly with little memory, after generating several error messages.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,8 @@
+root = true
+
+[*]
+indent_style = space
+indent_size = 4
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
--- a/include/slang/binding/BindContext.h
+++ b/include/slang/binding/BindContext.h
@@ -128,8 +128,8 @@ public:
 
     optional<int32_t> evalInteger(const ExpressionSyntax& syntax) const;
     optional<int32_t> evalInteger(const Expression& expr) const;
-    EvaluatedDimension evalDimension(const VariableDimensionSyntax& syntax,
-                                     bool requireRange) const;
+    EvaluatedDimension evalDimension(const VariableDimensionSyntax& syntax, bool requireRange,
+                                     bool isPacked) const;
 
     optional<ConstantRange> evalPackedDimension(const VariableDimensionSyntax& syntax) const;
     optional<ConstantRange> evalPackedDimension(const ElementSelectSyntax& syntax) const;
@@ -138,7 +138,8 @@ public:
     BindContext resetFlags(bitmask<BindFlags> addedFlags) const;
 
 private:
-    void evalRangeDimension(const SelectorSyntax& syntax, EvaluatedDimension& result) const;
+    void evalRangeDimension(const SelectorSyntax& syntax, bool isPacked,
+                            EvaluatedDimension& result) const;
 };
 
 } // namespace slang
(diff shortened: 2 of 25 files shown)


MikePopoloski:
Thanks for fixing. Can you add a test case to StatementParsingTests that reproduces the issue, just so we can prevent regressions in the future?

Also, I took a look at other places we call parseStatement() and I think this same bug might exist in Parser::parseRandCaseStatement. If you have time could you fix that location as well?

thingkingland:
I will do these. A question is whether there is a recommended "timeout" check in the unit test methodology. The failure scenarios is hanging/stuck for this test case, and others who run unit tests may think it is taking a long time to finish unit testing without seeing a failure.  For example, Apple has [wait(for:timeout:)](https://developer.apple.com/documentation/xctest/xctestcase/2806856-wait) for its unit tests. If not, I can try to create one.

MikePopoloski:
I like the idea of a timeout but I'm not sure how you could reasonably implement one with how the unit tests currently work -- they're just a C++ process running (and not using threads). If tests are invoked via ctest (which they are for CI builds) then I think there should be a default timeout in place already.

thingkingland:
1. I append a unit test to StatementParsingTests.cpp. ctest timeout works for such scenarios. I tried
```sh
ctest --timeout 2
```
and it detects the problem.

2. I add similar checks in Parser::parseRandCaseStatement. I do not add a unit test for this because slang rejects randcase.
```txt
randCase.sv:4:5: error: language feature not yet supported
```

thingkingland:
ctest default timeout seems to be 1500 seconds. To save the effort to use --timeout option, do you think append tests/unittests/CMakeLists.txt file with
```cmake
set_tests_properties(unittests PROPERTIES TIMEOUT 5) 
```
is a good idea? If it is, what is a good number for the timeout value?  On my Mac, unittests finishes in only 0.14 seconds.

MikePopoloski:
Yeah, I think it's a good idea to set the timeout. 5 seconds is probably fine; like you said, the unittests complete pretty quickly. You can make that a separate diff though, it's mostly unrelated to this one.

PR #310, opened 2020-08-17
macOS support for codegen, runtime and simtests.

thingkingland:
This is to port slang to macOS when using "cmake -DSLANG_INCLUDE_LLVM=ON ...".
1. To suppress build error
> CodeGenerator.h:87:24: error: private field 'compilation' is not used [-Werror,-Wunused-private-field]
    const Compilation& compilation;

insert in source/CMakeLists.txt
```cmake
if (APPLE)
     target_compile_options(slangcodegen PRIVATE -Wno-unused-private-field)
endif()
```
2. To suppress build error
> CodeGenFunction.cpp:65:13: error: 4 enumeration values not handled in switch: 'bitnot', 'reducand', 'reducor'... [-Werror,-Wswitch]
    switch (instr.kind) {

insert in source/codegen/CodeGenFunction.cpp
```c++
default: // Suppress clang -Wswitch
    break;
```
3. To resolve build error ld: undefined symbol "llvm::Module::dump() const", follow [stackoverflow suggestion](https://stackoverflow.com/questions/46367910/llvm-5-0-linking-error-with-llvmmoduledump)
```c++
#ifdef __APPLE__
        module->print(llvm::errs(), nullptr); // ld: undefined symbol llvm::Module::dump()
#else
        module->dump();
#endif
```
4. With the above 3 changes, build is successful but simtests crashes. The first problem at the end of JIT::run
```c++
    auto fp = (int (*)())(intptr_t)sym->getAddress();
    return fp();
```
there are messages
> JIT session error: Symbols not found: [ _printInt, _flush, _printStr ]
JIT session error: Failed to materialize symbols: { (0x1037d1fa0, { ___orc_anon.0, _main }) }

Follow [llvm doc](https://llvm.org/docs/ORCv2.html)
> On Linux this mangling is a no-op, but on other platforms it usually involves adding a prefix to the string (e.g. ‘_’ on Darwin). 

to insert in JIT.cpp
```c++
#ifdef __APPLE__
    // Mangle names according to https://llvm.org/docs/ORCv2.html
    auto& dl = jit->getDataLayout();
    MangleAndInterner Mangle(jit->getExecutionSession(), dl);
#endif
#ifdef __APPLE__
        auto err = jit->defineAbsolute(*Mangle(llvm::StringRef(name.data(), name.length())), sym);
#else
```
5. With the above change, simtests passes on debug build (-DCMAKE_BUILD_TYPE=Debug) but with -O1 or above optimization still crash at JITTests.cpp for bad address of "result" inside lambda call
```c++
    slang::runtime::setOutputHandler([&](string_view text) { result += text; });
```
I suspect that it is a clang optimization bug and work around it by
```c++
#ifdef __APPLE__
// Work around a clang optimization bug on deallocating "result" before lambda call
[[clang::optnone]]
#endif
```

With the above 5 changes, -DSLANG_INCLUDE_LLVM=ON build is successful and simtests 100% passes on macOS.
In addition. TIMEOUT properties are appended for unittests and simtests.

### DIFF
--- a/source/CMakeLists.txt
+++ b/source/CMakeLists.txt
@@ -200,7 +200,5 @@ if(SLANG_INCLUDE_LLVM)
 
     llvm_map_components_to_libnames(llvm_libs support core orcjit native nativecodegen)
     target_link_libraries(slangcodegen PRIVATE ${llvm_libs})
-    if (APPLE)
-        target_compile_options(slangcodegen PRIVATE -Wno-unused-private-field)
-    endif()
+    target_compile_options(slangcodegen PRIVATE -Wno-unused-private-field)
 endif()
--- a/source/CMakeLists.txt
+++ b/source/CMakeLists.txt
@@ -200,4 +200,7 @@ if(SLANG_INCLUDE_LLVM)
 
     llvm_map_components_to_libnames(llvm_libs support core orcjit native nativecodegen)
     target_link_libraries(slangcodegen PRIVATE ${llvm_libs})
+    if (APPLE)
+        target_compile_options(slangcodegen PRIVATE -Wno-unused-private-field)
+    endif()
 endif()
--- a/source/codegen/CodeGenFunction.cpp
+++ b/source/codegen/CodeGenFunction.cpp
@@ -71,7 +71,7 @@ llvm::Value* CodeGenFunction::emit(const Instr& instr) {
             return emitStore(instr.getOperands()[0], instr.getOperands()[1]);
         case InstrKind::negate:
             return emitNegate(instr.type, instr.getOperands()[0]);
-        default: // Suppress clang -Wswitch
+        default:
             break;
     }
     THROW_UNREACHABLE;
--- a/source/codegen/CodeGenFunction.cpp
+++ b/source/codegen/CodeGenFunction.cpp
@@ -71,6 +71,8 @@ llvm::Value* CodeGenFunction::emit(const Instr& instr) {
             return emitStore(instr.getOperands()[0], instr.getOperands()[1]);
         case InstrKind::negate:
             return emitNegate(instr.type, instr.getOperands()[0]);
+        default: // Suppress clang -Wswitch
+            break;
     }
     THROW_UNREACHABLE;
 }
--- a/source/codegen/CodeGenerator.cpp
+++ b/source/codegen/CodeGenerator.cpp
@@ -86,8 +86,13 @@ GeneratedCode CodeGenerator::finish() {
 
     // Verify all generated code.
     bool bad = llvm::verifyModule(*module, &llvm::errs());
-    if (bad)
+    if (bad) {
+#ifdef __APPLE__
+        module->print(llvm::errs(), nullptr); // ld: undefined symbol llvm::Module::dump()
+#else
         module->dump();
+#endif
+    }
 
     return GeneratedCode(std::move(ctx), std::move(module));
 }
--- a/source/codegen/CodeGenerator.cpp
+++ b/source/codegen/CodeGenerator.cpp
@@ -86,13 +86,8 @@ GeneratedCode CodeGenerator::finish() {
 
     // Verify all generated code.
     bool bad = llvm::verifyModule(*module, &llvm::errs());
-    if (bad) {
-#ifdef __APPLE__
+    if (bad)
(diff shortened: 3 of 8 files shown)


source/codegen/JIT.cpp
```diff
@@ -37,9 +37,18 @@ JIT::JIT() {
 
     // Register all exported simrt functions with the JIT.
     auto exported = slang::runtime::getExportedFunctions();
+#ifdef __APPLE__
+    // Mangle names according to https://llvm.org/docs/ORCv2.html
+    auto& dl = jit->getDataLayout();
+    MangleAndInterner Mangle(jit->getExecutionSession(), dl);
```
MikePopoloski, line 43: You probably don't need to hide this behind an ifdef for APPLE, it might just work on all platforms.

source/codegen/CodeGenerator.cpp
```diff
@@ -86,8 +86,13 @@ GeneratedCode CodeGenerator::finish() {
 
     // Verify all generated code.
     bool bad = llvm::verifyModule(*module, &llvm::errs());
-    if (bad)
+    if (bad) {
+#ifdef __APPLE__
```
MikePopoloski, line 90: This can probably just be unified and have all platforms call print()

source/codegen/CodeGenFunction.cpp
```diff
@@ -71,6 +71,8 @@ llvm::Value* CodeGenFunction::emit(const Instr& instr) {
             return emitStore(instr.getOperands()[0], instr.getOperands()[1]);
         case InstrKind::negate:
             return emitNegate(instr.type, instr.getOperands()[0]);
+        default: // Suppress clang -Wswitch
```
MikePopoloski, line 74: No need for the comment here, I should have just added this default case. This code isn't being built by CI builds right now.

source/CMakeLists.txt
```diff
@@ -200,4 +200,7 @@ if(SLANG_INCLUDE_LLVM)
 
     llvm_map_components_to_libnames(llvm_libs support core orcjit native nativecodegen)
     target_link_libraries(slangcodegen PRIVATE ${llvm_libs})
+    if (APPLE)
```
MikePopoloski, line 203: This is probably not Apple specific.

MikePopoloski:
I suspect some of these changes might not be Apple specific; the LLVM stuff isn't being built in CI builds right now. Perhaps that should be handled first, and only then focus on Apple-specific changes?

source/codegen/JIT.cpp
thingkingland: Yes, I remove #ifdef __APPLE__ in the new commit.

source/codegen/CodeGenerator.cpp
thingkingland: Yes, I remove #ifdef __APPLE__ and apply the replacement for all the platforms.

source/codegen/CodeGenFunction.cpp
thingkingland: Yes, I remove the comment.

source/CMakeLists.txt
thingkingland: Yes, I remove if(APPLE).

thingkingland:
I have not explained the changes in CMakeLists.txt of a later commit.
```cmake
set(CMAKE_SHARED_LIBRARY_CREATE_CXX_FLAGS "${CMAKE_SHARED_LIBRARY_CREATE_CXX_FLAGS} -undefined dynamic_lookup")
```
is for "cmake --DBUILD_SHARED_LIBS=ON ..." building shared libraries to succeed on macOS, based on this [stackoverflow suggestion](https://stackoverflow.com/questions/25421479/clang-and-undefined-symbols-when-building-a-library/30934307#30934307). Only macOS need this and Linux clang has this option as default.
```cmake
      if(STATIC_BUILD)
          message(SEND_ERROR "STATIC_BUILD=ON not supported on macOS." )
      endif()
```
is to have a clearer error message for "cmake -DSTATIC_BUILD=ON ..." instead of a mysterious "-lcrt.o" not found message. Apple does not support [statically linked binary.](https://developer.apple.com/library/archive/qa/qa1118/_index.html)

MikePopoloski:
Thanks!

PR #312, opened 2020-08-19
error out on unpacked array assignments with unequal fixed sizes

thingkingland:
SV-2007 standard §7.6 (Array Assignments) page 152 says the following SystemVerilog code should have type check error
```sv
  int A[10:1];
  int C[24:1];
  assign A = C;
```
but slang currently does not report any error. For a similar illegal parameter assignment
```sv
  localparam int a [1:0] = {1, 2};
  localparam int b [0:2] = a;
```
slang gives a good error in release build, but has an assertion failure in debug build at AssignmentExpressions.cpp:582
```c++
  ASSERT(!to.hasFixedRange() || !from.hasFixedRange());
```
A fix is included to report proper errors without any assertion failure. A unit test case is added to ensure proper error checking.

### DIFF
--- a/source/symbols/Type.cpp
+++ b/source/symbols/Type.cpp
@@ -401,6 +401,8 @@ bool Type::isAssignmentCompatible(const Type& rhs) const {
         // need to check the fixed size condition here, since the only way it would
         // matter is if the source (rhs) is dynamically sized, which can't be checked
         // until runtime.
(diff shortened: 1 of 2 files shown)


MikePopoloski:
Nice catch, I definitely should have caught this myself when I wrote it.

PR #313, opened 2020-08-19
Bit-stream type check on unpacked array/struct

thingkingland:
Now slang correctly detect an error where"event" is not a bit-stream type
```sv
  event a;
  localparam bits_a = $bits(a);
```
but not on unpacked array of event or unpacked struct of event
```sv
  event b [3:0];
  localparam bits_b = $bits(b);
  struct { event a; } c;
  localparam bits_c = $bits(c);
```
SV-2017 standard §6.24.3 on page 135 says for a bit-stream type "This definition is recursive".
This code enhances "Type::isBitstreamType" to recursively check elements of unpacked arrays and fields of unpacked struct. A unit test case is added to ensure proper error checking.

### DIFF
--- a/source/symbols/Type.cpp
+++ b/source/symbols/Type.cpp
@@ -201,19 +201,19 @@ bool Type::isStruct() const {
 }
 
 bool Type::isBitstreamType() const {
-    // TODO: classes
     if (isIntegral())
         return true;
     if (isUnpackedArray())
         return getArrayElementType()->isBitstreamType();
     if (isUnpackedStruct()) {
-        auto& us = as<UnpackedStructType>();
+        auto& us = getCanonicalType().as<UnpackedStructType>();
         for (auto& field : us.membersOfType<FieldSymbol>()) {
             if (!field.getType().isBitstreamType())
                 return false;
         }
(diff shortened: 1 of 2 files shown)


source/symbols/Type.cpp
```diff
@@ -202,7 +202,19 @@ bool Type::isStruct() const {
 
 bool Type::isBitstreamType() const {
     // TODO: classes
-    return isIntegral() || isUnpackedArray() || isUnpackedStruct();
+    if (isIntegral())
+        return true;
+    if (isUnpackedArray())
+        return getArrayElementType()->isBitstreamType();
+    if (isUnpackedStruct()) {
+        auto& us = as<UnpackedStructType>();
```
MikePopoloski, line 210: You have to use getCanonicalType() here in order to safely use as<>. Most other Type methods handle checking for canonical types for you automatically, but as<> casting is more low level.

thingkingland: Yes, I change the problematic line into
```c++
  auto& us = getCanonicalType().as<UnpackedStructType>();
```

PR #315, opened 2020-08-26
§6.24.3 Bit-stream casting support

thingkingland:
This is to support bit-stream casting in SV-2017 standard §6.24.3.
1. An optional parameter "destination" is added to Type::isBitstreamType(). For destination bit-stream types, associative arrays and classes are excluded.
```c++
    bool isBitstreamType(bool destination = false) const;
```
2. In the end of Type::isCastCompatible(), check legal bit-stream casting by
```c++
  if (l->isBitstreamType(true) && r->isBitstreamType()) {
        // bit-stream casting
        if (l->isFixedSize() && r->isFixedSize())
            return l->bitstreamWidth() == r->bitstreamWidth();
        else {
            // compile-time check of dynamic sizes
```
3. For constant evaluation of bit-stream casting, a function is added.
```c++
   ConstantValue Type::bitstreamCast(const ConstantValue&) const;
```
An alternative "ConstantValue ConstantValue::bitstreamCast(const Type&);" is not adopted to avoid creating new dependency from slangcore(ConstantValue) to slangcompiler(Type).

4. A few helper functions are added.
```c++
  std::vector<const ConstantValue*> ConstantValue::bitstream() const;
```
performs pack operation and creates a flatten list of SVInt/string from an aggregate ConstantValue. There is no flattening to a big SVInt to avoid exceeding SVINT::MAX_BITS. A new file "TypeHelpers.h" is added to include some helper functions. For example,
```c++
  static ConstantValue unpack(const Type& type, ....
```
performs unpack operation on the flattened list of SVInt/string from ConstantValue::bitstream(). It recursively traverses the type and use SVInt::slice and SVInt::concat to retrieve required bits for each packed type from the SVInt/string list.

5. A few unit tests are added.


source/symbols/TypeHelpers.h
```diff
@@ -0,0 +1,208 @@
...
+        cc.setSigned(type.isSigned());
+        return cc;
+    }
+    if (type.isString()) {
+        if (!dynamic)
+            return std::string();
+        auto width = dynamic;
+        ASSERT(width % CHAR_BIT == 0);
+        dynamic = 0;
+        SmallVectorSized<SVInt, 8> buffer;
```
MikePopoloski, line 162: This code is fully duplicated several times; please factor out into a helper method (or local lambda).

source/symbols/TypeHelpers.h
```diff
@@ -0,0 +1,208 @@
...
+                dynamic = 0;
+            }
+        }
+        else {
+            const auto& ct1 = ct.as<FixedSizeUnpackedArrayType>();
+            const auto& elem = ct1.elementType;
+            for (auto width = ct1.range.width(); width > 0; width--)
+                buffer.emplace(unpack(elem, iter, bit, dynamic));
+        }
+        return constContainer(ct, span<const ConstantValue>(buffer));
```
MikePopoloski, line 194: Are the explicit span constructors necessary everywhere? I would expect them to work implicitly by just passing the buffer.

source/symbols/TypeHelpers.h
```diff
@@ -0,0 +1,208 @@
...
+        auto msb = ci.getBitWidth() - bit - 1;
+        auto lsb = std::min(bit + width, ci.getBitWidth());
+        lsb = ci.getBitWidth() - lsb;
+        bit += msb - lsb + 1;
+        ASSERT(bit <= ci.getBitWidth());
+        return ci.slice(static_cast<int32_t>(msb), static_cast<int32_t>(lsb));
+    }
+    else {
+        std::string_view str = cp->str();
+        auto byte0 = bit / CHAR_BIT;
```
MikePopoloski, line 112: Instead of byte0 / byte1, perhaps firstByte / lastByte?

source/symbols/TypeHelpers.h
```diff
@@ -0,0 +1,208 @@
...
+static ConstantValue constContainer(const Type& type, span<const ConstantValue> elems) {
+    switch (type.kind) {
+        case SymbolKind::FixedSizeUnpackedArrayType:
+        case SymbolKind::DynamicArrayType:
+        case SymbolKind::UnpackedStructType:
+            return ConstantValue::Elements(elems.cbegin(), elems.cend());
+        case SymbolKind::QueueType:
+            return SVQueue(elems.cbegin(), elems.cend());
+        default:
+            return nullptr;
```
MikePopoloski, line 81: Is the default case here unreachable? If so, please use the THROW_UNREACHABLE macro

source/binding/AssignmentExpressions.cpp
```diff
@@ -548,6 +548,18 @@ Expression& ConversionExpression::fromSyntax(Compilation& compilation,
 }
 
 ConstantValue ConversionExpression::evalImpl(EvalContext& context) const {
+    if (!isImplicit && !type->isCastCompatible(*operand().type, false)) {
```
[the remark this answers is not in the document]
thingkingland: The reason I didn't put it inside static function"convert" is that I need ConversionExpression::isImplicit member variable, unless an extra argument of "convert" is added to pass "isImplicit". The existing "convert" code will return nullptr or assertion fail, if bitstream cast is not processed and returned before, so bitstream cast must be checked first.

An extra isCastCompatible argument and calling isCastCompatible again does not seem elegant. How about adding an extra "bool ConversionExpression::isBitstreamCast" member variable or combine "isImplicit" into an enum, to record what a "ConversionExpression" is?  Doing so may require slight modification of all code calling isCastCompatible.

include/slang/numeric/ConstantValue.h
```diff
@@ -150,6 +150,9 @@ class ConstantValue {
     /// Gets the size of this value when converted to a bitstream.
     bitwidth_t bitstreamWidth() const;
 
+    /// Performs pack operation to create a bit stream
+    std::vector<const ConstantValue*> bitstream() const;
```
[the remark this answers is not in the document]
thingkingland: Yes. I will move it into a static helper function inside TypeHelper.h.

source/numeric/ConstantValue.cpp
```diff
@@ -426,6 +426,34 @@ bitwidth_t ConstantValue::bitstreamWidth() const {
     return width;
 }
 
+using PackVector = decltype(std::declval<ConstantValue>().bitstream());
+static void pack(const ConstantValue& value, PackVector& packed) {
+    if (value.isInteger() || value.isString()) {
+        if (value.isInteger() || !value.str().empty())
```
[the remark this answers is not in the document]
thingkingland: Yes, I will do that.

source/numeric/ConstantValue.cpp
```diff
@@ -426,6 +426,34 @@ bitwidth_t ConstantValue::bitstreamWidth() const {
...
+    else if (value.isMap()) {
+        for (const auto& kv : *value.map()) {
+            pack(kv.second, packed);
+        }
+    }
+    else if (value.isQueue()) {
+        for (const auto& cv : *value.queue())
+            pack(cv, packed);
+    }
+    // TODO: classes
```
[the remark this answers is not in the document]
thingkingland: Yes, I will add an "THROW_UNREACHABLE;" at the end.

source/symbols/TypeHelpers.h
```diff
@@ -0,0 +1,208 @@
+
```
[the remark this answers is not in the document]
thingkingland: Yes, I will remove the blank line.

source/symbols/TypeHelpers.h
```diff
@@ -0,0 +1,208 @@
...
+        ASSERT(bit <= str.length() * CHAR_BIT);
+        if (lsb > 0 || msb < len - 1)
+            return ci.slice(static_cast<int32_t>(msb), static_cast<int32_t>(lsb));
+        else
+            return ci;
+    }
+}
+
+/// Performs unpack operation on a bit stream
+static ConstantValue unpack(const Type& type, PackVector::const_iterator& iter, bitwidth_t& bit,
```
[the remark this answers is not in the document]
thingkingland: I will try to do so.

source/symbols/TypeHelpers.h
thingkingland: Yes, I will do that.

source/symbols/TypeHelpers.h
thingkingland: I tried removing them, but compile failed. In one case, I simplified it into span<buffer>, clang/gcc accepted, but msvc failed.

source/symbols/TypeHelpers.h
thingkingland: Yes, I will change it.

source/symbols/TypeHelpers.h
thingkingland: Yes, I will change it into THROW_UNREACHABLE.

source/symbols/TypeHelpers.h
```diff
@@ -0,0 +1,208 @@
...
+#include <numeric>
+
+#include "slang/symbols/AllTypes.h"
+#include "slang/symbols/VariableSymbols.h"
+
+namespace slang {
+
+/// Model dynamic size of a type with aR+b where a and b are compile-time constants and R is runtime
+/// determined.
+static std::pair<bitwidth_t, bitwidth_t> linearCoefficients(const Type& type,
```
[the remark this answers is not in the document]
thingkingland: I change the name into "dynamicBitstreamSize". The purpose of this function is to handle two requests from IEEE standard.
```sv
     // runtime check
     struct {bit a[$]; shortint b;} a = {{1,2,3,4}, 67};
     int b = int'(a);
     // compile time error
     typedef struct {byte a[$]; bit b;} dest_t;
     int a;
     dest_t b = dest_t'(a);
```
IEEE standard asks to check some dynamic cases at compile time and does not say how, and I derived this algorithm. IEEE standard also says for "dest_'(source_t)", the destination type rule is

> for unbounded dynamically sized types, the conversion process is greedy: adjust the size of the first dynamically sized item in the destination to the remaining size; any remaining dynamically sized items are left empty.

This is why source and destination have different rules. While the dynamic size items of source type can have any runtime sizes, the destination type can have two situations.
1. All dynamic-sized items are empty.
2. The first item is filled, and the remaining is empty. If the parent/ancestors of the first item is dynamic sized, their size should be one, and the fixed-sized siblings should be counted once too.

I will add the above two quotes from IEEE standard as comments at certain places. I also improve the comments for "dynamicBitstreamSize" function to explain its argument and return value:
```c++
/// Return {a, b} to model the size of a dynamic size with "aR+b" where "a" and "b" are compile-time
/// constants and "R" is runtime determined.

// destination=0: source type where "a" is the greatest common divisor of element sizes of
// all dynamically sized items and "b" is the sum of all fixed sizes.

// destination=1: destination type with all dynamically sized items empty. "a" is zero and "b" is
// the sum of all fixed sizes.

// destination=2: destination type with first dynamically sized item filled and "a" is the element
// size of the first dynamically sized item. The remaining dynamically sized items are empty except
// ancestors of the first item need to have size one. "b" is the sum of all fixed sizes plus size
// of siblings of the first dynamically sized item if the common parent is dynamically sized.
```

source/symbols/TypeHelpers.h
```diff
@@ -0,0 +1,208 @@
...
+     */
+    auto width = type.getBitWidth();
+    if (width > 0)
+        return { 0, width };
+    if (type.isString())
+        return { destination == 1 ? 0 : CHAR_BIT, 0 };
+    bitwidth_t gcd = 0;
+    // TODO: check for overflow
+    if (type.isUnpackedArray()) {
+        auto [gcd1, width1] = linearCoefficients(*type.getArrayElementType(), destination);
```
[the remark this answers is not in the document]
thingkingland: I change the variable names as follows:

- gcd => multiplier
- width => fixedSize
- gcd1 => multiplierElem
- width1 => fixedSizeElem

source/symbols/TypeHelpers.h
```diff
@@ -0,0 +1,208 @@
...
+        const auto& ct = type.getCanonicalType();
+        if (ct.kind == SymbolKind::FixedSizeUnpackedArrayType) {
+            auto rw = ct.as<FixedSizeUnpackedArrayType>().range.width();
+            gcd = gcd1 * rw;
+            width = width1 * rw;
+        }
+        else if (destination) {
+            if (destination > 1) {
```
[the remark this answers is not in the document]
thingkingland: I am adding more comments like
```c++
        else if (destination) {    // dynamically sized destination type
            if (destination > 1) { // Fill first dynamically sized item otherwise keep empty
                if (!multiplierElem)
                    multiplier = fixedSizeElem; // element is fixed size
                else {
                    multiplier = multiplierElem; // element is dynamically sized
                    fixedSize = fixedSizeElem;
                }
            }
        }
        else
            multiplier = std::gcd(multiplierElem, fixedSizeElem); // dynamically sized source type
```

source/binding/AssignmentExpressions.cpp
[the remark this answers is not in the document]
MikePopoloski: Hmm... I'm not sure why isImplicit matters much here. If you check at the end of convert() for a bitstream cast, then there should be no problem, regardless of whether isImplicit is true or not. It might be some extra code to check for the bitstream cases in convert(), like for example if the target isIntegral() then you need an extra check to see if the source is one of the ones that previous worked (integral, float) or whether it's a bitstream case.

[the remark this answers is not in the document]
thingkingland: It is possible, but there are some complications. The difficult parts are
```c++
    if (to.isUnpackedArray() && from.isUnpackedArray()) {
    if (to.isByteArray()) {
```
For example, for
```sv
   localparam byte a[3:0] = "hello world";
```
Bit-stream casting requires sizes to be the same and need to reject this, but implicit conversion allows truncation and extension, the rules are not exactly the same. For the error message:
```c++
   context.addDiag(diag::ConstEvalDynamicToFixedSize, sourceRange)
```
though bitstreamCast can detect size mismatches too, it is nicer to have different error messages for bit-stream casting and implicit conversions to a user. Implicit conversion allows unpacked arrays of reals, but bit-stream casting forbids it. Of course, we can repeat more type checking on what kind of conversion is in "convert" function, but it defeats the purpose of reusing existing code.
I am thinking adding
```c++
enum class ConversionFlags {
    Implicit = 1,
    BitstreamCast = 2,
    Propagated = 4,
    Invalid = 8
}
BITMASK(ConversionFlags, Invalid)
```
to have more flags for ConversionExpression. In another problem I investigated, the rule in ConstantValue::convertToInt
```c++
    // [11.8.3] says that during an assignment we sign extend iff the rhs is signed.
    // That means we should resize first, and only then change the sign flag if desired.
```
is correct but it is different from the end of [11.8.2] on the same page

> When propagation reaches a simple operand as defined in 11.5, then that operand shall be converted to the propagated type and size. If the operand shall be extended, then it shall be sign-extended only if the propagated type is signed.

and that leads to incorrect evaluation of Slang on an expression
```sv
1'b0 ? 7'd100: 3'sb101
```
where 3'sb101 should have unsigned extension not signed extension. Some flags in ConversionExpression may be needed to handle different conversion rules in different context.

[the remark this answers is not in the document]
MikePopoloski: Hmm, those are some good points. I'll think about this some more and get back to you.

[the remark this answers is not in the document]
MikePopoloski: Yeah, I think it makes sense to have this as a separate clause, but I'd like it to be in ::convert so external callers can access it. That could either be via an isImplicit boolean parameter that gets passed in, or an additional enum. I haven't thought too much yet about the potential bug you posted about.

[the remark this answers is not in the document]
thingkingland: Is it ok for me to add a flag in ConversionExpression in addition to isImplicit
```c++
  bool isImplicit;
  bool isBitstreamCast;  // new
```
and create a bitmask\<ConversionFlags\> type to a pass as an optional parameter to ::convert?
With a single bitmask\<ConversionFlags\> parameter, ::convert can accept future enhancements beyond what is needed today. Then from ConversionExpression::evalImpl I can create a bitmask\<ConversionFlags\>  from what ConversionExpression has, pass it to ::convert, and let ::convert do all the work.
With this mechanism, inside ::convert, I can do
```c++
     if (flag & ConversionFlags::BitstreamCast) {
         return to.bitstreamCast(value);
     }
     if ((flag & ConversionFlags::Propagate) && value.isInteger() && to.isIntegral())
        value.integer().setSigned(to.isSigned()); // Fix bug on 1'b0 ? 7'd100: 3'sb101
```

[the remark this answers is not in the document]
MikePopoloski: It shouldn't be a bitmask right? Just a normal enum? The options are mutually exclusive.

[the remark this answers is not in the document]
thingkingland: Yes, an enum should be sufficient, and a bitmask is not necessary.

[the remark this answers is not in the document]
MikePopoloski: Ok, that all sounds good to me.

include/slang/binding/AssignmentExpressions.h
```diff
@@ -65,15 +65,19 @@ class AssignmentExpression : public Expression {
 struct CastExpressionSyntax;
 struct SignedCastExpressionSyntax;
 
+enum class ConvertCast : uint8_t { Propagate, Implicit, BitstreamCast, Explicit };
```
MikePopoloski, line 68: Could you call this ConversionKind instead?

include/slang/binding/AssignmentExpressions.h
```diff
@@ -90,7 +94,8 @@ class ConversionExpression : public Expression {
                                   const BindContext& context);
 
     static ConstantValue convert(EvalContext& context, const Type& from, const Type& to,
-                                 SourceRange sourceRange, ConstantValue&& value);
+                                 SourceRange sourceRange, ConstantValue&& value,
+                                 ConvertCast cast = ConvertCast::Implicit);
```
MikePopoloski, line 98: I wouldn't default this parameter. We should force callers to say what kind of conversion they are doing.

include/slang/binding/AssignmentExpressions.h
```diff
@@ -65,15 +65,19 @@ class AssignmentExpression : public Expression {
...
+
 /// Represents a type conversion expression (implicit or explicit).
 class ConversionExpression : public Expression {
 public:
-    bool isImplicit;
+    bool isImplicit() const { return cast_ <= ConvertCast::Implicit; }
+    void setCast(ConvertCast);
+    ConvertCast cast() const { return cast_; }
 
     ConversionExpression(const Type& type, bool isImplicit, Expression& operand,
```
MikePopoloski, line 77: Can you make this take the ConversionKind directly during construction? No need for the boolean anymore.

include/slang/binding/AssignmentExpressions.h
```diff
@@ -65,15 +65,19 @@ class AssignmentExpression : public Expression {
...
 struct SignedCastExpressionSyntax;
 
+enum class ConvertCast : uint8_t { Propagate, Implicit, BitstreamCast, Explicit };
+
 /// Represents a type conversion expression (implicit or explicit).
 class ConversionExpression : public Expression {
 public:
-    bool isImplicit;
+    bool isImplicit() const { return cast_ <= ConvertCast::Implicit; }
+    void setCast(ConvertCast);
```
MikePopoloski, line 74: You can make the conversionKind a free member, no need to wrap it in simple getters / setters.

source/binding/AssignmentExpressions.cpp
```diff
@@ -547,18 +550,35 @@ Expression& ConversionExpression::fromSyntax(Compilation& compilation,
     return *result;
 }
 
+void ConversionExpression::setCast(ConvertCast cast) {
+    ASSERT((cast_ == ConvertCast::Implicit || cast_ == ConvertCast::Explicit) &&
```
MikePopoloski, line 554: I'm not sure why this logic is necessary. I think we should only determine the conversion kind when we create the ConversionExpression and then nobody should really be modifying it later.

source/symbols/Type.cpp
```diff
@@ -481,24 +486,55 @@ bool Type::isAssignmentCompatible(const Type& rhs) const {
     return false;
 }
 
-bool Type::isCastCompatible(const Type& rhs) const {
+// Return 2 for bit-stream casting, 1 for valid explicit cast, 0 for invalid type cast
```
MikePopoloski, line 489: Rather than using an int return value here, which isn't self documenting, I think it would make more sense to nest this function in a higher-level one called something like isBitstreamCastable(). Similar to how isCastCompatible calls into isAssignmentCompatible which calls into isEquivalent, isBitstreamCastable would call into isCastCompatible first.

source/symbols/TypeHelpers.h
```diff
@@ -0,0 +1,297 @@
...
+#pragma once
+
+#include <numeric>
+
+#include "slang/symbols/AllTypes.h"
+#include "slang/symbols/VariableSymbols.h"
+
+namespace slang {
+
+/// Return {a, b} to model the size of a dynamic size with "aR+b" where "a" and "b" are compile-time
```
MikePopoloski, line 16: Nice documentation.

include/slang/binding/AssignmentExpressions.h
thingkingland: Yes, I will change it.

include/slang/binding/AssignmentExpressions.h
thingkingland: Yes, I will make the change.

include/slang/binding/AssignmentExpressions.h
thingkingland: I guess you meant to make this member "public" and other classes can freely modify it directly. I can make the change.

source/symbols/Type.cpp
thingkingland: How about making "isBitstreamCastable" and "isCastCompatible" two independent functions without wrapping each other? The purpose is to distinguish them and if "isBitstreamCastable" calls "isCastCompatible", the return value of "isBitstreamCastable" cannot be bool and must have 3 values, which still needs extra documenting. There are two places calling "isCastCompatible" now. With two independent functions,
```c++
            type.isCastCompatible(*rt) ? diag::NoImplicitConversion : diag::BadAssignment;
```
can become
```c++
            type.isCastCompatible(*rt) || type.isBitstreamCapaible(*rt)? diag::NoImplicitConversion : diag::BadAssignment;
```
and
```c++
    auto castKind = type.isCastCompatible(*operand.type);
    if (!castKind) {
        auto& diag = context.addDiag(diag::BadConversion, syntax.apostrophe.location());
        ...
    if (castKind > 1)

```
can become something like
```c++
    bool isBitstreamCast = false;
    bool isCastCompatible =  type.isCastCompatible(*operand.type);
    if (!isCastCompatible)
        isBitstreamCast = type.isBitstreamCastable(*operand.type);
    if (!isCastCompatible && !isBitstreamCast) {
       auto& diag = context.addDiag(diag::BadConversion, syntax.apostrophe.location());
        ...
    if (isBitstreamCast)
```

include/slang/binding/AssignmentExpressions.h
thingkingland: This is possible, but requires more messy change on existing code. The two ConversionExpression::fromSyntax functions at lines 522 and 470 of AssignmentExpressions.cpp both construct ConversionExpression first, then analyze compatibility,  return badExpr if not compatible. In other words, the ConversionKind data is available after construction. To make ConversionKind available during construction, the code logic in the two fromSyntax function needs more changes.
My original intention is to minimize modification of existing good code. If you still think it is fine to change the two fromSyntax functions significantly, I can make the changes.

MikePopoloski: Yeah, I think that's fine. Having class properties be immutable after construction is pretty valuable, worth some refactoring I think.

source/symbols/Type.cpp
MikePopoloski: Yeah, that works too.

include/slang/binding/AssignmentExpressions.h
thingkingland: Yes, I initialize ConversionKind in in the constructor and it does not need a change later. The refactoring of fromSyntax functions turns out to be simple. I only need to modify a few expressions and the existing code logic and flow remain intact.

include/slang/binding/AssignmentExpressions.h
thingkingland: I make it a public const member.
```c++
public:
    const ConversionKind castKind;
```
which cannot be modified after construction.

source/binding/AssignmentExpressions.cpp
thingkingland: I remove this. After setting ConversionKind in construction, this kind of code is totally unnecessary.

thingkingland:
I git merge upstream, resolve only 1 simple conflict, and push it. I visually inspect the merged result for its difference to upstream/master and my final commit, and it looks good.
"git rebase" has some troubles to produce good results this time. If the merged result is not good enough, and you still need a rebase, please let me know and I will try rebase again.

MikePopoloski:
Thanks, this is good.

PR #317, opened 2020-08-30
Fix incorrect constant evaluation for mixed unknowns or mixed signedness

thingkingland:
This is to fix some incorrect values of constant evaluation for some test cases with mixed unknowns or signedness. Unit tests showing the problems are included.
1. Reduction OR should be one as long as there is a one bit, regardless whether there are unknowns. An unknown value was incorrectly returned whenever there was an unknown bit.
```c++
    CHECK(session.eval("| 3'b0x1").integer() == 1);
```
A fix in SVInt::reductionOr() improves it.
```c++
    if (unknownFlag)
        return countOnes() > 0 ? logic_t(true) : logic_t::x;
```
2. The four bitwise operators crashed when the first operand had unknowns but the second operand didn't, and ignored the unknown in the second when the first did not have unknowns.
```c++
    CHECK_THAT(session.eval("3'b000 ^ 3'b0x1").integer(), exactlyEquals("3'b0x1"_si));
    CHECK_THAT(session.eval("3'b0x1 ^ 3'b000").integer(), exactlyEquals("3'b0x1"_si));
    CHECK_THAT(session.eval("3'b000 | 3'b0x1").integer(), exactlyEquals("3'b0x1"_si));
    CHECK_THAT(session.eval("3'b0x1 | 3'b000").integer(), exactlyEquals("3'b0x1"_si));
    CHECK_THAT(session.eval("3'b111 & 3'b0x1").integer(), exactlyEquals("3'b0x1"_si));
    CHECK_THAT(session.eval("3'b0x1 & 3'b111").integer(), exactlyEquals("3'b0x1"_si));
    CHECK_THAT(session.eval("3'b000 ~^ 3'b0x1").integer(), exactlyEquals("3'b1x0"_si));
    CHECK_THAT(session.eval("3'b0x1 ^~ 3'b000").integer(), exactlyEquals("3'b1x0"_si));
```
A new function SVInt::makeUnknown() is created to make the first operand unknown when the second has unknowns. Four functions are revised to handle unknowns better.
```c++
    SVInt operator&(const SVInt& rhs) const;
    SVInt operator|(const SVInt& rhs) const;
    SVInt operator^(const SVInt& rhs) const;
    SVInt xnor(const SVInt& rhs) const;
```
3. SVInt::reductionAnd() has a similar fix like SVInt::reductionOr(), but it is not sufficient to fix
```c++
    CHECK(session.eval("& 3'b0x1").integer() == 0);
```
The x/z extension in SVInt::fromPow2Digits
```c++
        // If the most significant bit is X or Z, we need to extend that out to the full range.
```
is not supposed to apply on 3'b0x1 whose msb is 0, but because NumberParser::addDigit removed all leading zeros, it incorrectly transformed 3'b0x1 into 3'bxx1. The fix in NumberParser::addDigit
```c++
    if (digit.isUnknown())
        hasUnknown = true; // Keep one leading zero, if any, for msb extension
    else {
        ASSERT(digit.value < maxValue);
        if (digits.size() == 1 && digits.front().value == 0) {
            if (digit.value == 0)
                return; // at most one leading zero
            else
                digits.pop(); // If first nonzero not unknown, no leading zeros
        }
    }
    digits.append(digit);
```
preserves one leading zero if the first nonzero bit is an unknown. If the first nonzero bit is not unknown, all leading zeros are still removed as before.

4. Context-determined sign extension should use propagated type, and 3'sb101 should have unsigned extension below:
```c++
    CHECK(session.eval("1'b0 ? 7'd100: 3'sb101").integer() == 5);
```
In ConstantValue::convertToInt:
```c++
    // [11.8.3] says that during an assignment we sign extend iff the rhs is signed.
    // That means we should resize first, and only then change the sign flag if desired.
```
is correct in its context but [11.8.2] has a different rule:

> When propagation reaches a simple operand as defined in 11.5, then that operand shall be converted to the propagated type and size. If the operand shall be extended, then it shall be sign-extended only if the propagated type is signed.

The solution is to mark in Expression::PropagationVisitor that the conversion is a propagated type
```c++
       if (needConversion) {
            Expression::selfDetermined(context, result);
            result = context.scope.getCompilation().emplace<ConversionExpression>(
                newType, ConversionKind::Propagated, expr, expr.sourceRange);
        }
```
and in ConversionExpression::convert, signedness is propagated for this kind of conversion
```c++
    if (castKind == ConversionKind::Propagated && value.isInteger() && to.isIntegral())
        value.integer().setSigned(to.isSigned());
```
5. == operator with unknowns should be fixed too.
```c++
   if (unknownFlag || rhs.unknownFlag)
        return !(*this ^ rhs).reductionOr();
```
6. The following 4 logic operators use reductionOr instead of == 0.
```c++
    logic_t operator!() const { return !reductionOr(); }
    logic_t operator&&(const SVInt& rhs) const { return reductionOr() && rhs.reductionOr(); }
    logic_t operator||(const SVInt& rhs) const { return reductionOr() || rhs.reductionOr(); }
    explicit operator logic_t() const { return reductionOr(); }
```
reductionOr delivers the same result as ==, but is more efficient than ==, which has overhead to create temporary second operand and check/adjust lengths/signedness of two operands. With the fix in reductionOr and == on unknowns, the following logic operators with unknowns work correctly now.
```c++
    CHECK(session.eval("3'b0x1 || 1'b0").integer() == 1);
    CHECK(session.eval("3'b0x1 && 1'b1").integer() == 1);
    CHECK(session.eval("3'b0x1 == 0").integer() == 0);
    CHECK(session.eval("2'b1x ?16'h1234:16'h7890").integer() == 4660);
```

### DIFF
--- a/source/binding/AssignmentExpressions.cpp
+++ b/source/binding/AssignmentExpressions.cpp
@@ -579,6 +579,9 @@ ConstantValue ConversionExpression::convert(EvalContext& context, const Type& fr
         return cv;
     }
 
+    if (castKind == ConversionKind::Propagated && value.isInteger() && to.isIntegral())
+        value.integer().setSigned(to.isSigned());
+
     if (to.isIntegral())
         return value.convertToInt(to.getBitWidth(), to.isSigned(), to.isFourState());
 
--- a/source/binding/AssignmentExpressions.cpp
+++ b/source/binding/AssignmentExpressions.cpp
@@ -579,6 +579,9 @@ ConstantValue ConversionExpression::convert(EvalContext& context, const Type& fr
         return cv;
     }
 
+    // [11.8.2] last bullet says: the operand shall be sign-extended only if the propagated type is
+    // signed. It is different from [11.8.3] ConstantValue::convertToInt uses.
+    // ConversionKind::Propagated marked in Expression::PropagationVisitor
     if (castKind == ConversionKind::Propagated && value.isInteger() && to.isIntegral())
         value.integer().setSigned(to.isSigned());
 
--- a/source/numeric/SVInt.cpp
+++ b/source/numeric/SVInt.cpp
@@ -861,7 +861,7 @@ SVInt SVInt::pow(const SVInt& rhs) const {
 
 logic_t SVInt::reductionAnd() const {
     if (unknownFlag)
-        return logic_t::x;
+        return countZeros() > 0 ? logic_t(false) : logic_t::x;
 
     uint64_t mask;
     bitwidth_t bitsInMsw;
@@ -880,7 +880,7 @@ logic_t SVInt::reductionAnd() const {
 
 logic_t SVInt::reductionOr() const {
     if (unknownFlag)
-        return logic_t::x;
+        return countOnes() > 0 ? logic_t(true) : logic_t::x;
 
     if (isSingleWord())
         return logic_t(val != 0);
@@ -1059,17 +1059,30 @@ SVInt& SVInt::operator&=(const SVInt& rhs) {
             return *this &= rhs.extend(bitWidth, bothSigned);
     }
 
+    if (!hasUnknown() && rhs.hasUnknown())
+        makeUnknown();
     if (isSingleWord())
         val &= rhs.val;
     else {
         uint32_t words = getNumWords(bitWidth, false);
         if (unknownFlag) {
-            for (uint32_t i = 0; i < words; i++)
-                pVal[i + words] = (pVal[i + words] | rhs.pVal[i + words]) &
-                                  (pVal[i + words] | pVal[i]) & (rhs.pVal[i + words] | rhs.pVal[i]);
+            if (rhs.isSingleWord()) {
+                pVal[1] &= rhs.val;
+                pVal[0] = ~pVal[1] & pVal[0] & rhs.val;
+            }
+            else {
+                if (rhs.hasUnknown())
+                    for (uint32_t i = 0; i < words; i++)
+                        pVal[i + words] = (pVal[i + words] | rhs.pVal[i + words]) &
+                                          (pVal[i + words] | pVal[i]) &
+                                          (rhs.pVal[i + words] | rhs.pVal[i]);
+                else
+                    for (uint32_t i = 0; i < words; i++)
+                        pVal[i + words] &= rhs.pVal[i];
 
-            for (uint32_t i = 0; i < words; i++)
-                pVal[i] = ~pVal[i + words] & pVal[i] & rhs.pVal[i];
+                for (uint32_t i = 0; i < words; i++)
+                    pVal[i] = ~pVal[i + words] & pVal[i] & rhs.pVal[i];
+            }
         }
         else {
             for (uint32_t i = 0; i < words; i++)
@@ -1077,6 +1090,7 @@ SVInt& SVInt::operator&=(const SVInt& rhs) {
         }
     }
     clearUnusedBits();
+    checkUnknown();
     return *this;
 }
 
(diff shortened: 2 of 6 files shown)


source/numeric/SVInt.cpp
```diff
@@ -1059,24 +1059,38 @@ SVInt& SVInt::operator&=(const SVInt& rhs) {
...
-            for (uint32_t i = 0; i < words; i++)
-                pVal[i + words] = (pVal[i + words] | rhs.pVal[i + words]) &
-                                  (pVal[i + words] | pVal[i]) & (rhs.pVal[i + words] | rhs.pVal[i]);
+            if (rhs.isSingleWord()) {
+                pVal[1] &= rhs.val;
+                pVal[0] = ~pVal[1] & pVal[0] & rhs.val;
+            }
+            else {
+                if (rhs.hasUnknown())
+                    for (uint32_t i = 0; i < words; i++)
```
MikePopoloski, line 1075: Can you please use braces around any if or for body that spans more than one line, even if it is logically just a single statement.

source/numeric/SVInt.cpp
```diff
@@ -1685,7 +1726,7 @@ SVInt& SVInt::assignSlowCase(const SVInt& rhs) {
 
 logic_t SVInt::equalsSlowCase(const SVInt& rhs) const {
     if (unknownFlag || rhs.unknownFlag)
```
MikePopoloski, line 1728: Can you add a comment saying that if there are unknowns we can't know whether the numbers are equal, but we might be able to know if they are definitely not equal, which is what the xor + reductionOr is doing.

tests/unittests/EvalTests.cpp
```diff
@@ -1684,3 +1684,39 @@ typedef struct { logic[-2:5] a[][]; string b; bit c[$]; } f;
...
+
+    // logical and reduction operators on mixed unknowns
+    CHECK(session.eval("! 3'b0x1").integer() == 0);
+    CHECK(session.eval("| 3'b0x1").integer() == 1);
+    CHECK(session.eval("& 3'b0x1").integer() == 0);
+    CHECK_THAT(session.eval("& 3'bx1").integer(), exactlyEquals(SVInt(logic_t::x)));
+    CHECK(session.eval("3'b0x1 || 1'b0").integer() == 1);
+    CHECK(session.eval("3'b0x1 && 1'b1").integer() == 1);
+
+    // Equaulity operators with uknowns
```
MikePopoloski, line 1714: "Equality" and "unknowns"

source/binding/AssignmentExpressions.cpp
```diff
@@ -579,6 +579,9 @@ ConstantValue ConversionExpression::convert(EvalContext& context, const Type& fr
         return cv;
     }
 
+    if (castKind == ConversionKind::Propagated && value.isInteger() && to.isIntegral())
```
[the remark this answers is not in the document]
thingkingland: Yes, I add comments
```c++
    // [11.8.2] last bullet says: the operand shall be sign-extended only if the propagated type is signed.
    // It is different from [11.8.3] ConstantValue::convertToInt uses.
    // ConversionKind::Propagated marked in Expression::PropagationVisitor
```

source/numeric/SVInt.cpp
thingkingland: Yes, for example, the braces are inserted for
```c++
                if (rhs.hasUnknown()) {
                    for (uint32_t i = 0; i < words; i++) {
                        pVal[i + words] = (pVal[i + words] | rhs.pVal[i + words]) &
                                          (pVal[i + words] | pVal[i]) &
                                          (rhs.pVal[i + words] | rhs.pVal[i]);
                    }
                }
                else {
                    for (uint32_t i = 0; i < words; i++)
                        pVal[i + words] &= rhs.pVal[i];
                }
```

source/numeric/SVInt.cpp
thingkingland: Yes, I add
```c++
    if (unknownFlag || rhs.unknownFlag) {
        // We can't know whether the numbers are definitely equal, but if there is a 0/1 pair, it is
        // definitely not equal. xor detects 0/1 pairs for each bit and !reductionOr collects all
        // pairs.
        return !(*this ^ rhs).reductionOr();
    }
```

tests/unittests/EvalTests.cpp
thingkingland: I correct the typos.

PR #320, opened 2020-09-02
Fix shift operator constant evaluation with negative right operand or unknowns

thingkingland:
This is to fix problems for shift operator constant evaluation.
1. SV-2017 standard [11.4.10] says the right operand of shift operators is always treated as an unsigned number.
```c++
    CHECK_THAT(session.eval("30'b10xz0110111<<-3'sb1").integer(),
               exactlyEquals("30'b00000000000010xz01101110000000"_si));
```
The shift amount should be $unsigned(-3'sb1) = 7, but was incorrectly calculated to UINT_MAX. The fix creates a new private SVInt function
```c++
    // $unsigned(*this) value truncated at bitwidth_t::max()
    bitwidth_t unsignedAmount() const;
```
to calculate the correct shift amount.

2. For arithmetic right shift, when the left operand has unknowns and the right doesn't, the result is not reliable.
```c++
    CHECK_THAT(session.eval("5'sbz001x>>>2").integer(), exactlyEquals("5'bzzz00"_si));
    CHECK_THAT(session.eval("129'sb0xzxz>>>1").integer(), exactlyEquals("129'sb0xzx"_si));
```
The code in SVInt::ashr
```c++
    if (contractedWidth <= 64 && getNumWords() > 1) {
        // signExtend won't be safe as it will assume it is operating on a single-word
        // input when it isn't, so let's manually take care of that case here.
        SVInt result = SVInt::allocUninitialized(bitWidth, signFlag, unknownFlag);
        uint64_t newVal = tmp.pVal[0] << (SVInt::BITS_PER_WORD - contractedWidth);
        result.pVal[0] = uint64_t((int64_t)newVal >> (SVInt::BITS_PER_WORD - contractedWidth));
        for (size_t i = 1; i < getNumWords(); ++i) {
            // sign extend the rest based on original sign
            result.pVal[i] = (val & (1ULL << 63)) ? 0 : ~0ULL;
        }
        result.clearUnusedBits();
        return result;
    }
    // Pretend our width is just the width we shifted to, then signExtend
    tmp.bitWidth = contractedWidth;
```
did not handle unknowns. When contractedWidth > 64, unknowns might be incorrect in sext. The comments said "signExtend won't be safe", but the real problem was "tmp.bitWidth = contractedWidth;" corrupted pVal/val memory when the number of words changes because of reducing bitWidth. A solution is to replace the above problematic code with
```c++
   tmp = tmp.trunc(contractedWidth);
```
When SVInt::trunc sets bitWidth to a new value, it updates pVal/val accordingly.

3. This led to another bug in SVInt::trunc that cleared signFlag for isSingleWord().
```c++
    if (isSingleWord()) {
        uint64_t mask = bits == 64 ? UINT64_MAX : (1ull << bits) - 1;
        return SVInt(bits, val & mask, false);
    }
```
This did not look intentional, because if not  isSingleWord(), SVInt::trunc preserves signFlag. The fix preserves signFlag for isSingleWord().

### DIFF
--- a/include/slang/numeric/SVInt.h
+++ b/include/slang/numeric/SVInt.h
@@ -556,7 +556,7 @@ private:
     void checkUnknown();
     void makeUnknown();
 
-    // $unsigned(*this) value truncated at bitwidth_t::max()
+    // $unsigned(*this) value saturated at bitwidth_t::max()
     bitwidth_t unsignedAmount() const;
 
     static constexpr uint32_t whichWord(bitwidth_t bitIndex) { return bitIndex / BITS_PER_WORD; }
--- a/include/slang/numeric/SVInt.h
+++ b/include/slang/numeric/SVInt.h
@@ -556,6 +556,9 @@ private:
     void checkUnknown();
     void makeUnknown();
 
+    // $unsigned(*this) value truncated at bitwidth_t::max()
+    bitwidth_t unsignedAmount() const;
+
     static constexpr uint32_t whichWord(bitwidth_t bitIndex) { return bitIndex / BITS_PER_WORD; }
     static constexpr uint32_t whichBit(bitwidth_t bitIndex) { return bitIndex % BITS_PER_WORD; }
     static constexpr uint64_t maskBit(bitwidth_t bitIndex) { return 1ULL << whichBit(bitIndex); }
--- a/source/numeric/SVInt.cpp
+++ b/source/numeric/SVInt.cpp
@@ -490,7 +490,7 @@ bitwidth_t SVInt::unsignedAmount() const {
     bitwidth_t bits = getActiveBits();
     if (bits > sizeof(bitwidth_t) * CHAR_BIT)
         return std::numeric_limits<bitwidth_t>::max();
-    return static_cast<bitwidth_t>(isSingleWord() ? val : pVal[0]);
+    return static_cast<bitwidth_t>(*getRawPtr());
 }
 
 SVInt SVInt::shl(const SVInt& rhs) const {
@@ -599,7 +599,10 @@ SVInt SVInt::ashr(bitwidth_t amount) const {
         return *this;
     if (amount >= bitWidth)
         return SVInt(bitWidth, *this >= 0 ? 0 : UINT64_MAX, signFlag);
-    if (!signFlag || !(*this)[int32_t(bitWidth) - 1]) // unsigned or positive
+
+    // !(*this)[int32_t(bitWidth) - 1]) checks msb==0
+    // Not precisely !isNegative() which is msb!=1 or msb==[0xz]
+    if (!signFlag || !(*this)[int32_t(bitWidth) - 1]) // unsigned or nonnegative
         return lshr(amount);
 
     bitwidth_t contractedWidth = bitWidth - amount;
--- a/source/numeric/SVInt.cpp
+++ b/source/numeric/SVInt.cpp
@@ -485,15 +485,24 @@ void SVInt::shrinkToFit() {
         *this = resize(minBits);
 }
 
+bitwidth_t SVInt::unsignedAmount() const {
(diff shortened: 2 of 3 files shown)


source/numeric/SVInt.cpp
```diff
@@ -576,39 +586,26 @@ SVInt SVInt::ashr(const SVInt& rhs) const {
...
 }
 
 SVInt SVInt::ashr(bitwidth_t amount) const {
-    if (!signFlag)
-        return lshr(amount);
     if (amount == 0)
         return *this;
     if (amount >= bitWidth)
         return SVInt(bitWidth, *this >= 0 ? 0 : UINT64_MAX, signFlag);
+    if (!signFlag || !(*this)[int32_t(bitWidth) - 1]) // unsigned or positive
```
MikePopoloski, line 602: You can use isNegative() instead of checking the last bit manually.

source/numeric/SVInt.cpp
```diff
@@ -576,39 +586,26 @@ SVInt SVInt::ashr(const SVInt& rhs) const {
...
-            result.pVal[i] = (val & (1ULL << 63)) ? 0 : ~0ULL;
-        }
-        result.clearUnusedBits();
-        return result;
-    }
+
     // Pretend our width is just the width we shifted to, then signExtend
-    tmp.bitWidth = contractedWidth;
-    return tmp.sext(bitWidth);
+    return lshr(amount).trunc(contractedWidth).sext(bitWidth);
```
MikePopoloski, line 608: Thanks, this is much cleaner.

include/slang/numeric/SVInt.h
```diff
@@ -556,6 +556,9 @@ class SVInt : SVIntStorage {
     void checkUnknown();
     void makeUnknown();
 
+    // $unsigned(*this) value truncated at bitwidth_t::max()
```
MikePopoloski, line 559: s/truncated/saturated

source/numeric/SVInt.cpp
```diff
@@ -485,15 +485,24 @@ void SVInt::shrinkToFit() {
         *this = resize(minBits);
 }
 
+bitwidth_t SVInt::unsignedAmount() const {
+    // [11.4.10] The right operand of shift operators is always treated as an unsigned number
+    bitwidth_t bits = getActiveBits();
+    if (bits > sizeof(bitwidth_t) * CHAR_BIT)
+        return std::numeric_limits<bitwidth_t>::max();
+    return static_cast<bitwidth_t>(isSingleWord() ? val : pVal[0]);
```
MikePopoloski, line 493: You can replace the conditional here with just *getRawPtr()

source/numeric/SVInt.cpp
thingkingland: The purpose here is to detect isPositive(), or msb==0. isNegative()is msb==0. !isNegative() is msb== 0 or x or z.
So !isNegative() is not precisely isPositive(), when taking unknowns into consideration.

include/slang/numeric/SVInt.h
thingkingland: Yes, I will change it.

source/numeric/SVInt.cpp
thingkingland: Yes, I will change it.

source/numeric/SVInt.cpp
MikePopoloski: Ah, I see. That's pretty subtle, can you expand the comment here to indicate what's happening?

thingkingland: I add comments
```c++
    // !(*this)[int32_t(bitWidth) - 1]) checks msb==0
    // Not precisely !isNegative() which is msb!=1 or msb==[0xz]
    if (!signFlag || !(*this)[int32_t(bitWidth) - 1]) // unsigned or nonnegative
```

PR #322, opened 2020-09-09
§11.4.14 Streaming operators (pack/unpack) support

thingkingland:
This is to support SV-2017 [11.4.14] streaming operators. The major C++ code changes include
1. A new C++ class "StreamingConcatenationExpression" is created to represent streaming operator expressions. "sliceSize > 0" means right-to-left (<<), and sliceSize==0 means left-to-right (>>) which ignores slice size.
2. "BindFlags::StreamingAllowed" restricts the use of streaming operators only at assignment-like context or the argument of a bit-stream casting.
3. File "symbols/TypeHelpers.h" is moved to "binding/Bitstream.cpp (h)" so the helper functions can be shared between bit-stream casting and streaming operators.
4. All bitstreamWidth() function return types are changed from bitwidth_t to std::size_t. While each integral/packed type size is represented by uint32_i, an unpacked type will have uint64_t for its total size.
5. "ConversionKind::StreamingConcat" will represent implicit bit-stream casting "dest = {>>{src}}", while "ConversionKind::BitstreamCast" represents explicit bit-stream casting "dest'(src)". They share many code but have slightly different rules. For "dest'(src)", the size requirement is $bits(dest)==$bits(src), while for "dest={>>{src}}", the requirement is $bits(dest) >= $bits(src).
6. There are two TODO items, which can be done in the next step. One is "with" expression, whose use will cause NotYetSupported error. The other is "first-declared member of untagged union", which will cause not bit-stream type error now.


source/binding/OperatorExpressions.cpp
```diff
@@ -1498,6 +1501,146 @@ void ReplicationExpression::serializeTo(ASTSerializer& serializer) const {
...
+                context.addDiag(diag::BadStreamType, arg->sourceRange)
+                    << std::string("expression") << type;
+                return badExpr(compilation, badResult());
+            }
+        }
+
+        buffer.append(arg);
+    }
+
+    // Streaming operator has no meaningful type. Use void as a placeholder.
```
MikePopoloski, line 1583: This diff adds a ton of special casing all over the place for streaming expressions, mostly because of this decision right here. Is it not possible to have the streaming expression determine its type from the assignment-like context? Then there's no need for a new kind of implicit conversion or special assignment rules. Nested streaming concatenations are cast to a packed array of bit using a bit-stream cast, so it seems like it should be possible to get the type there as well.

include/slang/symbols/Type.h
```diff
@@ -56,7 +56,7 @@ class Type : public Symbol {
     bitwidth_t getBitWidth() const;
 
     /// Gets $bits of the type. Returns zero if the type does not have a statically known size.
-    bitwidth_t bitstreamWidth() const;
+    std::size_t bitstreamWidth() const;
```
MikePopoloski, line 59: I think this is a good change, but can you use "size_t" without the std:: prefix, everywhere in the code?

include/slang/binding/Bitstream.h
```diff
@@ -0,0 +1,60 @@
...
+class Expression;
+class ConstantValue;
+class BindContext;
+class EvalContext;
+class SourceRange;
+
+class Bitstream {
+public:
+    /// Compile-time check that the source and destination types have the same dynamic bit-stream
+    /// sizes.
```
MikePopoloski, line 25: Can you add a comment about which types are valid here, since only the ones you've explicitly instantiated in the cpp file will work.

source/binding/Expression.cpp
```diff
@@ -165,7 +165,10 @@ const Expression& Expression::bindLValue(const ExpressionSyntax& lhs, const Type
 const Expression& Expression::bindRValue(const Type& lhs, const ExpressionSyntax& rhs,
                                          SourceLocation location, const BindContext& context) {
     Compilation& comp = context.scope.getCompilation();
-    Expression& expr = create(comp, rhs, context, BindFlags::None, &lhs);
+    bitmask<BindFlags> extraFlags = context.instance || (context.flags & BindFlags::EnumInitializer)
```
MikePopoloski, line 168: Is the streaming operator actually disallowed for enum initializers or ports? Actually, interestingly enough, it seems like streaming operators might not actually be allowed in constant expressions according to the grammar, but that seems like an oversight in the grammar.

source/binding/Bitstream.cpp
```diff
@@ -0,0 +1,664 @@
...
+                                                  : (*iter)->str().length() * CHAR_BIT));
+        ASSERT(iter != packed.cend() && ++iter == packed.cend());
+    }
+    return cv;
+}
+
+bool Bitstream::canBeTarget(const StreamingConcatenationExpression& lhs, const Expression& rhs,
+                            const BindContext& context) {
+    if (rhs.kind != ExpressionKind::Streaming && !rhs.type->isBitstreamType()) {
+        context.addDiag(diag::BadStreamType, rhs.sourceRange) << std::string("source") << *rhs.type;
```
MikePopoloski, line 422: I've been avoiding parameterizing diagnostic messages using English words like this (unless they are already language keywords), since it makes it much harder in the future for someone to translate the diagnostics into another language. I don't know if that would ever actually happen, but it's not much harder to define a few diagnostics instead.

source/binding/OperatorExpressions.cpp
thingkingland: Initially I planned to create a special type for streaming operators too. The closest type is an unpacked struct type. However, struct types needs to be declared in the same scope to be compatible, needs a name, and no existing code to create a struct type except from a syntax. Because it is still quite different from a struct type, doing so may introduce a lot of special casing in type functions of Type.h/Type.cpp for general purposes. The current approach adds casing only around assignments and conversion, and seemed to be less messy than changing type classes/functions.

They are pros and cons between these two approaches. Please let me know if you think the type approach will be better.

include/slang/symbols/Type.h
thingkingland: Yes, I will change them.

include/slang/binding/Bitstream.h
thingkingland: Yes, I will add comments. Only two types Type and StreamingConcatenationExpression for the template.

source/binding/Expression.cpp
thingkingland: This is not precisely defined in the standard.
1. I currently allow streaming operator in parameter initializers. The standard says "assignment" not "assignment-like context". If it were "assignment-like context", constant expression to initialize parameters should be allowed. It depends on interpretation. Let me know your decision. I can change it.
2. I currently disallow it for ports. Again, if it is "assignment-like context", ports should be allowed. The complexity is for instance array, where the type is needed to divide the expression among  instances, and there is no well-defined type for streaming operators.
3. I currently disallow it for enum initializers. There are special code and I have not paid attention to it. When it becomes important to support it, it can be a TODO item.

source/binding/Bitstream.cpp
thingkingland: The diag::BadStreamType have 3 categories for "expression", "target" and "source". I will change it into 3 different diagCodes without the expression/target/source parameter, as you suggested.

source/binding/OperatorExpressions.cpp
thingkingland: For constant expressions, [11.2.1] page 255 says

> Constant expressions can use any of the operators defined in Table 11-1.

and streaming operators are in Table 11-1.

source/binding/Expression.cpp
thingkingland: For constant expressions, [11.2.1] page 255 says

> Constant expressions can use any of the operators defined in Table 11-1.

and streaming operators are in Table 11-1.

include/slang/binding/BindContext.h
```diff
@@ -66,7 +66,12 @@ enum class BindFlags {
 
     /// Expression is for a static variable's initializer. References to automatic
     /// variables will be disallowed.
-    StaticInitializer = 1 << 9
+    StaticInitializer = 1 << 9,
+
+    /// Streaming operator is allowed in assignment target, assignment source and bit-stream casting
```
shuizaiku7, line 71: According to IEEE, a streaming_concatenation can also be a stream_expression in another streaming_concatenation.
Does this need to be handled?

thingkingland: In addition, streaming_concatenation as a stream_expression in another streaming_concatenation is covered in the unit tests.
```c++
        "shortint a; byte b[2]; int c = {<<3{a, {<<5{b}}}};",
        "shortint a; byte b[2]; int c; assign {<<3{{<<5{b}}, a}}=c;",

    {<<16{{<<4{a[0]}},a[1],{<<3{a[2]}}}} = bar;
    {>>{a,{<<{b}},c}} = bar;
```

thingkingland: While the code is correct, you pointed out that the comment is incomplete. I polish the comment to be
```c++
   /// Streaming operator is allowed in assignment target, assignment source, bit-stream casting
   /// argument, or stream expressions of another streaming concatenation. This flag is cleared for
   /// nested subexpressions, unless they are directly parenthesized.
```
and it will be available in the next pull request.

source/binding/Expression.cpp
MikePopoloski: All the simulators I tried allow it for ports, so I think it makes sense to allow it. We can explicitly disallow it for dividing among instance array ports, since that doesn't seem to be supported by other tools (and it's confusing anyway). Also, if we allow it for parameters we should allow it for enum initializers as well.

source/binding/OperatorExpressions.cpp
```diff
@@ -1498,6 +1501,145 @@ void ReplicationExpression::serializeTo(ASTSerializer& serializer) const {
...
+                    return badExpr(compilation, badResult());
+                }
+                sliceSize = sliceExpr.type->bitstreamWidth();
+            }
+            else {
+                optional<int32_t> count = context.evalInteger(sliceExpr);
+                if (!count)
+                    return badExpr(compilation, badResult());
+                if (*count <= 0) {
+                    context.requireGtZero(count, sliceExpr.sourceRange);
```
MikePopoloski, line 1540: You can simply this by checking the result of the requireGtZero call in the if statement instead of checking the value yourself.

source/binding/OperatorExpressions.cpp
MikePopoloski: What I meant was that you could maybe simplify some other uses if you make top-level concatenations have an actual type, for places where you know the assignment target (which you know because it gets passed into this function). Nested concatenations could still have a placeholder type (the void type works ok for this), but then at least a top-level streaming concat expression has a real type associated with it.

source/binding/OperatorExpressions.cpp
```diff
@@ -1498,6 +1501,145 @@ void ReplicationExpression::serializeTo(ASTSerializer& serializer) const {
...
+    }
+    else if (isRightToLeft)
+        sliceSize = 1;
+
+    SmallVectorSized<Expression*, 8> buffer;
+    for (const auto argSyntax : syntax.expressions) {
+        Expression* arg;
+        if (assignmentTarget &&
+            argSyntax->expression->kind == SyntaxKind::StreamingConcatenationExpression)
+            arg = &create(compilation, *argSyntax->expression, context, BindFlags::StreamingAllowed,
```
MikePopoloski, line 1557: You probably don't need to pass the assignmentTarget through here, you aren't using it for anything right?

source/binding/Bitstream.cpp
```diff
@@ -0,0 +1,656 @@
...
+ConstantValue Bitstream::evaluateTarget(const StreamingConcatenationExpression& lhs,
+                                        const Expression& rhs, EvalContext& context) {
+    ConstantValue rvalue = rhs.eval(context);
+    if (!rvalue)
+        return nullptr;
+
+    auto srcSize = rvalue.bitstreamWidth();
+    auto targetWidth = lhs.bitstreamWidth();
+    size_t dynamicSize = 0;
+    if (rhs.kind == ExpressionKind::Streaming) { // srcSize == targetWidth + dynamicSize
```
MikePopoloski, line 622: Comments like this are confusing to me... is it asserting that the expression is true? Or is it saying that you're about to check whether it is true or not, and then issue an error if not?

source/binding/Bitstream.cpp
```diff
@@ -0,0 +1,656 @@
...
+}
+
+/// Performs unpack operation of streaming concatenation target on a bit-stream.
+static bool unpackConcatenation(const StreamingConcatenationExpression& lhs, PackIterator& iter,
+                                const PackIterator iterEnd, bitwidth_t& bitOffset,
+                                size_t& dynamicSize, EvalContext& context,
+                                SmallVector<ConstantValue>* dryRun = nullptr) {
+    for (auto stream : lhs.streams()) {
+        if (stream->kind == ExpressionKind::Streaming) {
+            auto concat = stream->as<StreamingConcatenationExpression>();
```
MikePopoloski, line 569: This should be auto&. It's unfortunate that this compiled, making a copy here is not good.

MikePopoloski (commented):
Ok, I mostly agree with the approach now. I left various comments about improving the implementation.

source/binding/Expression.cpp
thingkingland: It is good to know this information, and streaming operators should be supported in port connection and enum initializer. Can I add a TODO comment in both Expression::bindRValue and Expression::bindLValue for now, and let them be in a future pull request?

include/slang/binding/Bitstream.h
```diff
@@ -0,0 +1,59 @@
...
+    // Only Type and StreamingConcatenationExpression are allowed for T1/T2
+    template<typename T1, typename T2>
+    static bool dynamicSizesMatch(const T1& destination, const T2& source);
+
+    /// Performs a bit-stream cast of @a value  to @a type. If the conversion is not valid,
+    /// returns nullptr (invalid value).
+    static ConstantValue evaluateCast(const Type& type, ConstantValue&&, SourceRange, EvalContext&,
```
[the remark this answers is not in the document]
thingkingland: Yes, I will do that.

source/binding/AssignmentExpressions.cpp
```diff
@@ -303,6 +304,19 @@ Expression& Expression::convertAssignment(const BindContext& context, const Type
             return *result;
         }
 
+        if (expr.kind == ExpressionKind::Streaming) {
+            if (Bitstream::canBeSource(type, expr.as<StreamingConcatenationExpression>(),
+                                       context)) {
+                // Create implicit streaming concatenation conversion
```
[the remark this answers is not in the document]
thingkingland: I change the comment into
```c++
                // Add an implicit bit-stream casting otherwise types are not assignment compatible.
                // The size rule is not identical to explicit bit-stream casting so a different
                // ConversionKind is used.
```

source/binding/AssignmentExpressions.cpp
```diff
@@ -388,7 +405,8 @@ Expression& AssignmentExpression::fromSyntax(Compilation& compilation,
     // "assignment-like context", except if the left hand side does not
     // have a self-determined type. That can only be true if the lhs is
     // an assignment pattern without an explicit type.
```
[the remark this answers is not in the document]
thingkingland: I append the comment there with
```c++
    // However, streaming concatenation does not have an explicit type either here so it is excluded
    // and will lead to diag::AssignmentPatternNoContext error later.
```

source/binding/OperatorExpressions.cpp
thingkingland: I think I need to improve comments in this place.
1. I add
```c++
    // The sole purpose of assignmentTarget here is for Type::isBitstreamType to exclude associative
    // arrays and classes for target/destination. Streaming concatenation is self-determined and its
    // size and type should not be affected by assignmentTarget.
```
We need a way to know whether this streaming concatenation is a source or target, so Type::isBitstreamType can know whether to exclude associative arrays. Ideally, the "Type* assignmentTarget" argument should be replaced with "bool isDestination". However, this function is called through Expression::create or Expression::selfDetermined, and changing its arguments will require the change of those functions too. So I play a trick to reuse "assignmentTarget" for the purpose of isDestination.

2. I replace the comment "Streaming operator has no meaningful type. Use void as a placeholder." with
```c++
    // Streaming concatenation has no explicit type. Use void to prevent unintentional problems when
    // its type is accidentally passed to context-determine expressions.
```
When a streaming concatenation appears as the target of an assignment, the rhs of the assignment should be self-determined. The size, signedness and type compatibility checking of rhs should not be affected by the streaming concatenation, especially when the rhs is context-determined.

source/binding/AssignmentExpressions.cpp
```diff
@@ -428,10 +450,16 @@ Expression& AssignmentExpression::fromComponents(
     if (!lhs.verifyAssignable(context, nonBlocking, assignLoc))
         return badExpr(compilation, result);
 
-    result->right_ =
-        &convertAssignment(context, *lhs.type, *result->right_, assignLoc, lhs.sourceRange);
-    if (result->right_->bad())
-        return badExpr(compilation, result);
+    if (lhs.kind == ExpressionKind::Streaming) {
```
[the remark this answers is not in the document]
thingkingland: This is difficult.
1. Bitstream::canBeTarget needs rhs to check whether rhs size matches and whether rhs has a bit-stream type, while verifyAssignable has no argument to pass rhs.
2. The call
```c++
       result->right_ =
            &convertAssignment(context, *lhs.type, *result->right_, assignLoc, lhs.sourceRange);
```
pass lhs.type without lhs. It is difficult to define and a lhs.type of streaming concatenation for existing convertAssignment to always work, and without passing lhs, convertAssignment cannot know if lhs is a streaming concatenation in order to do special processing. It is difficult to reuse existing convertAssignment for streaming concatenation target.

Streaming concatenation target check require both lhs and rhs together. verifyAssignable has only lhs without rhs. convertAssignment has lhs.type and rhs, but we need lhs not just lhs.type.

source/binding/AssignmentExpressions.cpp
```diff
@@ -513,7 +546,16 @@ Expression& ConversionExpression::fromSyntax(Compilation& compilation,
     }
 
     if (!type->isCastCompatible(*operand.type)) {
-        if (!type->isBitstreamCastable(*operand.type)) {
+        if (operand.kind == ExpressionKind::Streaming) {
+            if (!Bitstream::isBitstreamCast(*type,
+                                            operand.as<StreamingConcatenationExpression>())) {
+                auto& diag = context.addDiag(diag::BadConversion, syntax.apostrophe.location());
+                diag << std::string("streaming concatenation") << *type;
```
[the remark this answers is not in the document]
thingkingland: Yes, I create a new diag code.

> error BadStreamCast "streaming concatenation cannot be converted to type {}"

source/binding/OperatorExpressions.cpp
thingkingland: Yes, this is a good idea. I simplify it into
```c++
                if (!context.requireGtZero(count, sliceExpr.sourceRange))
                    return badExpr(compilation, badResult());
```

source/binding/Bitstream.cpp
thingkingland: The intension is to check the sizes. I improve the comments to
```c++
    if (rhs.kind == ExpressionKind::Streaming) {
        // Check srcSize == targetWidth + dynamicSize, then issue an error if not
    ...
    else {
        // Check srcSize >= targetWidth + dynamicSize, then issue an error if not
```

source/binding/Bitstream.cpp
thingkingland: Yes, I change it for all places. Usually clang should warn about this, I don't know why it doesn't in this case.
My original code referenced ConcatenationExpression::operands(). The existing code has a lot of
```c++
        for (auto operand : operands()) {
```
I assume I should change them too.

source/binding/Expression.cpp
MikePopoloski: Yeah, TODOs are fine.

source/binding/AssignmentExpressions.cpp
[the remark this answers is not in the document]
MikePopoloski: Ok, that makes sense.

source/binding/Bitstream.cpp
MikePopoloski: I was referring to the `auto concat = stream->as<>`, the as() call returns a reference right? So `concat` will be a copy of the object, unless I'm missing something.

thingkingland: Yes, you are right. I will correct it.

thingkingland: I found another line having the same problem and corrected it too.
```c++
        auto& source = rhs.as<StreamingConcatenationExpression>();
```

source/binding/OperatorExpressions.cpp
MikePopoloski: I see, thanks, this helped me understand.

PR #325, opened 2020-09-14
Allow streaming operators in enum initializers and port connections.

thingkingland:
1. Streaming operators are now allowed in enum initializer and port connections. Corresponding unit tests are added.
2. When testing enum initializer, it was found that [6.19]

> Any enumeration encoding value that is outside the representable range of the enum base type shall be an error. For an unsigned base type, this occurs if the cast truncates the value and any of the discarded bits are nonzero. For a signed base type, this occurs if the cast truncates the value and any of the discarded bits are not equal to the sign bit of the result. 

had not been implemented. A new diagnostic code EnumValueOutOfRange is added. Implicit conversion may truncate out of range value silently. In Expression::convertAssignment, the following code checked EnumValueUnknownBits by a redundant evaluation before implicit coversion
```c++
    // If this is an enum initializer and we just discarded unknown bits by
    // implicitly converting to a 2-state result, the standard says we
    // should declare this an error.
    if ((context.flags & BindFlags::EnumInitializer) != 0 && !type.isFourState()) {
        ConstantValue cv = context.tryEval(expr);
        if (cv.isInteger() && cv.integer().hasUnknown()) {
            context.addDiag(diag::EnumValueUnknownBits, expr.sourceRange) << cv << type;
        }
    }
```
In theory, we could tryEval every enum value and check out of range there. However, the solution is not to add implicit conversion  or check EnumValueUnknownBits for enum initializer in Expression::convertAssignment
```c++
        // Do not convert (truncate) enum initializer so out of range value can be checked
        if (!(context.flags & BindFlags::EnumInitializer))
            result = compilation.emplace<ConversionExpression>(type, ConversionKind::Implicit,
                                                               *result, result->sourceRange);
```
The EnumValueUnknownBits check is moved to EnumType::fromSyntax/setInitializer together with new EnumValueOutOfRange check. After the two checkers, EnumType::fromSyntax/setInitializer performs an implicit cast of the value to enum base type, so the end result is the same as before.


include/slang/binding/Expression.h
```diff
@@ -140,7 +140,8 @@ class Expression {
 
     /// Binds the left hand side of an assignment-like expression from the given syntax nodes.
     static const Expression& bindLValue(const ExpressionSyntax& lhs, const Type& rhs,
-                                        SourceLocation location, const BindContext& context);
+                                        SourceLocation location, const BindContext& context,
+                                        bool isInout = false);
```
MikePopoloski, line 144: Can you make this not defaulted? Anyone who calls this function should provide a value for whether it's an inout or not.

source/symbols/AllTypes.cpp
```diff
@@ -311,7 +311,62 @@ const Type& EnumType::fromSyntax(Compilation& compilation, const EnumTypeSyntax&
...
+            if (!value.hasUnknown()) {
+                if (base->isSigned() && exactlyEqual(msb, logic_t(true)))
+                    good = value.countLeadingOnes() >= diff;
+                else
+                    good = value.countLeadingZeros() >= diff;
+            }
+            else {
+                if (!base->isSigned())
+                    msb = logic_t(false);
+                for (auto i = value.getBitWidth() - 1; i >= bitWidth; i--) {
```
MikePopoloski, line 344: This seems like it could be pretty slow, accessing individual bits isn't particularly fast. Maybe this whole block of code should be a method on SVInt, and it can look at the words instead of individual bits?

source/symbols/AllTypes.cpp
```diff
@@ -311,7 +311,62 @@ const Type& EnumType::fromSyntax(Compilation& compilation, const EnumTypeSyntax&
...
+                    if (!exactlyEqual(value[int32_t(i)], msb)) {
+                        good = false;
+                        break;
+                    }
+                }
+            }
+            if (!good) {
+                scope.addDiag(diag::EnumValueOutOfRange, loc) << value << *base;
```
MikePopoloski, line 352: I think this should setValue(nullptr) for the member, otherwise someone could still access this later and see a value that doesn't match the type of the enum.

MikePopoloski (commented):
Eventually I was planning to have a similar truncation check for all expressions, to issue a warning for implicit conversions that discard bits (and then I would just make that an error instead if we were in an enum initializer). What you have here is ok though if you think it's better to keep those checks separate.

include/slang/binding/Expression.h
thingkingland: Yes, I will make the change.

source/symbols/AllTypes.cpp
```diff
@@ -311,7 +311,62 @@ const Type& EnumType::fromSyntax(Compilation& compilation, const EnumTypeSyntax&
         previous = ev.getValue();
         previousRange = ev.getInitializer()->sourceRange;
 
-        checkValue(previous, previousRange.start());
+        if (!previous)
+            return;
+        auto loc = previousRange.start();
+        auto& value = previous.integer();
```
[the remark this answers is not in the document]
thingkingland: I think .isInteger() must be true at that place. If the value is not integral like a real,
```c++
bool checkEnumInitializer(const BindContext& context, const Type& lt, const Expression& rhs) {
    // [6.19] says that the initializer for an enum value must be an integer expression that
    // does not truncate any bits. Furthermore, if a sized literal constant is used, it must
    // be sized exactly to the size of the enum base type. It's not well defined what happens
    // if the sized constant is used further down in some sub-expression, so what we check here
    // is cases where it seems ok to suppress the error:
    // - Unsized literals, variable references
    // - Unary, binary, conditional expressions of the above

    const Type& rt = *rhs.type;
    if (!rt.isIntegral()) {
        context.addDiag(diag::ValueMustBeIntegral, rhs.sourceRange);
        return false;
    }
```
should give a ValueMustBeIntegral error and make Expression::convertAssignment return badExpr, and the code never reaches setInitializer.

source/symbols/AllTypes.cpp
thingkingland: Yes, I add after scope.addDiag for both EnumValueUnknownBits and EnumValueOutOfRange
```c++
                ev.setValue(nullptr);
                previous = nullptr;
```

source/symbols/AllTypes.cpp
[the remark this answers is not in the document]
MikePopoloski: Ah, good point.

source/symbols/AllTypes.cpp
thingkingland: I implement two new methods in SVInt.
```c++
    /// Returns true if all bits in [bitWidth-1:msb] are exactly equal.
    bool isSignExtendedFrom(bitwidth_t msb) const;

    /// If bit msb is nonzero, duplicate it to all bits in [bitWidth-1:msb] .
    void signExtendFrom(bitwidth_t msb);
```
The first one simplifies the EnumValueOutOfRange check
```c++
            if (!base->isSigned()) {
                good = exactlyEqual(value[int32_t(bitWidth)], logic_t(false)) &&
                       value.isSignExtendedFrom(bitWidth);
            }
            else
                good = value.isSignExtendedFrom(bitWidth - 1);
```
The second one replace the code in SVInt::
```c++
   // Pretend our width is just the width we shifted to, then signExtend
    return lshr(amount).trunc(contractedWidth).sext(bitWidth);
```
which allocates/deallocates two extra new SVInt's, with more efficient, sign extension in-place
```c++
    auto tmp = lshr(amount);
    tmp.signExtendFrom(contractedWidth - 1);
    return tmp;
```

source/numeric/SVIntHelpers.h
```diff
@@ -108,6 +108,53 @@ static void signExtendCopy(uint64_t* output, const uint64_t* input, bitwidth_t o
...
+            return false;
+    }
+
+    if (numWords - 1 == word)
+        mask &= topWordMask;
+    mask >>= bit;
+    return (input[word] >> bit) == mask;
+}
+
+// If word.bit is one, make all bits above word.bit become one.
```
MikePopoloski, line 134: "Otherwise, make them all zero."

source/numeric/SVIntHelpers.h
```diff
@@ -108,6 +108,53 @@ static void signExtendCopy(uint64_t* output, const uint64_t* input, bitwidth_t o
...
+static void signExtend(uint64_t* input, uint32_t numWords, uint32_t word, uint32_t bit,
+                       uint64_t topWordMask) {
+    bool sign = (input[word] & (1ULL << bit)) != 0;
+    if (sign && numWords - 1 > word) {
+        input[numWords - 1] = topWordMask;
+        numWords--;
```
MikePopoloski, line 140: Same issue here with wraparound.

source/numeric/SVInt.cpp
```diff
@@ -1504,6 +1505,54 @@ SVInt SVInt::sext(bitwidth_t bits) const {
     return result;
 }
 
+bool SVInt::isSignExtendedFrom(bitwidth_t msb) const {
+    if (msb >= bitWidth - 1)
+        return true;
+
+    if (isSingleWord()) {
+        auto signBits = val >> msb;
+        return !signBits || signBits == (1ULL << (bitWidth - msb)) - 1;
```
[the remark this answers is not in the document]
thingkingland: Yes, this is a bug. I fix it by using getTopWordMask.
```c++
    getTopWordMask(bitsInMsw, maskMsw);

    if (isSingleWord()) {
        auto signBits = val >> msb;
        return !signBits || signBits == maskMsw >> msb;
    }
```

source/numeric/SVIntHelpers.h
```diff
@@ -108,6 +108,53 @@ static void signExtendCopy(uint64_t* output, const uint64_t* input, bitwidth_t o
...
 }
 
+// Check whether all the bits above word.bit is identical to word.bit.
+static bool isSignExtended(const uint64_t* input, uint32_t numWords, uint32_t word, uint32_t bit,
+                           uint64_t topWordMask) {
+    bool sign = (input[word] & (1ULL << bit)) != 0;
+    if (sign && numWords - 1 > word) {
+        if (input[numWords - 1] != topWordMask)
+            return false;
+        numWords--;
```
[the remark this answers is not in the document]
thingkingland: This should not be a problem. The condition
```c++
    if (sign && numWords - 1 > word) {
```
ensures that numWords - 1 > word >=0 or numWords > 1. After numWords--, numWords is still greater than zero.

source/numeric/SVIntHelpers.h
thingkingland: I make the change.

source/numeric/SVIntHelpers.h
thingkingland: This should not be a problem, as explained above.

source/numeric/SVIntHelpers.h
thingkingland: Instead of changing the comment, I change the implementation to match the comment. When the function is called by SVInt::ashr, all bits above msb must be zero. Not doing redundant zero copy improves efficiency a little bit. The comment for SVInt::signExtendFrom is changed to this updated implementation:
```c++
    /// If bit msb is nonzero, duplicate it to all bits in [bitWidth-1:msb].
    void signExtendFrom(bitwidth_t msb);
```

PR #326, opened 2020-09-21
$11.4.14.4 Stream expression "with" range expression support

thingkingland:
This is to to complete a TODO item for stream operator support about "with" range expressions, whose details is described in [11.4.14.4]. The rules look similar to some other sections, but are not identical. For example, "[expression+:expression]" looks like indexed part-select, but the right operand of indexed part-select must be an elaboration-time constant, while "with" expressions do not require it. As a result, functions in RangeSelectExpression and  ElementSelectExpression cannot be directly called, but some of their code is either copy-pasted or referenced.
1. Inside class StreamingConcatenationExpression, struct StreamExpression and struct WithExpression are created to represent "with" expressions.
2. Bitstream.h adds 3 new helper functions:
```c++
    /// Validates stream expression "with" expression range
    static bool validStreamWithRange(const Type& arrayType, WithRangeKind kind,
                                     optional<int32_t> left, optional<int32_t> right,
                                     std::variant<const BindContext*, EvalContext*> context,
                                     SourceRange leftRange, SourceRange rightRange);

    /// Derives constant width from "with" expression range
    static optional<int32_t> withRangeWidth(WithRangeKind kind, optional<int32_t> left,
                                            optional<int32_t> right);

    /// Evaluates stream with expression to a constant range
    static optional<ConstantRange> evaluateWith(
        const Type& arrayType, const StreamingConcatenationExpression::WithExpression& with,
        EvalContext& context);
```
3. Size checks and constant evaluation of streaming concatenation are enhanced to take "with" into consideration. Unit tests are expanded for "with".

### DIFF
--- a/include/slang/binding/OperatorExpressions.h
+++ b/include/slang/binding/OperatorExpressions.h
@@ -261,13 +261,13 @@ public:
 
     struct WithExpression {
         WithRangeKind kind;
-        Expression* left;
-        Expression* right;
-        optional<int32_t> width; // constant width
+        not_null<const Expression*> left;
+        const Expression* right;
+        optional<int32_t> width; // elaboration-time constant width
     };
     struct StreamExpression {
-        Expression* operand;
-        WithExpression* with;
+        not_null<const Expression*> operand;
+        const WithExpression* with;
     };
 
     StreamingConcatenationExpression(const Type& type, size_t sliceSize,
--- a/include/slang/binding/OperatorExpressions.h
+++ b/include/slang/binding/OperatorExpressions.h
@@ -248,6 +248,10 @@ private:
 
 struct StreamingConcatenationExpressionSyntax;
 
+#define RANGE(x) x(Simple) x(IndexedUp) x(IndexedDown) x(Bit)
+ENUM(WithRangeKind, RANGE); // RangeSelectionKind + Bit
+#undef RANGE
+
 /// Represents a streaming concatenation
 class StreamingConcatenationExpression : public Expression {
 public:
@@ -255,13 +259,24 @@ public:
     /// concatenation. Otherwise, it's a right-to-left concatenation.
     const size_t sliceSize;
 
+    struct WithExpression {
+        WithRangeKind kind;
+        Expression* left;
+        Expression* right;
+        optional<int32_t> width; // constant width
+    };
+    struct StreamExpression {
+        Expression* operand;
+        WithExpression* with;
+    };
+
     StreamingConcatenationExpression(const Type& type, size_t sliceSize,
-                                     span<const Expression* const> streams,
+                                     span<const StreamExpression* const> streams,
                                      SourceRange sourceRange) :
         Expression(ExpressionKind::Streaming, type, sourceRange),
         sliceSize(sliceSize), streams_(streams) {}
 
-    span<const Expression* const> streams() const { return streams_; }
+    span<const StreamExpression* const> streams() const { return streams_; }
 
     ConstantValue evalImpl(EvalContext& context) const;
     bool verifyConstantImpl(EvalContext& context) const;
@@ -276,15 +291,21 @@ public:
 
     template<typename TVisitor>
     void visitExprs(TVisitor&& visitor) const {
-        for (auto op : streams())
-            op->visit(visitor);
+        for (auto stream : streams()) {
+            stream->operand->visit(visitor);
+            if (stream->with) {
+                stream->with->left->visit(visitor);
+                if (stream->with->right)
+                    stream->with->right->visit(visitor);
+            }
+        }
     }
 
     bool isFixedSize() const;
     size_t bitstreamWidth() const;
 
 private:
-    span<const Expression* const> streams_;
+    span<const StreamExpression* const> streams_;
 };
 
 struct OpenRangeExpressionSyntax;
--- a/source/binding/Bitstream.cpp
+++ b/source/binding/Bitstream.cpp
@@ -359,7 +359,7 @@ static ConstantValue unpackBitstream(const Type& type, PackIterator& iter,
     if (type.isIntegral()) {
         auto cc = concatPacked(type.getBitWidth(), type.isFourState());
         cc.setSigned(type.isSigned());
-        return std::move(cc);
+        return cc;
     }
 
     if (type.isString()) {
@@ -654,7 +654,7 @@ ConstantValue Bitstream::reOrder(ConstantValue&& value, size_t sliceSize, size_t
         result.emplace_back(std::move(**iter++));
     sliceOrAppend(iter);
 
-    return std::move(result);
+    return result;
 }
 
 /// Performs unpack operation of streaming concatenation target on a bit-stream.
@@ -1009,7 +1009,7 @@ ConstantValue Bitstream::resizeToRange(ConstantValue&& value, ConstantRange rang
             sliceValue.insert(sliceValue.end(), old.cbegin() + lower, old.cbegin() + upper);
             sliceValue.insert(sliceValue.end(), more, defaultValue);
             ASSERT(sliceValue.size() == range.width());
-            return std::move(sliceValue);
+            return sliceValue;
         }
         else {
             ASSERT(value.isQueue());
@@ -1017,7 +1017,7 @@ ConstantValue Bitstream::resizeToRange(ConstantValue&& value, ConstantRange rang
             SVQueue sliceValue(old->cbegin() + lower, old->cbegin() + upper);
             sliceValue.insert(sliceValue.end(), more, defaultValue);
             ASSERT(sliceValue.size() == range.width());
-            return std::move(sliceValue);
+            return sliceValue;
         }
     }
     return std::move(value);
--- a/source/binding/Bitstream.cpp
+++ b/source/binding/Bitstream.cpp
@@ -105,7 +105,7 @@ static std::pair<size_t, size_t> dynamicBitstreamSize(
                 fixedSizeStream = fixedSizeElem * rw;
             }
             else {
-                if (mode != BitstreamSizeMode::DestFill)
+                if (mode != BitstreamSizeMode::DestEmpty)
                     multiplierStream = fixedSizeElem;
                 fixedSizeStream = fixedSizeElem;
             }
@@ -137,7 +137,7 @@ bool Bitstream::dynamicSizesMatch(const T1& destination, const T2& source) {
         dynamicBitstreamSize(source, BitstreamSizeMode::Source);
     auto [destEmptyMultiplier, destEmptyFixedSize] =
         dynamicBitstreamSize(destination, BitstreamSizeMode::DestEmpty);
-    ASSERT(!destEmptyMultiplier && !sourceMultiplier == source.isFixedSize());
+    ASSERT(!destEmptyMultiplier);
 
     if (destEmptyFixedSize >= sourceFixedSize) {
         auto diff = destEmptyFixedSize - sourceFixedSize;
@@ -149,7 +149,6 @@ bool Bitstream::dynamicSizesMatch(const T1& destination, const T2& source) {
 
     auto [destFillMultiplier, destFillFixedSize] =
         dynamicBitstreamSize(destination, BitstreamSizeMode::DestFill);
-    ASSERT(!destFillMultiplier == destination.isFixedSize());
 
     /* Follow IEEE standard to check dynamic-sized types at compile-time.
      // runtime error
@@ -201,8 +200,7 @@ static size_t bitstreamCastRemainingSize(const T& destination, size_t srcSize) {
     // Calculate remaining size to dynamically fill.
     auto [destFillMultiplier, destFillFixedSize] =
         dynamicBitstreamSize(destination, BitstreamSizeMode::DestFill);
-    if (srcSize < destFillFixedSize ||
-        (destFillMultiplier && (srcSize - destFillFixedSize) % destFillMultiplier != 0))
+    if (srcSize < destFillFixedSize || (srcSize - destFillFixedSize) % destFillMultiplier != 0)
         return srcSize + 1;
 
     // Size to fill the first dynamically sized item.
@@ -514,7 +512,7 @@ bool Bitstream::canBeTarget(const StreamingConcatenationExpression& lhs, const E
         sourceWidth = source.bitstreamWidth();
         if (lhs.isFixedSize() && source.isFixedSize())
             good = targetWidth == sourceWidth;
-        else if (!with)
+        else
             good = dynamicSizesMatch(lhs, source);
     }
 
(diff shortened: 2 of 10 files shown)


include/slang/binding/OperatorExpressions.h
```diff
@@ -248,20 +248,35 @@ class ReplicationExpression : public Expression {
...
 /// Represents a streaming concatenation
 class StreamingConcatenationExpression : public Expression {
 public:
     /// The size of the blocks to slice and reorder: if 0, this is a left-to-right
     /// concatenation. Otherwise, it's a right-to-left concatenation.
     const size_t sliceSize;
 
+    struct WithExpression {
+        WithRangeKind kind;
+        Expression* left;
```
MikePopoloski, line 264: I think left is always non-null, so can you use the not_null<> template to indicate that?

source/binding/Bitstream.cpp
```diff
@@ -586,19 +663,20 @@ static bool unpackConcatenation(const StreamingConcatenationExpression& lhs, Pac
...
     for (auto stream : lhs.streams()) {
-        if (stream->kind == ExpressionKind::Streaming) {
-            auto& concat = stream->as<StreamingConcatenationExpression>();
+        auto& operand = *stream->operand;
+        if (operand.kind == ExpressionKind::Streaming) {
+            auto& concat = operand.as<StreamingConcatenationExpression>();
             if (dryRun || !concat.sliceSize) {
-                if (!unpackConcatenation(stream->as<StreamingConcatenationExpression>(), iter,
-                                         iterEnd, bitOffset, dynamicSize, context, dryRun)) {
+                if (!unpackConcatenation(operand.as<StreamingConcatenationExpression>(), iter,
```
MikePopoloski, line 670: Could just use "concat" here instead of casting again.

source/binding/Bitstream.cpp
```diff
@@ -679,9 +822,178 @@ ConstantValue Bitstream::evaluateTarget(const StreamingConcatenationExpression&
...
+    if (iter == iterEnd) {
+        ASSERT(dynamicSize == 0);
+        if (bitOffset > 0) {
+            // Target longer than source
+            context.addDiag(diag::BadStreamSize, lhs.sourceRange) << srcSize + bitOffset << srcSize;
+        }
+    }
+    else if (rhs.kind == ExpressionKind::Streaming) {
+        // Target shorter than source
+        auto tSize = srcSize - (*iter++)->bitstreamWidth() + bitOffset;
```
MikePopoloski, line 838: Is there any chance this size computation can underflow and wrap around?

source/binding/Bitstream.cpp
```diff
@@ -178,15 +202,33 @@ static size_t bitstreamCastRemainingSize(const T& destination, size_t srcSize) {
...
+/// Formats the width of a dynamically-sized bitstream to a formula like 8×n+3 in an error message
+template<typename T>
+static std::string formatWidth(const T& bitstream, bool destination = false) {
+    FormatBuffer buffer;
+    auto mode = destination ? BitstreamSizeMode::DestFill : BitstreamSizeMode::Source;
+    auto [multiplier, fixedSize] = dynamicBitstreamSize(bitstream, mode);
+    if (!multiplier)
+        buffer.format("{}", fixedSize);
+    else if (!fixedSize)
+        buffer.format("{}×n", multiplier);
```
MikePopoloski, line 224: Should probably just use '*' instead of the special character.

source/binding/Bitstream.cpp
```diff
@@ -178,15 +202,33 @@ static size_t bitstreamCastRemainingSize(const T& destination, size_t srcSize) {
...
         return srcSize + 1;
+    }
 
     // Size to fill the first dynamically sized item.
     return srcSize - destFillFixedSize;
 }
 
+/// Formats the width of a dynamically-sized bitstream to a formula like 8×n+3 in an error message
+template<typename T>
+static std::string formatWidth(const T& bitstream, bool destination = false) {
```
MikePopoloski, line 217: It doesn't seem like this parameter should be defaulted. Also, would it make the call sites more clear if you pass the BitstreamSizeMode instead?

source/binding/Bitstream.cpp
```diff
@@ -622,13 +700,76 @@ static bool unpackConcatenation(const StreamingConcatenationExpression& lhs, Pac
...
+                if (dynamicSize > 0 && !stream->with->width) {
+                    auto withSize = elemType->bitstreamWidth() * with.width();
+                    if (withSize >= dynamicSize)
+                        dynamicSize = 0;
+                    else
+                        dynamicSize -= withSize;
+                }
+            }
+            auto rvalue = unpackBitstream(
+                !stream->with
```
MikePopoloski, line 722: This expression is too complicated to pass as a parameter, can you pull it out into its own temporary variable?

source/binding/Bitstream.cpp
```diff
@@ -622,13 +700,76 @@ static bool unpackConcatenation(const StreamingConcatenationExpression& lhs, Pac
...
+                if (stream->with) {
+                    if (with.left != with.right && arrayType.isQueue() && !rvalue.isQueue())
+                        rvalue = constContainer(arrayType, rvalue.elements());
+
+                    if (!arrayType.hasFixedRange()) {
+                        ASSERT(with.left <= with.right);
+                        // When the array is a variable-size array, it shall be resized to
+                        // accommodate the range expression.
+                        const auto oldValue = lvalue.load();
+                        auto minSize = static_cast<uint32_t>(with.right + 1);
```
MikePopoloski, line 744: Is with.right guaranteed to be positive here?

source/binding/Bitstream.cpp
```diff
@@ -679,9 +822,178 @@ ConstantValue Bitstream::evaluateTarget(const StreamingConcatenationExpression&
...
+        if (!index) {
+            context.addDiag(diag::ConstEvalArrayIndexInvalid, expr.sourceRange) << cs << arrayType;
+        }
+        return index;
+    };
+
+    auto cl = evalInt(*with.left);
+    if (!cl)
+        return std::nullopt;
+    decltype(cl) cr = std::nullopt;
```
MikePopoloski, line 966: Using decltype() here seems not necessary, the full type is not long or hard to spell out.

include/slang/binding/OperatorExpressions.h
MikePopoloski: Also can these be stored as const Expressions?

include/slang/binding/OperatorExpressions.h
```diff
@@ -248,20 +248,35 @@ class ReplicationExpression : public Expression {
...
+        Expression* right;
+        optional<int32_t> width; // constant width
+    };
+    struct StreamExpression {
+        Expression* operand;
+        WithExpression* with;
```
MikePopoloski, line 270: These objects aren't very big, perhaps it would be better to make them optional instead of heap allocating?

include/slang/binding/OperatorExpressions.h
```diff
@@ -276,15 +291,21 @@ class StreamingConcatenationExpression : public Expression {
...
+            }
+        }
     }
 
     bool isFixedSize() const;
     size_t bitstreamWidth() const;
 
 private:
-    span<const Expression* const> streams_;
+    span<const StreamExpression* const> streams_;
```
MikePopoloski, line 308: Same here, you can just store these directly in the list instead of indirectly through a pointer.

source/binding/OperatorExpressions.cpp
```diff
@@ -1510,14 +1511,16 @@ Expression& StreamingConcatenationExpression::fromSyntax(
...
+            compilation.getErrorType(), sliceSize, span<const StreamExpression*>(),
+            syntax.sourceRange());
     };
 
     if (!(context.flags & BindFlags::StreamingAllowed)) {
         context.addDiag(diag::BadStreamContext, syntax.operatorToken.location());
         return badExpr(compilation, badResult());
     }
 
+    // slice_size
```
MikePopoloski, line 1523: Can you expand on this comment? It doesn't seem to be adding much in its current form.

source/binding/OperatorExpressions.cpp
```diff
@@ -1596,10 +1676,52 @@ ConstantValue StreamingConcatenationExpression::evalImpl(EvalContext& context) c
...
+                else
+                    cv = std::move(cv).at(size_t(range->left));
+            }
+            else if (range->lower() > 0 || range->width() != cv.size()) {
+                auto upper = static_cast<uint32_t>(range->upper());
+                auto lower = static_cast<uint32_t>(range->lower());
+                auto size = static_cast<uint32_t>(cv.size());
+                auto more = upper >= size ? upper - size + 1 : 0;
+                upper = std::min(upper + 1, size);
+                if (cv.isUnpacked()) {
```
MikePopoloski, line 1705: This code seems duplicated in Bitstream.cpp, can it be shared?

source/binding/OperatorExpressions.cpp
```diff
@@ -1620,19 +1748,29 @@ void StreamingConcatenationExpression::serializeTo(ASTSerializer& serializer) co
...
     if (!streams().empty()) {
         serializer.startArray("streams");
-        for (auto op : streams())
-            serializer.serialize(*op);
+        for (auto op : streams()) {
+            serializer.startObject();
+            serializer.write("operand", op->operand);
+            if (op->with) {
+                serializer.write("withKind", toString(op->with->kind));
+                serializer.write("left", op->with->left);
```
MikePopoloski, line 1756: You're calling these with pointer values... I wonder if this is picking the write<T> overloads that are intended for integers? That seems not good.

include/slang/binding/OperatorExpressions.h
thingkingland: Yes, I add not_null and const where appropriate.
```c++
    struct WithExpression {
        WithRangeKind kind;
        not_null<const Expression*> left;
        const Expression* right;
        optional<int32_t> width; // elaboration-time constant width
    };
    struct StreamExpression {
        not_null<const Expression*> operand;
        const WithExpression* with;
    };
```

source/binding/Bitstream.cpp
thingkingland: Yes, I replace/remove three redundant as\<StreamingConcatenationExpression\>() in the file.

source/binding/Bitstream.cpp
thingkingland: No, this should never happen. "srcSize" is the sum of bitstreamWidth() from the entire vector, and "(*iter++)" is only an element of the vector. I add an assertion to ensure this assumption in the future.
```c++
        ASSERT(srcSize >= (*iter++)->bitstreamWidth());
```

source/binding/Bitstream.cpp
thingkingland: Yes, I replace the fancy unicode multiplication symbol with ASCII character '*'.

source/binding/Bitstream.cpp
thingkingland: Yes, I replace "bool destination = false" with "BitstreamSizeMode mode".

source/binding/Bitstream.cpp
thingkingland: Yes, "with" is the return value of Bitstream::evaluateWith function and at the end of Bitstream::evaluateWith, there are
```c++
    ASSERT(with.kind != WithRangeKind::Bit || result.left == result.right);
    ASSERT(result.left >= 0 && result.right >= 0);
    return result;
```

source/binding/Bitstream.cpp
thingkingland: Yes, I replace it with optional<int32_t>.

include/slang/binding/OperatorExpressions.h
thingkingland: Yes, I replace it with
```c++
        optional<WithExpression> with;
```

thingkingland: However, I am not sure whether it makes a big difference. When "with" is a pointer, I use the bump allocator
```c++
 with = compilation.emplace<WithExpression>(WithExpression{ kind, left, right, width });
```
If "with" is optional, it is heap allocated together in
```c++
    return *compilation.emplace<StreamingConcatenationExpression>(
        compilation.getVoidType(), sliceSize, buffer.ccopy(compilation), syntax.sourceRange());
```
Eventually both methods are heap allocated by the bump allocator. The difference is only between allocated separately or allocated together.

include/slang/binding/OperatorExpressions.h
thingkingland: I can make the change. However, StreamExpression is currently allocated by the bump allocator
```c++
      buffer.append(compilation.emplace<StreamExpression>(StreamExpression{ arg, with }));
```
If I make the change, it will be heap allocated at
```c++
    return *compilation.emplace<StreamingConcatenationExpression>(
```
Can you confirm that it really improves allocation?

source/binding/OperatorExpressions.cpp
thingkingland: I insert some important quotes from IEEE standard at places around relevant code.
```c++
       // The slice_size determines the size of each block. If specified, it may be a constant
       // integral expression or a simple type.

                // It shall be an error for the value to be zero or negative.

          // Left-to-right streaming using >> shall cause the slice_size to be ignored and no
          // re-ordering performed.
```

source/binding/OperatorExpressions.cpp
thingkingland: I fix the bug by dereferencing the pointers.
```c++
            serializer.write("operand", *op->operand);
            if (op->with) {
                serializer.write("withKind", toString(op->with->kind));
                serializer.write("left", *op->with->left);
                if (op->with->right)
                    serializer.write("right", *op->with->right);
            }
```

include/slang/binding/OperatorExpressions.h
thingkingland: I reversed the change for now. I have not fully understood the benefit of the change, because both methods use bump allocator on heap. I will make the change if you confirm there is a benefit doing so.

source/binding/OperatorExpressions.cpp
thingkingland: Yes, I consolidate them into a new helper function declared in Bitstream.h:
```c++
    /// Resize a constant array value from [0:size-1] to [lower:upper]
    static ConstantValue resizeToRange(const Type& arrayType, ConstantValue&& value,
                                       ConstantRange range, bool keepArray = false);
```

include/slang/binding/OperatorExpressions.h
MikePopoloski: It's slightly more efficient to not do the extra pointer indirections, but yeah, the bump allocator definitely makes the heap allocations much less costly. It's fine with me to leave it this way.

Issue #332, opened 2020-11-01
Ability to make "UnknownModule" error suppressible and include those instances in --ast-json output


pistoletpierre:
I'm looking for a quick way to read in a design and get a complete list of instances in it. slang's --ast-json output + a bit of Python has gotten me 90% of the way there. The remaining 10% is dealing with undefined module instantiations. With what I'm doing it's more common than not to see instantiations of modules in a design for which HDL module definitions are either not available or deliberately excluded from a list of source files to read in and elaborate (standard cells, memories, other analog macros, submodules that have already been synthesized, etc...). One option is two make 2 passes: the first to get the list of errors and generate empty module definitions and the second to include the missing modules' generated empty definitions, The other is to add the ability to slang to treat undefined modules as black boxes, which seems like a good thing for it to be able to do.

I had a quick go at picking some potentially low-hanging fruit in [slang/source/symbols/InstanceSymbols.cpp](https://github.com/MikePopoloski/slang/blob/c0cf9a643f05df63268ecbfe561e9af37c9a62a8/source/symbols/InstanceSymbols.cpp#L221), but quickly realized that I was going to have to go down a rabbit hole to hack something together that would compile.

How long do you reckon it would take to become familiar enough with slang's structure to accomplish such a thing?

MikePopoloski:
Is the problem that you still want to see port connection information for the unknown modules in the AST? If not, it seems fairly easy to just suppress the UnknownModule error itself based on some option in the compilation options.

pistoletpierre:
I'm not even interested in the port connection information - just for an "InstanceBody"' to appear in the AST JSON output with "name" and "definition" fields. The "members" field of a missing module's "InstanceBody" could be empty for those instances. Right now if a particular module is missing, the AST JSON written to stdout leaves out any mention of any instances of that module. Compare the following 2 outputs with top.sv being the SV code block that follows the bash code block:

```bash
slang  -D MISSING_MODULE2 --ast-json -  top.sv | grep module2
#vs
slang                     --ast-json -  top.sv | grep module2
```

```systemverilog
// top.sv

`ifndef MISSING_MODULE2
module my_module2 (
  input  logic i_clk,
  output logic o_clk
);
endmodule
`endif

module my_module (
  input  logic i_clk,
  output logic o_clk
);

  my_module2 my_module2_inst (
    .i_clk(i_clk),
    .o_clk(o_clk)
  );

endmodule

module my_module3 (
  input  logic i_clk,
[...]

PR #370, opened 2020-11-28
downgraded UnknownModule from error to warning, added ability to print to stderr

pistoletpierre:
The first commit is RE issue #332 (UnknownModule) (thank you - including the ports was a nice touch) & a piece of cake to decide yes/no. I'm hoping for yes (clearly), as unknown modules are common enough that it seems like a more appropriate default to have as a warning that can be upgraded (-Werror=unknown-module) to an error if needed.

To call the second commit half-baked is generous but I'm submitting it regardless as a working proof of concept in an attempt to encourage you to mull over possible divisions of output streams as opposed to everything going to stdout. As coded in the 2nd commit, it enables one to separate the --ast-json output written to stdout from other outputs such as diagStr (which, when written to stderr, allows one to pipe the --ast-json output (stdout) to a file, preserving the parseability of the json while seeing errors (when present) in the stderr stream)

### DIFF
--- a/include/slang/util/OS.h
+++ b/include/slang/util/OS.h
@@ -62,6 +62,23 @@ public:
             fmt::vprint(format, fmt::make_format_args(args...));
         }
     }
+
+    /// Prints formatted text to stderr, handling Unicode conversions where necessary.
+    template<typename... Args>
+    static void error_print(string_view format, const Args&... args) {
+        fmt::vprint(stderr, format, fmt::make_format_args(args...));
+    }
+
+    /// Prints colored formatted text to stdout, handling Unicode conversions where necessary.
+    template<typename... Args>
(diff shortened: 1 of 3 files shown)


MikePopoloski:
I'm fine with the change to put diagnostic output on stderr. I'm pretty skeptical of making unknown module a warning though... any design that instantiates an unknown module is broken; it can't be simulated or synthesized. You may have some specific use-case where it's fine, but I don't think I would call that a common use-case.

If you're only using the tool to dump AST output, does it matter whether an error is issued or not? You can simply ignore it.

MikePopoloski:
I made diagnostics print to stderr and added a --ignore-unknown-modules flag to suppress errors for unknown instantiations, which I think covers what you need.

pistoletpierre:
Awesome - thank you! I just tried it out and it's fabulous.

PR #391, opened 2021-03-30
can I assist with 'assert property' support?

jankcorn:
I was wondering if I could help get the remaining SVA support into slang.
To make the question more concrete, I did a first, simple pass at changes, which you can look over in the attached patch file (it is definitely _not_ expected that you would merge it 'as is'!).

Summary of changes:
1) added OneStepLiteral, TimingControlExpression, TimingControlExpressionConcatenation to 'ExpressionKind' (in include/slang/binding/Expression.h).
2) added AndSequence, BinarySequenceDelay, IffProperty, ImpliesProperty, IntersectProperty, NonOverlappedFollowedByProperty, NonOverlappedImplicationProperty, OrSequence, OverlappedFollowedByProperty, OverlappedImplicationProperty, SUntilProperty, SUntilWithProperty, ThroughoutSequence, UntilProperty, UntilWithProperty, WithinSequence
to BinaryOperator (this follows their definitions as 'BinaryExpression' in syntax.txt)
3) added class definitions for TimingControlExpression, TimingControlExpressionConcatenationExpression to OperatorExpression.h
4) added class definitions for PropertyDeclarationSymbol, SequenceDeclarationSymbol, ImmediateAssertionMemberSymbol, ConcurrentAssertionMemberSymbol to MemberSymbols.h
5) added temporary hack for parsing 'sequence_match_item' (in grammar.txt) into Parser_expressions.cpp.  [this is really an inappropriate place to put it, but a real fix would be
surgery to syntax.txt]
6) added temporary hack for parsing 'boolean_abbrev' (in grammar.txt) into Parser_expressions.cpp.  [this really needs a change to syntax.txt.  The parsing ambiguity between boolean_abbrev and normal subscripted references looks easily solvable with peek(1) to look ahead in the symbol stream.  yeah!)

I can see that you are busy updating lots of other functionality.  Is it too much of a distraction to handle supporting this functionality via pull requests?  Would you prefer to wait until you have enough time to do it yourself?

If you are willing to go ahead with adding this functionality, I would be quite happy to chop up the changes into more manageable (discrete) pull requests.
From you, I need the 2 changes to syntax.txt for (5) and (6) above (since they will set the direction for how the generated data structures will work) and (eventually) suggestions/directions on structuring the json output artifacts.

Ideas?


MikePopoloski:
Hi, thanks for the interest and I appreciate the pull request. Tackling SV assertions as a first PR is certainly ambitious :P I think you have some of the right ideas here but because this is a large complicated set of features I would like the chance to approach it more systematically, to make sure I'm covering all the bases. As one small example, it's not clear to me that combining sequence and property expressions with regular expressions is the right approach, it's something I did as a quick rough first pass when writing the parser and would like to revisit and rethink.

If you are really excited about contributing I don't want to shut that down, I would instead try to direct you to smaller projects first to learn more of the codebase. If you just want to have assertion support sooner, it's been my plan to finish up and release v0.7 relatively soon and then start work on v0.8, which will be focused on assertions, so that shouldn't be too far away.

jankcorn:
Thanks very much for the quick reply!

I agree that a more systematic approach is best, including separating
the parsing of seq/prop expressions from regular expressions -- they
are quite distinctly separated in the LRM grammar.
It is very, very good to hear that you will be focusing on assertions
in the relatively near future.  Yeah!
I am mainly interested in getting assertion support sooner, so I will
wait until you get to it.
My main long term interest is in building a model checker using slang.

Feeling guilty as a 'taker', if you have some smaller projects related
to synthesizable verilog, I would be happy to try them out to get more
accustomed to the code base.
Is the issues list the best place to start?
Thanks again for all your help.

MikePopoloski:
Ah, no problem. No need to feel guilty either, this project is a hobby project that I like working on. The issue list has some things that need doing but they aren't always well specified enough for someone to look at them and know what to do, they're more like notes to myself. I can try to pick out some things and better define them to give to you, but only if you actively want to work on something. If you have better things to do with your time, you can just wait for me to get around to it :)

jankcorn:
Thanks for the info!

For now, I'll nose around the edges for problems that arise while
running tests here.
(In my experience defining a problem in order to hand it off is almost
as much work as fixing it myself ;-) !)

I was thinking of starting with a random problem I ran across in the
sv_tests:tests/chapter-16/16.10--property-local-var-uvm.sv.
In an instantiation of 'uvm_resource_db', which takes a type
parameter, an error is flagged when a function defined by the module
is referenced:
     error: value of type 'T' cannot be assigned to type 'clk_gen_if'
('T' was the module parameter).
We'll see how much progress I make :)

PR #392, opened 2021-04-15
added test/workaround hack for type parameter matching error

jankcorn:
I added a  file 'tests/regression/type_parameter.v' to demonstrate a compilation error
when using type parameters and generating JSON output. (this arises when
compiling uvm code in sv-tests, using the '--ast-json' flag)

When compiling the test by:
    ./bin/slang --ast-json xx.json ../tests/regression/type_parameter.v
it gets the error:
     error: value of type 'T' cannot be assigned to type 'clk_gen_if'
When compiling the unit test by running 'make test' in the regression build
directory, it compiles without error.

The error appears to be triggered by setting 'disableInstanceCaching' in tools/driver/driver.cpp,
which causes the 2 instantiations of T (one passed into the constructor and the other
in the function parameter type after specialization) to get differing types generated in the parser.

I expect the actual error is in ' Type::isAssignmentCompatible', but
am unsure what is the best fix.  Should this equality check be changed?
Alternatively, should the 'disableInstanceCaching' check be overridden
when instantiating type parameters?

Ideas?
Thanks!

### DIFF
--- a/source/types/Type.cpp
+++ b/source/types/Type.cpp
@@ -540,6 +540,8 @@ bool Type::isAssignmentCompatible(const Type& rhs) const {
 
         auto& lv = l->as<VirtualInterfaceType>();
         auto& rv = r->as<VirtualInterfaceType>();
+        // the following check fails when disableInstanceCaching set from tools/driver astJsonFile flag
+        // (even though the getDefinition() values and parameters match)
         if (&lv.iface != &rv.iface)
             return false;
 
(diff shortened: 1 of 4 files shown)


MikePopoloski:
Thanks for the report. This does appear to be a bug; instance caching should never affect correctness, only performance. This check was a little too lazy and should try harder to match up parameter values.

MikePopoloski:
I added a fix for this, it ended up being fairly simple. Let me know if you run into problems with it.

jankcorn:
Your fix works well.  So fast! :-)

Thanks very much!

PR #406, opened 2021-08-15
Disallow non-blocking assignments in final blocks

jrudess:
This update flags an error if a non-blocking assignment is present in a final block.   However the fix does not resolve the original failing test case due to the final-block also being inside a program block.

https://chipsalliance.github.io/sv-tests-results/#Slang|ivtest|regress-sv_sv_timeunit_prec3b_iv

Please let me know if you'd like any updates done differently.


jrudess:
Woops, just realized you wanted this as a warning.  Will correct.

source/binding/AssignmentExpressions.cpp
```diff
@@ -337,8 +337,14 @@ Expression& Expression::convertAssignment(const BindContext& context, const Type
 Expression& AssignmentExpression::fromSyntax(Compilation& compilation,
                                              const BinaryExpressionSyntax& syntax,
                                              const BindContext& context) {
-    if (!context.flags.has(BindFlags::AssignmentAllowed)) {
-        if (!context.flags.has(BindFlags::NonProcedural) &&
+
+    bool isNonBlocking = syntax.kind == SyntaxKind::NonblockingAssignmentExpression;
+    bool nonBlockInFinal = context.flags.has(BindFlags::Final) && isNonBlocking;
+
+    if (!context.flags.has(BindFlags::AssignmentAllowed) || nonBlockInFinal) {
```
MikePopoloski, line 344: I think this would be a little bit more clear if you just check nonBlockInFinal in a separate if condition; there isn't much reason to combine them into one here since you just check the condition again inside.

source/binding/CallExpression.cpp
```diff
@@ -301,7 +301,7 @@ Expression& CallExpression::fromArgs(Compilation& compilation, const Subroutine&
     if (bad)
         return badExpr(compilation, result);
 
-    if (context.flags.has(BindFlags::FunctionOrFinal) &&
+    if ((context.flags.has(BindFlags::Function) || context.flags.has(BindFlags::Final)) &&
```
MikePopoloski, line 304: Should be able to write this a bit more succinctly as `context.flags.has(BindFlags::Function | BindFlags::Final)` (here and elsewhere in this diff)

MikePopoloski (commented):
This looks pretty good, thanks for the PR. I left a couple of minor comments, but the biggest thing is to add a unit test that exercises the new warning. You could add it to the "Disallowed assignment contexts" test or make a new one, up to you.

MikePopoloski:
Also as you note this won't fix the case in sv-tests; that will need a PR in the sv-tests repo to blacklist that specific test I think.

MikePopoloski:
Thanks, looks good to me.

PR #407, opened 2021-08-17
Implement coverage control functions and predefined macros

jrudess:
This PR adds parsing support for the coverage control functions and coverage macros specified in 20.14, 40.3.1, and 40.3.2

### DIFF
--- a/source/compilation/builtins/CoverageFuncs.cpp
+++ b/source/compilation/builtins/CoverageFuncs.cpp
@@ -1,0 +1,116 @@
+//------------------------------------------------------------------------------
+// CoverageFuncs.cpp
+// Coverage control functions
+//
+// File is under the MIT license; see LICENSE for details
+//------------------------------------------------------------------------------
+#include "slang/binding/SystemSubroutine.h"
+#include "slang/compilation/Compilation.h"
+#include "slang/diagnostics/DeclarationsDiags.h"
+#include "slang/diagnostics/SysFuncsDiags.h"
+
+namespace slang::Builtins {
+
+class CoverageControlFunc : public SimpleSystemSubroutine {
+public:
+    CoverageControlFunc(Compilation& comp) : SimpleSystemSubroutine("$coverage_control", SubroutineKind::Function, 4,
+       { &comp.getIntType(), &comp.getIntType(), &comp.getIntType(), &comp.getStringType() }, comp.getIntType(), false) {};
+
+    ConstantValue eval(EvalContext&, const Args&,
+                       const CallExpression::SystemCallInfo&) const final {
+        return nullptr;
+    }
+};
+
+class CoverageGetMaxFunc : public SimpleSystemSubroutine {
+public:
+    CoverageGetMaxFunc(Compilation& comp) : SimpleSystemSubroutine("$coverage_get_max", SubroutineKind::Function, 3,
+       { &comp.getIntType(), &comp.getIntType(), &comp.getStringType() }, comp.getIntType(), false) {};
+
+    ConstantValue eval(EvalContext&, const Args&,
+                       const CallExpression::SystemCallInfo&) const final {
+        return nullptr;
+    }
+};
+
+class CoverageGet : public SimpleSystemSubroutine {
+public:
+    CoverageGet(Compilation& comp) : SimpleSystemSubroutine("$coverage_get", SubroutineKind::Function, 3,
+       { &comp.getIntType(), &comp.getIntType(), &comp.getStringType() }, comp.getIntType(), false) {};
+
+    ConstantValue eval(EvalContext&, const Args&,
+                       const CallExpression::SystemCallInfo&) const final {
+        return nullptr;
+    }
+};
+
+class CoverageMerge : public SimpleSystemSubroutine {
+public:
+    CoverageMerge(Compilation& comp) : SimpleSystemSubroutine("$coverage_merge", SubroutineKind::Function, 2,
+       { &comp.getIntType(), &comp.getStringType() }, comp.getIntType(), false) {};
+
+    ConstantValue eval(EvalContext&, const Args&,
+                       const CallExpression::SystemCallInfo&) const final {
+        return nullptr;
+    }
+};
+
+class CoverageSave : public SimpleSystemSubroutine {
+public:
+    CoverageSave(Compilation& comp) : SimpleSystemSubroutine("$coverage_save", SubroutineKind::Function, 2,
+       { &comp.getIntType(), &comp.getStringType() }, comp.getIntType(), false) {};
+
+    ConstantValue eval(EvalContext&, const Args&,
+                       const CallExpression::SystemCallInfo&) const final {
+        return nullptr;
+    }
+};
+
+class GetCoverage : public SimpleSystemSubroutine {
+public:
+    GetCoverage(Compilation& comp) : SimpleSystemSubroutine("$get_coverage", SubroutineKind::Function, 0,
+        {}, comp.getRealType(), false) {};
+
+    ConstantValue eval(EvalContext&, const Args&,
+                       const CallExpression::SystemCallInfo&) const final {
+        return nullptr;
+    }
+};
+
+class SetCoverageDbName : public SimpleSystemSubroutine {
+public:
+    SetCoverageDbName(Compilation& comp) : SimpleSystemSubroutine("$set_coverage_db_name", SubroutineKind::Function, 1,
+        { &comp.getStringType() }, comp.getVoidType(), false) {};
+
+    ConstantValue eval(EvalContext&, const Args&,
+                       const CallExpression::SystemCallInfo&) const final {
+        return nullptr;
+    }
+};
+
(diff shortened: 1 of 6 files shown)


MikePopoloski (commented):
Thanks for the contribution. A couple notes:
- These should all inherit from NonConstantFunction I think, as they are not callable in a constant context.
- If you aren't doing anything special in the body of the class then you don't actually need to define a new class; you can just instantiate the base class and pass in the arguments.
- A few of these functions, such as $coverage_control, *do* need special handling of their arguments, as the final argument can be either a string *or* a hierarchical path name. There's no way to define a normal function that can take a hierarchical path name like that as an argument, so you'll need to do the binding manually. You can see as an example the DumpPortsTask which binds a hierarchical reference. You'll need to optionally check for a string here too.

jrudess:
SimpeSystemSubroutine defines bindArguments with 'final'.   For supporting $coverage_control and $coverage_get_max, I can define a new class 'CverageNameOrHierFunc which derives from SystemSubroutine.  This means some slight duplication for the private members in SimpleSystemSubroutine and the eval/verifyConstant implementations.

Or would you prefer removing the 'final' from SimpleSystemSubroutine::bindArgument()?

MikePopoloski:
Deriving from SystemSubroutine is the right call. There isn't much code actually being duplicated, and the simple variant is for "normal" functions that behave like functions a user can write. For special ones deriving from the root class is best.

jrudess:
You mentioned optionally checking for string, but I wasn't sure what to add for this.  The unit test passes  an integer to the argument and it is rejected with an implicit-conversion error.  But the error only mentions that it will accept 'string', so that can be potentially improved to specify either string or hierarchical reference.

source/compilation/builtins/CoverageFuncs.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+void registerCoverageFuncs(Compilation& c) {
+#define REGISTER(name, ...) \
+    c.addSystemSubroutine(std::make_unique<name>(__VA_ARGS__))
+    REGISTER(CoverageNameOrHierFunc, "$coverage_control", c.getIntType(), 3, 4,
+             std::vector{ &c.getIntType(), &c.getIntType(), &c.getIntType(), &c.getStringType() });
+    REGISTER(CoverageNameOrHierFunc, "$coverage_get_max", c.getIntType(), 2, 3,
+             std::vector{ &c.getIntType(), &c.getIntType(), &c.getStringType() });
+
+    REGISTER(NonConstantFunction, "$coverage_get", c.getIntType(), 3,
```
MikePopoloski, line 75: coverage_get also has the same instance or string syntax that the other two do.

source/compilation/builtins/CoverageFuncs.cpp
```diff
@@ -0,0 +1,89 @@
...
+             std::vector{ &c.getIntType(), &c.getIntType(), &c.getIntType() });
+    REGISTER(NonConstantFunction, "$coverage_merge", c.getIntType(), 2,
+             std::vector{ &c.getIntType(), &c.getStringType() });
+    REGISTER(NonConstantFunction, "$coverage_save", c.getIntType(), 2,
+             std::vector{ &c.getIntType(), &c.getStringType() });
+    REGISTER(NonConstantFunction, "$get_coverage", c.getIntType());
```
MikePopoloski, line 81: I think get_coverage is supposed to return a `real`

MikePopoloski (commented):
DumpPortsTask has an example of optionally handling a string or an instance --basically you bind as a HierarchicalReference if the expression is a NameSyntax, otherwise you bind normally. Then when you check argument expressions for correctness, if it's a hierarchical reference you check that the symbol it is pointing to is either 1) a module instance or 2) a value symbol (i.e. net or variable) that is typed with a type that can be a string.

jrudess:
For the test case below, the first system-call should pass, and the 2nd system-call should fail, but both are currently passing.

The hierarchical reference to logic variable `top.x` gets true for `sym.isValue()`, so it doesn't fall through to check if it is an Instance or Module causing a false failure.   In the `sym.isValue()` conditional-branch, would you have expected canBeStringLike() to return false for a variable of logic-type?

```
module A(input logic x);
endmodule

module top;
    logic x;
    string modName = "A";

    A a(.*);

    initial begin
        int result;
        result = $coverage_control(`SV_COV_RESET, `SV_COV_TOGGLE, `SV_COV_MODULE, modName);
        result = $coverage_control(`SV_COV_RESET, `SV_COV_TOGGLE, `SV_COV_MODULE, top.x);
    end

endmodule
```

MikePopoloski:
The canBeStringLike function exists because many system functions have wording such as "shall be a variable of integral, unpacked array of byte, or string data types to which the resulting string shall be written". Basically an unpacked array of byte or any integral type can be implicitly passed to most system functions that expect a "string". Yes, that means goofy things like a single bit is allowed -- this would be a good place for a warning probably. 

Whether or not this should be allowed for the coverage functions is unclear -- the spec doesn't say. Usually the above mentioned behavior is for functions that existed in original Verilog, before there was a string type, so I think it's fine to err on the side of being more restrictive for the coverage functions which are definitely newer and if someone complains we can revisit.

source/compilation/builtins/CoverageFuncs.cpp
```diff
@@ -0,0 +1,107 @@
...
+        SystemSubroutine(name, SubroutineKind::Function),
+        argTypes(argTypes), returnType(&returnType),
+        nameOrHierIndex(nameOrHierIndex), requiredArgs(requiredArgs) {
+        ASSERT(requiredArgs <= argTypes.size());
+        ASSERT(nameOrHierIndex <= argTypes.size());
+    };
+
+    virtual const Expression& bindArgument(size_t argIndex, const BindContext& context,
+                                           const ExpressionSyntax& syntax, const Args& args) const final {
+        const BindContext* ctx = &context;
```
MikePopoloski, line 30: This variable isn't really doing anything is it?

source/compilation/builtins/CoverageFuncs.cpp
```diff
@@ -0,0 +1,107 @@
...
+        return Expression::bindArgument(*argTypes[argIndex], ArgumentDirection::In, syntax, *ctx);
+    }
+
+    const Type& checkArguments(const BindContext& context, const Args& args,
+                               SourceRange range, const Expression*) const {
+        auto& comp = context.getCompilation();
+        if (!checkArgCount(context, false, args, range, requiredArgs, argTypes.size()))
+            return comp.getErrorType();
+
+        auto& arg = args[nameOrHierIndex];
```
MikePopoloski, line 49: No need for a reference here, it's just a pointer.

source/compilation/builtins/CoverageFuncs.cpp
```diff
@@ -0,0 +1,107 @@
...
+            if (sym.isValue()) {
+                auto& type = sym.as<ValueSymbol>().getType();
+                if (!type.isString()) {
+                    context.addDiag(diag::BadSystemSubroutineArg, arg->sourceRange)
+                        << type << kindStr();
+                    return comp.getErrorType();
+                }
+            } else if (sym.kind != SymbolKind::Instance || !sym.as<InstanceSymbol>().isModule()) {
+                if (!context.scope->isUninstantiated())
+                    context.addDiag(diag::ExpectedModuleName, arg->sourceRange);
```
MikePopoloski, line 61: It's subtle, but I think the ExpectedModuleInstance is a better diagnostic here.

MikePopoloski (commented):
This is looking really good, just a couple more cleanup items and then we're good to go.

jrudess:
Thanks, I I suspected there was some spec language involving strings and integral types.

The only other check I have considered, but not implemented, is limiting the integer-arguments to the coverage built-in define values in $40.3.1 (either with an error or warning).  Any opinion on whether that's going too far on the checking?

source/compilation/builtins/CoverageFuncs.cpp
jrudess: Copy-paste :-)

jrudess:
One spot where the spec is perhaps a little vague is whether an interface-instance should really be excluded from being allowed for the hierarchical references.  I updated the unit-test to cover that case which currently throws an error.

I cannot remember from real-world experience what VCS or Questa will do with coverage of logic inside interfaces.  I have a suspicion that interfaces are probably allowed by the big-three, but I think I'd need to create a real-world test case to try it out to prove that.

MikePopoloski:
I wouldn't bother trying to validate the other arguments; they could always in theory be run-time determined and then it would be impossible to validate anyway. It doesn't hurt to add linter-style warnings for things like this, but there are plenty of other areas where time spent adding lint warnings would give you more bang for buck.

Agreed that interface instances are a gray area. I don't have a preference for whether they should be allowed or not, so you can decide and make the code changes accordingly.

jrudess:
Sounds good.  I'll leave everything as is for now then and look into allowing interfaces for some of the system tasks/functions once I've tested for consensus among the various simulator implementations.

PR #409, opened 2021-08-20
Implement PLA tasks

jrudess:
The implementation isn't complete, but creating the PR to facilitate some questions and review.

For the PLA tasks, $20.17.3 lists the following requirement:
> As shown in the examples in 20.17, PLA input terms, output terms, and memory shall be specified in ascending order.`

e.g. `logic [0:7] mem [0:7]`

I've implemented this check for the input and output args, but only partially for the mem arg.   I'm sure I've missed something obvious, but can you point me in the right direction for getting the dimensions of an array?

### DIFF
--- a/scripts/diagnostics.txt
+++ b/scripts/diagnostics.txt
@@ -703,6 +703,7 @@ error SampledValueLocalVar "local variables cannot be used in sampled value syst
 error SampledValueMatched "'matched' method cannot be used in sampled value system functions"
 error GlobalSampledValueAssertionExpr "global clocking future sampled value functions may be invoked only in sequence and property expressions"
 error SequenceMethodLocalVar "local variables can only be passed to sequences on which '{}' is called if they are the entire argument and not a sub-expression"
+error PlaRangeInAscendingOrder "Must specify dimensions in ascending order for PLA input type {}"
 warning format-real FormatRealInt "real value provided for integer format specifier '%{}'"
 warning finish-num BadFinishNum "finish argument must have value of 0, 1, or 2"
 warning missing-format MissingFormatSpecifier "missing format specifier"
--- a/scripts/diagnostics.txt
+++ b/scripts/diagnostics.txt
@@ -703,7 +703,7 @@ error SampledValueLocalVar "local variables cannot be used in sampled value syst
 error SampledValueMatched "'matched' method cannot be used in sampled value system functions"
 error GlobalSampledValueAssertionExpr "global clocking future sampled value functions may be invoked only in sequence and property expressions"
 error SequenceMethodLocalVar "local variables can only be passed to sequences on which '{}' is called if they are the entire argument and not a sub-expression"
-error PlaRangeInAscendingOrder "Must specify dimensions in ascending order for PLA input type {}"
+error PlaRangeInAscendingOrder "must specify dimensions in ascending order for PLA input type {}"
 warning format-real FormatRealInt "real value provided for integer format specifier '%{}'"
 warning finish-num BadFinishNum "finish argument must have value of 0, 1, or 2"
 warning missing-format MissingFormatSpecifier "missing format specifier"
--- a/source/compilation/builtins/SystemTasks.cpp
+++ b/source/compilation/builtins/SystemTasks.cpp
@@ -608,6 +608,58 @@ public:
     }
 };
 
+class PlaTask : public SystemTaskBase {
+public:
+    PlaTask(const std::string& name) :
+        SystemTaskBase(name), requiredArgs(3) {};
+
+    const Type& checkArguments(const BindContext& context, const Args& args, SourceRange range,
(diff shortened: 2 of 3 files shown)


MikePopoloski:
You can check for an unpacked type and pull the dimensions out manually, but there are helper methods on Type to do this for you:
hasFixedRange() returns true if the type has a declared static range, and getFixedRange will return it. This works for fixed-size unpacked arrays as well as any integral type (so you could use it for the packed portion as well and avoid the manual casting).

source/compilation/builtins/SystemTasks.cpp
```diff
@@ -608,6 +608,58 @@ class SdfAnnotateTask : public SystemTaskBase {
     }
 };
 
+class PlaTask : public SystemTaskBase {
+public:
+    PlaTask(const std::string& name) :
+        SystemTaskBase(name), requiredArgs(3) {};
```
MikePopoloski, line 614: No need for the member variable if this is always going to be constant. Can just put it right in the checkArgCount call or make it a local constant there.

scripts/diagnostics.txt
```diff
@@ -703,6 +703,7 @@ error SampledValueLocalVar "local variables cannot be used in sampled value syst
 error SampledValueMatched "'matched' method cannot be used in sampled value system functions"
 error GlobalSampledValueAssertionExpr "global clocking future sampled value functions may be invoked only in sequence and property expressions"
 error SequenceMethodLocalVar "local variables can only be passed to sequences on which '{}' is called if they are the entire argument and not a sub-expression"
+error PlaRangeInAscendingOrder "Must specify dimensions in ascending order for PLA input type {}"
```
MikePopoloski, line 706: nit: error messages should start lower case

source/compilation/builtins/SystemTasks.cpp
```diff
@@ -608,6 +608,58 @@ class SdfAnnotateTask : public SystemTaskBase {
...
+                }
+
+                // TODO: Check that array dimensions are in ascending order
+
+                auto& arrayType = *type.getArrayElementType();
+                if (arrayType.isSimpleBitVector()) {
+                    auto range = arrayType.getCanonicalType().as<IntegralType>().getBitVectorRange();
+                    if (range.left > range.right) {
+                        context.addDiag(diag::PlaRangeInAscendingOrder, args[i]->sourceRange) << arrayType;
+                        return context.getCompilation().getErrorType();
```
MikePopoloski, line 638: You have the compilation in local var `comp` so you can just use that here and below.

source/compilation/builtins/SystemTasks.cpp
```diff
@@ -730,6 +782,21 @@ void registerSystemTasks(Compilation& c) {
     TASK("$q_full", SubroutineKind::Function, 1, 1);
 
 #undef TASK
+
+#define PLA_TASK(name) c.addSystemSubroutine(std::make_unique<PlaTask>(name))
+    std::vector<std::string> arrayType = {"$async", "$sync"};
```
MikePopoloski, line 787: Not that it matters a whole lot, but since these are fixed size I'd use std::array

source/compilation/builtins/SystemTasks.cpp
```diff
@@ -608,6 +608,61 @@ class SdfAnnotateTask : public SystemTaskBase {
...
+            if (type.hasFixedRange() && !isValidRange(type)) {
+                return badRange(context, *args[i]);
+            }
+        }
+
+        return comp.getVoidType();
+    }
+
+protected:
+    bool isValidRange(const Type& type) const {
```
MikePopoloski, line 655: These helpers can be static.

source/compilation/builtins/SystemTasks.cpp
```diff
@@ -608,6 +608,61 @@ class SdfAnnotateTask : public SystemTaskBase {
...
+
+            if (type.hasFixedRange() && !isValidRange(type)) {
+                return badRange(context, *args[i]);
+            }
+        }
+
+        return comp.getVoidType();
+    }
+
+protected:
```
MikePopoloski, line 654: private instead of protected since there are no derived classes

source/compilation/builtins/SystemTasks.cpp
```diff
@@ -608,6 +608,61 @@ class SdfAnnotateTask : public SystemTaskBase {
     }
 };
 
+class PlaTask : public SystemTaskBase {
+public:
+    PlaTask(const std::string& name) :
+        SystemTaskBase(name) {};
+
+    const Type& checkArguments(const BindContext& context, const Args& args, SourceRange range,
+                               const Expression*) const override {
```
MikePopoloski, line 617: final since no derived classes

MikePopoloski (commented):
Looking good, just a few minor cleanups and then good to go

Issue #496, opened 2021-11-15
clang -Wshadow warning


mytbk:
```
/home/irudog/SourceCode/slang/source/compilation/builtins/MiscSystemFuncs.cpp:318:39: error: declaration shadows a structured binding [-Werror,-Wshadow]
                    [&visitor](auto&& arg) {
                                      ^
/home/irudog/SourceCode/slang/source/compilation/builtins/MiscSystemFuncs.cpp:316:33: note: previous declaration is here
            for (auto& [formal, arg] : aie.arguments) {
                                ^
1 error generated.
```

Kuree:
I can confirm this with `clang-13`.

MikePopoloski:
Fixed in 32458b8bba69d8928220ed9b3f0155018ee59d72

PR #499, opened 2021-11-19
Fix an error with clang-13

Kuree:
See https://github.com/MikePopoloski/slang/issues/496.


MikePopoloski:
Thanks for the PR! I'd prefer a more descriptive name instead of just adding an underscore so I just made the change myself since it's simple.

PR #502, opened 2021-11-23
Fix "use of bitwise '|' with boolean operands"

jesec:
Newer clang complains:

source/binding/Statements.cpp:2439:16: error: use of bitwise '|' with boolean operands [-Werror,-Wbitwise-instead-of-logical]
        bad |= expr.bad() | stmt.bad();
               ^~~~~~~~~~~~~~~~~~~~~~~
                          ||
source/binding/Statements.cpp:2439:16: note: cast one or both operands to int to silence this warning

Ref: https://wiki.sei.cmu.edu/confluence/display/c/EXP46-C.+Do+not+use+a+bitwise+operator+with+a+Boolean-like+operand


MikePopoloski:
Thanks for the PR!

PR #547, opened 2022-06-10
Remove specialization of std::hash for std::tuple

zachd-hrt:
The current specialization makes slang incompatible with other libraries that also include one, such as folly.

### DIFF
--- a/external/flat_hash_map.hpp
+++ b/external/flat_hash_map.hpp
@@ -1496,5 +1496,3 @@ struct power_of_two_std_hash : std::hash<T>
 };
 
 } // end namespace ska
-
-using ska::flat_hash_set;
--- a/external/flat_hash_map.hpp
+++ b/external/flat_hash_map.hpp
@@ -1497,5 +1497,4 @@ struct power_of_two_std_hash : std::hash<T>
 
 } // end namespace ska
 
-using ska::flat_hash_map;
 using ska::flat_hash_set;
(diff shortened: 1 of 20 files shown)


include/slang/util/FlatHashMap.h
```diff
@@ -0,0 +1,5 @@
+#include <flat_hash_map.hpp>
+
+#include "slang/util/Hash.h"
+template<typename K, typename V, typename H = slang::Hasher<K>, typename E = std::equal_to<K>,
```
MikePopoloski, line 4: You can just put this in Hash.h, no need for a separate file for just one definition. Also it should be namespaced anyway.

include/slang/util/Hash.h
```diff
@@ -49,16 +49,23 @@ struct HashValueImpl<Tuple, 0> {
 
 } // namespace slang
 
-namespace std {
+namespace slang {
+
+// Specialize a user-defined type instead of std::hash
+template <typename T> struct Hasher
```
MikePopoloski, line 55: Please run clang format on all of the changes.

external/flat_hash_map.hpp
```diff
@@ -1497,5 +1497,4 @@ struct power_of_two_std_hash : std::hash<T>
 
 } // end namespace ska
 
-using ska::flat_hash_map;
 using ska::flat_hash_set;
```
MikePopoloski, line 1500: Can you include the same change for flat_hash_set as well?

MikePopoloski (commented):
Thanks for the PR!

include/slang/util/FlatHashMap.h
MikePopoloski: Also I think this file is causing the build to fail due to lack of #pragma once

zachd-hrt:
Thanks for the comments. I've made the suggested changes.

PR #555, opened 2022-06-24
Remove stdc++fs flag since it's no longer needed for modern compilers

Kuree:


MikePopoloski:
I still ensure that slang builds with gcc-8 to make life easier for users on older systems. It would be nice to bump the minimum gcc version at some point but not sure when that will be. If this is causing problems for you, your PR will need to be more selective about when it includes the flag.

Kuree:
Thanks for the suggestion. I will add cmake check to add the flag if the version is below requirement.

The reason I'm filing this PR is that recent clang on Windows doesn't take `-lstdc++fs` flag anymore, see [here](https://github.com/Kuree/fsim/runs/7034008642?check_suite_focus=true#step:5:377).

PR #568, opened 2022-07-22
Fix parsing issue, colon followed by comment

HungMingWu:


tests/unittests/StatementTests.cpp
```diff
@@ -268,6 +268,25 @@ endmodule
     NO_COMPILATION_ERRORS;
 }
 
+TEST_CASE("Basic Case statements") {
```
MikePopoloski, line 271: This isn't really testing basic case statements, it's testing the lexing of comments next to the :/ operator. I think I'd just put a more targeted test into LexerTests instead so that it's clear what we're actually trying to catch here.

MikePopoloski (commented):
Thanks for the PR. This is indeed a weird case. I think your changes aren't sufficient though, since presumably the same problem arises with block comments (:/*) so we might as well handle them here too.

HungMingWu:
I have modified code as your suggestion. Hope it helps.

source/parsing/Lexer.cpp
```diff
@@ -375,6 +375,12 @@ Token Lexer::lexToken(KeywordVersion keywordVersion) {
                     advance();
                     return create(TokenKind::ColonEquals);
                 case '/':
+                    switch (peek(1)) {
+                    case '/':
+                        [[fallthrough]];
```
MikePopoloski, line 380: You don't need the [[fallthrough]] attribute if the case body is empty.

MikePopoloski (commented):
Thanks, this looks good. Can you update the one comment I made and also be sure to run clang-format on your changes and then I'll merge.

MikePopoloski:
Looks good, thanks!

PR #584, opened 2022-08-15
Implement simple observer mechanism

HungMingWu:
The usage is to separate plugin from driver itself.

### DIFF
--- a/include/slang/driver/Driver.h
+++ b/include/slang/driver/Driver.h
@@ -44,6 +44,9 @@ public:
     /// A list of syntax trees that have been parsed.
     std::vector<std::shared_ptr<SyntaxTree>> syntaxTrees;
 
+    /// A list of callback function on
+    std::vector<std::function<void(const RootSymbol&)>> onSuccessCallback;
+
     /// A container for various options that can be parsed and applied
     /// to the compilation process.
     struct Options {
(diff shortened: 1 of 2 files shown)


MikePopoloski:
Thanks for the PR. Can you say more about the intended usage? It's not obvious to me why this is useful, since the compilation calls return success or failure already, can't you simply check the return value and run any logic you want to run? Not sure what the callbacks are letting you do that you couldn't do already.

HungMingWu:
My original thought is append a hook before check errors and print warning/error
``` cpp
bool succeeded = diagEngine.getNumErrors() == 0;
// Insert here
if (!quiet) {
    OS::print("{} error{}, {} warning{}\n", diagEngine.getNumErrors(), )...;
}
```
Without these, if I have more than one separate programs for different purpose.
All I need to implement these code snippet.
``` cpp
auto compilation = driver.createCompilation();
ok &= driver.reportCompilation(*compilation, quiet == true);
if (ok) {
  // Do something for RootSymbol.
}
```
I think it is a kind of code redundant,, but maybe you think it is all right, that's ok.

MikePopoloski:
Yeah, I don't think this helps you save much code, if any. Checking whether there are errors is straightforward, just three lines of code:
```
for (auto& diag : compilation.getAllDiagnostics())
        diagEngine.issue(diag);

    bool succeeded = diagEngine.getNumErrors() == 0;
```
So you could run that yourself outside the Driver. Alternatively I could factor that out into its own function so that you can put your logic around it without needing a callback. Callbacks add complexity and obscure the control flow -- sometimes that's needed but in this case I don't think we do.

PR #587, opened 2022-08-17
update python bindings in response to overloaded setMappingsFromPragmas()

pistoletpierre:
Restore buildability of pyslang after setMappingsFromPragmas(BufferID id)


MikePopoloski:
Thanks!

PR #590, opened 2022-08-21
Add new diagnostic: Register only assigned on reset

Sustrak:
Added a new warning which triggers when a register is only assigned while the design is on reset.

### DIFF
--- a/source/compilation/ElabVisitors.h
+++ b/source/compilation/ElabVisitors.h
@@ -518,4 +518,72 @@ struct DefParamVisitor : public ASTVisitor<DefParamVisitor, false, false> {
     const InstanceSymbol* hierarchyProblem = nullptr;
 };
 
+// This visitor is for adding the only-assigned-on-reset diagnostic to the registers variables that
+// are only assigned on reset
+struct OnlyResetAssignment : public ASTVisitor<OnlyResetAssignment, true, false> {
+    struct AlwaysFFVisitor : public ASTVisitor<AlwaysFFVisitor, true, false> {
+        struct LookupLhsIdentifier : public ASTVisitor<LookupLhsIdentifier, true, true> {
+            explicit LookupLhsIdentifier(string_view name) : name(name) {}
+
+            void handle(const AssignmentExpression& expression) {
+                if (expression.left().kind == ExpressionKind::NamedValue) {
+                    auto namedValue = expression.left().as<NamedValueExpression>();
+                    if (namedValue.symbol.name == name)
+                        found = true;
+                }
+            }
+
+            string_view name;
+            bool found = false;
+        };
+
+        explicit AlwaysFFVisitor(string_view name) : name(name){};
+
+        void handle(const ConditionalStatement& statement) {
+            const auto& statementKind = statement.conditions.begin()->expr->kind;
+            if (statementKind == ExpressionKind::UnaryOp ||
+                statementKind == ExpressionKind::NamedValue) {
+                string_view identifier;
+                if (statementKind == ExpressionKind::UnaryOp)
+                    identifier = statement.conditions.begin()
+                                     ->expr->as<UnaryExpression>()
+                                     .operand()
+                                     .as<NamedValueExpression>()
+                                     .symbol.name;
+                else
+                    identifier =
+                        statement.conditions.begin()->expr->as<NamedValueExpression>().symbol.name;
+
+                const auto isReset = identifier.find("reset") != std::string_view::npos ||
+                                     identifier.find("rst") != std::string_view::npos;
+
+                if (isReset) {
+                    LookupLhsIdentifier visitor(name);
+                    statement.ifTrue.visit(visitor);
+                    if (visitor.found) {
+                        visitor.found = false;
+                        statement.ifFalse->visit(visitor);
+                        if (!visitor.found)
+                            error = true;
+                    }
+                }
+            }
+        }
+
+        string_view name;
+        bool error = false;
+    };
+
+    void handle(const VariableSymbol& symbol) const {
+        auto firstDriver = symbol.getFirstDriver();
+        if (firstDriver && firstDriver->isInAlwaysFFBlock()) {
+            AlwaysFFVisitor visitor(symbol.name);
+            firstDriver->containingSymbol->visit(visitor);
+            if (visitor.error) {
+                symbol.getParentScope()->addDiag(diag::OnlyAssignedOnReset, symbol.location)
+                    << symbol.name;
+            }
+        }
+    }
+};
 } // namespace slang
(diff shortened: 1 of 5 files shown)


jrudess:
I liked the idea of this diagnostic, so I gave it a spin on some projects.  It has a small issue with structs.

```systemverilog
struct.sv:10:9: warning: register 's' is only assigned on reset. This will cause warnings or errors in synthesis tools [-Wonly-assigned-on-reset]
    s_t s;
        ^
```
```systemverilog
module top;
    logic clk;
    logic rst_b;
    logic ns, ps;

    typedef struct packed {
        logic a;
        logic b;
    } s_t;
    s_t s;

    always_ff @(posedge clk, negedge rst_b) begin
        if (~rst_b)         s <= '0;
        else if (ns == 0) s.a <= 1'b1;
        else if (ps == 0) s.b <= 1'b1;
    end

endmodule
```

jrudess:
Similar variant for packed arrays:
```systemverilog
packed.sv:8:22: warning: register 'fifo' is only assigned on reset. This will cause warnings or errors in synthesis tools [-Wonly-assigned-on-reset]
    logic [4:0][2:0] fifo;
                     ^
```
```systemverilog
module top;
    logic clk;
    logic rst_b;
    logic push;
    logic ptr;
    logic [2:0] data;

    logic [4:0][2:0] fifo;

    always_ff @(posedge clk or negedge rst_b)
        if (~rst_b) fifo <= '0;
    else if (push) begin
        for (int i = 0; i <= 4; i++) begin
            if (ptr == i) fifo[i] <= data;
        end
    end
endmodule
```

source/compilation/ElabVisitors.h
```diff
@@ -518,4 +518,72 @@ struct DefParamVisitor : public ASTVisitor<DefParamVisitor, false, false> {
...
 
+// This visitor is for adding the only-assigned-on-reset diagnostic to the registers variables that
+// are only assigned on reset
+struct OnlyResetAssignment : public ASTVisitor<OnlyResetAssignment, true, false> {
+    struct AlwaysFFVisitor : public ASTVisitor<AlwaysFFVisitor, true, false> {
+        struct LookupLhsIdentifier : public ASTVisitor<LookupLhsIdentifier, true, true> {
+            explicit LookupLhsIdentifier(string_view name) : name(name) {}
+
+            void handle(const AssignmentExpression& expression) {
+                if (expression.left().kind == ExpressionKind::NamedValue) {
```
MikePopoloski, line 529: This check is why this isn't working for writes to struct or array members. You can use the Expression::getSymbolReference() method to get you the symbol referred to by an expression in a more generic way. For example:
```
if (auto symbol = expression.left().getSymbolReference(); symbol && symbol->name == name) { found = true; }
```

source/compilation/ElabVisitors.h
```diff
@@ -518,4 +518,72 @@ struct DefParamVisitor : public ASTVisitor<DefParamVisitor, false, false> {
...
+
+            string_view name;
+            bool found = false;
+        };
+
+        explicit AlwaysFFVisitor(string_view name) : name(name){};
+
+        void handle(const ConditionalStatement& statement) {
+            const auto& statementKind = statement.conditions.begin()->expr->kind;
+            if (statementKind == ExpressionKind::UnaryOp ||
```
MikePopoloski, line 544: Should this also be checking for binary expressions? like:
```sv
if (reset == 1'b0) begin end
```

source/compilation/ElabVisitors.h
```diff
@@ -518,4 +518,72 @@ struct DefParamVisitor : public ASTVisitor<DefParamVisitor, false, false> {
...
+                statementKind == ExpressionKind::NamedValue) {
+                string_view identifier;
+                if (statementKind == ExpressionKind::UnaryOp)
+                    identifier = statement.conditions.begin()
+                                     ->expr->as<UnaryExpression>()
+                                     .operand()
+                                     .as<NamedValueExpression>()
```
MikePopoloski, line 551: It's not guaranteed that the operand is a NamedValueExpression, so this will assert / crash if run on code that violates that.

source/compilation/ElabVisitors.h
```diff
@@ -518,4 +518,72 @@ struct DefParamVisitor : public ASTVisitor<DefParamVisitor, false, false> {
...
+                                     .symbol.name;
+                else
+                    identifier =
+                        statement.conditions.begin()->expr->as<NamedValueExpression>().symbol.name;
+
+                const auto isReset = identifier.find("reset") != std::string_view::npos ||
+                                     identifier.find("rst") != std::string_view::npos;
```
MikePopoloski, line 558: This will match variables like `first` which seems undesirable?

source/compilation/ElabVisitors.h
```diff
@@ -518,4 +518,72 @@ struct DefParamVisitor : public ASTVisitor<DefParamVisitor, false, false> {
...
+
+                if (isReset) {
+                    LookupLhsIdentifier visitor(name);
+                    statement.ifTrue.visit(visitor);
+                    if (visitor.found) {
+                        visitor.found = false;
+                        statement.ifFalse->visit(visitor);
```
MikePopoloski, line 565: Conditional statements don't need an ifFalse statement, so this will crash for code that just has an if with no matching else in it.

source/compilation/ElabVisitors.h
```diff
@@ -518,4 +518,72 @@ struct DefParamVisitor : public ASTVisitor<DefParamVisitor, false, false> {
...
+            }
+        }
+
+        string_view name;
+        bool error = false;
+    };
+
+    void handle(const VariableSymbol& symbol) const {
+        auto firstDriver = symbol.getFirstDriver();
+        if (firstDriver && firstDriver->isInAlwaysFFBlock()) {
```
MikePopoloski, line 579: Do you want to check the other drivers too? Some codebases use an initial block to assign to variables that otherwise only get modified in an always_ff, so you might miss those cases depending on which block comes first.

Sustrak:
Hi @jrudess @MikePopoloski thanks for the feedback. I'll check and correct the mistakes I've done in the implementation of my code, that you both have pointed out. Very appreciated for that.

Regarding 

> I think we want checks like this to live in a separate tool, like clang's "clang-tidy" tool.

Do you think this tool could live in this repo in a similar way the driver does? Maybe with the name slang-tidy?

Also 

> I bet it would still benefit from having an API in slang like clang's "AST Matchers" to make it even easier and more succinct to add checks like this.

I also think it would be very nice to have. I can try to generalize some parts of my code in order to start implementing some of these "AST Matchers"

One again, thanks for the feedback

salut!

MikePopoloski:
> Do you think this tool could live in this repo in a similar way the driver does? Maybe with the name slang-tidy?
Yes, that sound good to me. If you want to try that feel free.

> I also think it would be very nice to have. I can try to generalize some parts of my code in order to start implementing some of these "AST Matchers"
That would be good. One nice thing about the way clang does it is that you can register a bunch of matchers and then run the find operation once, so that adding more than one of these checks doesn't turn this into a big O(n^2) problem.

Sustrak:
Hi, I'm closing this MR since I'll try to better implement the solution with the ideas mentioned above.

PR #591, opened 2022-08-23
Enhance SyntaxVisitor functionality, support multiple operations

HungMingWu:


source/syntax/SyntaxVisitor.cpp
```diff
@@ -71,55 +77,54 @@ struct CloneVisitor {
...
+                        if (!listBuffer)
+                            backfillList(i);
+                        listBuffer->append(replaceChange->second);
+                    }
+                    else
+                        cloned->setChild(i, replaceChange->second);
+                }
+            }
+            else {
+                if (IsList) {
```
MikePopoloski, line 101: I think you want to keep the original logic here of:
```
if (IsList && listBuffer) 
    listBuffer->append(child->visit(*this));
else
    cloned->setChild(i, child->visit(*this));
```

If there are no remove or replace changes then you only want to add to the listBuffer if it's already been created, and otherwise just continue doing setChild() on the cloned object. This avoids the need to backfill the list in the common case where the rewriter didn't have any changes to the list at all.

tests/unittests/VisitorTests.cpp
```diff
@@ -96,6 +96,213 @@ endmodule
...
+    localparam int test_t__count2 = 3;
+    typedef enum int { FOO = 1, BAR = 2, BAZ = 3 } test_t;
+    localparam int test_t__count3 = 3;
+    localparam int test_t__count4 = 3;
+    function void foo(argA,int i, output r,argZ);
+    endfunction
+endmodule
+)");
+    }
+    SECTION("Combone insert and replace operation on oldNodes") {
```
MikePopoloski, line 173: typo: "Combine"

tests/unittests/VisitorTests.cpp
```diff
@@ -96,6 +96,213 @@ endmodule
...
+module M;
+    localparam int test_t__count1 = 3;
+    localparam int test_t__count2 = 3;
+    localparam int test_t__count3 = 3;
+    function void foo(argA,int i, output r,argZ);
+    endfunction
+endmodule
+)");
+    }
+    SECTION("Combone insert and remove operation on oldNodes") {
```
MikePopoloski, line 240: Same here

MikePopoloski (commented):
Thanks for the PR to fix this up, this looks really good. I left some comments on improvements.

source/syntax/SyntaxVisitor.cpp
HungMingWu: I have fixed all issues except this, I cannot pass unit test by your suggestion.
Maybe you can help on this modification.

MikePopoloski: Can you give some indication about what is failing with this change? I'm pretty sure what you have here isn't right because there's no point in having the backfillList() method at all if you're going to build the listBuffer on every child you see. An alternative implementation (which would be less efficient but simpler) would be to always have the listBuffer created and not need all the checks to see if it's been created already. Otherwise, we can figure out why my proposed change didn't work and change it to fix whatever the issue is.

HungMingWu: Here is my latest change
``` cpp
            if (auto it = commits.removeOrReplace.find(child);
                it != commits.removeOrReplace.end()) {
                if (auto replaceChange = std::get_if<detail::ReplaceChange>(&it->second);
                    replaceChange) {
                    if (IsList) {
                        if (!listBuffer)
                            backfillList(i);
                        listBuffer->append(replaceChange->second);
                    }
                    else
                        cloned->setChild(i, replaceChange->second);
                }
            }
            else {
                if (IsList && listBuffer)
                    listBuffer->append(child->visit(*this));
                else
                    cloned->setChild(i, child->visit(*this));
            }
```
And the error test is 
``` cpp
VisitorTests.cpp(96): FAILED:
  CHECK( SyntaxPrinter::printFile(*tree) == R"(
`define ENUM_MACRO(asdf) \
    typedef enum int {\
        FOO = 1,\
        BAR = 2,\
        BAZ = 3\
    } asdf;
module M;
    `ENUM_MACRO(test_t)
    localparam int test_t__count = 3;
endmodule
)" )
with expansion:
  "
  `define ENUM_MACRO(asdf) \
      typedef enum int {\
          FOO = 1,\
          BAR = 2,\
          BAZ = 3\
      } asdf;
  module M;
      localparam int test_t__count = 3;
  endmodule
  "
  ==
  "
  `define ENUM_MACRO(asdf) \
      typedef enum int {\
          FOO = 1,\
          BAR = 2,\
          BAZ = 3\
      } asdf;
  module M;
      `ENUM_MACRO(test_t)
      localparam int test_t__count = 3;
  endmodule
  "
```
It lacks ``ENUM_MACRO(test_t)` since no backfillList

MikePopoloski: I think that's because you're missing the `listBuffer->append(child->visit(*this));` lines that were in the original code for insertBefore and insertAfter. You still need to visit the original child in addition to adding the new nodes.

HungMingWu: Here is the code snippet as you desire, keep `listBuffer->append(child->visit(*this));` in insertBefore and insertAfter
``` cpp
            if (auto it = commits.insertBefore.find(child); it != commits.insertBefore.end()) {
                for (const auto& change : it->second) {
                    if (!listBuffer)
                        backfillList(i);
                    listBuffer->append(change.second);
                    listBuffer->append(child->visit(*this));
                }
            }

            if (auto it = commits.removeOrReplace.find(child);
                it != commits.removeOrReplace.end()) {
                if (auto replaceChange = std::get_if<detail::ReplaceChange>(&it->second);
                    replaceChange) {
                    if (IsList) {
                        if (!listBuffer)
                            backfillList(i);
                        listBuffer->append(replaceChange->second);
                    }
                    else
                        cloned->setChild(i, replaceChange->second);
                }
            }
            else {
                if (IsList && listBuffer)
                    listBuffer->append(child->visit(*this));
                else
                    cloned->setChild(i, child->visit(*this));
            }

            if (auto it = commits.insertAfter.find(child); it != commits.insertAfter.end()) {
                for (const auto& change : it->second) {
                    if (!listBuffer)
                        backfillList(i);
                    listBuffer->append(child->visit(*this));
                    listBuffer->append(change.second);
                }
            }
        }
```
But It fails in my appended unit test
``` cpp
class MultipleRewriter : public SyntaxRewriter<MultipleRewriter> {
            // Ignore
            void handle(const TypedefDeclarationSyntax& decl) {
                if (decl.type->kind != SyntaxKind::EnumType)
                    return;

                // Create a new localparam hardcoded with the number of entries in the enum.
                auto type = model.getDeclaredSymbol(decl.type->as<EnumTypeSyntax>());
                REQUIRE(type);

                size_t count = type->as<EnumType>().members().size();
                auto& newNode1 = parse(fmt::format("\n    localparam int {}__count1 = {};",
                                                   decl.name.valueText(), count));
                insertBefore(decl, newNode1);
                auto& newNode2 = parse(fmt::format("\n    localparam int {}__count2 = {};",
                                                   decl.name.valueText(), count));
                insertBefore(decl, newNode2);
                auto& newNode3 = parse(fmt::format("\n    localparam int {}__count3 = {};",
                                                   decl.name.valueText(), count));
                insertAfter(decl, newNode3);
                auto& newNode4 = parse(fmt::format("\n    localparam int {}__count4 = {};",
                                                   decl.name.valueText(), count));
                insertAfter(decl, newNode4);
            }
};
```
I insert four statements surrounding this node.
The result is not expected as my think since each insert operation would invoke `listBuffer->append(child->visit(*this));`
``` txt
VisitorTests.cpp(171): FAILED:
  CHECK( SyntaxPrinter::printFile(*tree) == R"(
module M;
    localparam int test_t__count1 = 3;
    localparam int test_t__count2 = 3;
    typedef enum int { FOO = 1, BAR = 2, BAZ = 3 } test_t;
    localparam int test_t__count3 = 3;
    localparam int test_t__count4 = 3;
    function void foo(argA,int i, output r,argZ);
    endfunction
endmodule
)" )
with expansion:
  "
  module M;
      localparam int test_t__count1 = 3;
      typedef enum int { FOO = 1, BAR = 2, BAZ = 3 } test_t;
      localparam int test_t__count2 = 3;
      typedef enum int { FOO = 1, BAR = 2, BAZ = 3 } test_t;
      typedef enum int { FOO = 1, BAR = 2, BAZ = 3 } test_t;
      typedef enum int { FOO = 1, BAR = 2, BAZ = 3 } test_t;
      localparam int test_t__count3 = 3;
      typedef enum int { FOO = 1, BAR = 2, BAZ = 3 } test_t;
      localparam int test_t__count4 = 3;
      function void foo(argA,int i, output r,argZ);
      endfunction
  endmodule
  "
  ==
  "
  module M;
      localparam int test_t__count1 = 3;
      localparam int test_t__count2 = 3;
      typedef enum int { FOO = 1, BAR = 2, BAZ = 3 } test_t;
      localparam int test_t__count3 = 3;
      localparam int test_t__count4 = 3;
      function void foo(argA,int i, output r,argZ);
      endfunction
  endmodule
  "
```

MikePopoloski: I see, with the additional kinds of modifications, backfilling becomes too difficult to do correctly. I think in that case you can revert back to the way you had it, but remove the backfillList() function and always unconditionally create the listBuffer is IsList is true. Then you can remove all the extra if (!listBuffer) checks and simplify the code.

HungMingWu: I am not sure whether it is your thought?
``` cpp
        auto backfillList = [&](size_t index) {
            if (cloned->kind != SyntaxKind::SyntaxList && cloned->kind != SyntaxKind::SeparatedList)
                throw std::logic_error("Can't use insertBefore or insertAfter on a non-list node");

            if (listBuffer)
                return;
            listBuffer.emplace();
            for (size_t i = 0; i < index; i++)
                listBuffer->append(cloned->getChild(i));
        };

         if (auto it = commits.insertBefore.find(child); it != commits.insertBefore.end()) {
                for (const auto& change : it->second) {
                    backfillList(i);
                    listBuffer->append(change.second);
                }
            }

            if (auto it = commits.removeOrReplace.find(child);
                it != commits.removeOrReplace.end()) {
                if (auto replaceChange = std::get_if<detail::ReplaceChange>(&it->second);
                    replaceChange) {
                    if (IsList) {
                        backfillList(i);
                        listBuffer->append(replaceChange->second);
                    }
                    else
                        cloned->setChild(i, replaceChange->second);
                }
            }
            else {
                if (IsList) {
                    backfillList(i);
                    listBuffer->append(child->visit(*this));
                }
                else
                    cloned->setChild(i, child->visit(*this));
            }

            if (auto it = commits.insertAfter.find(child); it != commits.insertAfter.end()) {
                for (const auto& change : it->second) {
                    backfillList(i);
                    listBuffer->append(change.second);
                }
            }
``

File: source/syntax/SyntaxVisitor.cpp [hunk 4]
```diff
@@ -54,6 +50,8 @@ struct CloneVisitor {
             if (cloned->kind != SyntaxKind::SyntaxList && cloned->kind != SyntaxKind::SeparatedList)
```
MikePopoloski, line 50: There's no reason to have this backfill function at all anymore. You call it on all paths through the loop, so index will always be zero on the first call and it will do nothing but emplace() on the listBuffer, and every subsequent call will early return. Just make listBuffer not optional and get rid of this lambda function.

PR #592, opened 2022-08-24
Fix object clone issue

HungMingWu:
I met the following issue
```  cpp
MemberSyntax* ptr = ...;
MemberSyntax* cloned = ptr->clone(alloc);
```
Due to no vtbl on class
Such as `DataDeclarationSyntax` or `TypedefDeclarationSyntax` cannot clone properly.


MikePopoloski:
Hmm, I definitely don't want to add a vtable to all of these syntax nodes, especially when I think there's a much simpler solution to adding clone() to these parent classes (which could be generated by the syntax_gen script by detecting that it's a non-final class):
```
template<typename T>
struct CloneVisitor {
    template<typename U>
    T* visit(const U& node, BumpAllocator& alloc) {
        return node.clone(alloc);
    }

    T* visitInvalid(const SyntaxNode&) { return nullptr; }
};

MemberSyntax* MemberSyntax::clone(BumpAllocator& alloc) {
    CloneVisitor<MemberSyntax> visitor;
    return visit(visitor, alloc);
}
```

HungMingWu:
I have direction how to implement it, but  I have another question about your sugesstion.
> detecting that it's a non-final class
Here is the syntax.txt
``` txt
TimingControl final=false
empty

Delay base=TimingControl multiKind=true
token hash
Expression delayValue
```
Here is the original generated header file

``` cpp
struct TimingControlSyntax : public SyntaxNode {
    // Ignore
};

struct DelaySyntax : public TimingControlSyntax {
    // Ignore
    DelaySyntax* clone(BumpAllocator& alloc) const;
};
```
Do you want to move clone function from `DelaySyntax ` to  `TimingControlSyntax ` ?

HungMingWu:
I found another issue according your proposal.
``` txt
WithClause
token with
token openParen
Expression expr
token closeParen
```
Since there is no final attribute, so it doesn't generate clone function.

So what's the direction
1. Mark the syntax with `final=false`
2. Append a new attribute, like `virtualClone` in the earlier commit.

MikePopoloski:
I'm not sure I follow your line of thought. My proposal was to always generate a clone() function in every syntax type. The body for final types, (which we already know today because that's the default if you don't set `final=false`) will continue to look just like it does today. For non-final types, which previously did not have a clone function, they will now get an implementation that uses the visitor as discussed above.

An alternative approach if you don't like that one would be to put the clone() method on the SyntaxNode base type and use a visitor to get down to the final type and call its clone directly. That's probably less generated code overall and so a bit easier on compile times.

PR #595, opened 2022-08-24
Use current source dir in cmake

zyedidia:
If slang is included as a sub-project within another project, using `CMAKE_SOURCE_DIR` will cause an error when trying to find `gitversion`. The build should use `CMAKE_CURRENT_SOURCE_DIR` instead.


MikePopoloski:
Thanks for the PR! At some point I plan to have an example project showing use via subdirectory with cmake so that will help flush out issues like this sooner.

PR #596, opened 2022-08-24
Parse legacy vendor worflows

udif:
This PR intends to make it easier to work with legacy workflows. It contains 3 unrelated changes (2 of which are unfortunately on the same commit):
1. Allow long command line option descriptions to be broken by newlines and neatly indented on the help screen.
2. Ignore Specman and VHDL files given on the command line.
3. Parse legacy EDA vendor file lists (relates to item 2 above).

I really hope my coding style is OK. Reading your codebase made me do lots of C++ catch-ups, but on the other hand, I restricted myself not to use any C++ 20 construct (`string_view::ends_with()` for example).

**As for the PR itself:**
A legacy codebase might contain for example thousands of files, split across dozens of file lists,
each one holding a separate block of 3rd party IP.

Unfortunately, each vendor tool has a unique command line interface, sometimes with thousands of possible flags.
`slang` cannot parse these file lists as-is.

Instead of taking a large effort , going over the file lists and converting those into equivalent `slang` file lists, this PR does the opposite - it allows us to define the vendor commands which we want to ignore and the vendor commands we want to rename into equivalent `slang` commands.

We define two new `slang` command line options:
```
  --cmd_ignore <vendor_cmd>,<N>            Define rule to ignore vendor command <vendor_cmd> with its following <N> parameters.
                                           A command of the form +xyz will also match any vendor command of the form +xyz+abc,
                                           as +abc is the command's argument, and doesn't need to be matched.
  --cmd_rename <vendor_cmd>,<slang_cmd>    Define rule to rename vendor command <vendor_cmd> into existing <slang_cmd>
```

Here is an example, combining all new features:
```
slang --cmd_ignore --xx,1 --cmd_ignore --ww,0 --cmd_ignore +abc,0 --cmd_rename --yy,--version --cmd_rename --zz,--help --xx 12121 --ww +abc+xyz x.vhd x.e x.v
```
First we define the `--xx` command to be ignored with it's 1 extra parameter.
Next we define the `+abc` command to be ignored too, which has no parameters.
Next we define the vendor command `--yy` to bre renamed into `slang`'s `--version` option.
Next we define the vendor command `--zz` to bre renamed into `slang`'s `--help` option.
Next we use the vendor `--xx 12121` option, which is ignored with its parameter.
Next we use the vendor `--ww` command which is also ignored (no parameter)
Finally, we ignore the vendor `+abc` command which has an embedded `+xyz` parameter.
Now we get 3 files, `x.vhd` , `x.e`,  (which we both ignore) and `x.v` which we read.

A typical usage would be to run the top level file list of your legacy code, see which commands `slang` is complaining about, and add them to your custom `vendor.f` file. once you are done, you can simply run:

```
slang -f vendor.f -f legacy_filelist.f
```

Using this technique, I was able to parse my legacy codebase without any unknown vendor command, but unfortunately `slang` has a few language coverage holes, and maybe a bug or two (one of the `` `define``'s is not detected, which is I know to be false, as other tools reads this design just fine).
I'll try to get to those bugs next week.

### DIFF
--- a/include/slang/util/CommandLine.h
+++ b/include/slang/util/CommandLine.h
@@ -226,8 +226,8 @@ public:
     /// @return true on success, false if an errors occurs.
     bool parse(int argc, const char* const argv[]);
 
-    std::string setIgnoreCommand(string_view value);
-    std::string setRenameCommand(string_view value);
+    std::string addIgnoreCommand(string_view value);
+    std::string addRenameCommand(string_view value);
 #if defined(_MSC_VER)
     /// Parse the provided command line (MSVC wchar-style).
     /// @return true on success, false if an errors occurs.
@@ -360,11 +360,6 @@ private:
     /// key is the vendor command name (including any leading +/- symbols)
     /// value is the slang command name to be used instead
     std::map<std::string, std::string> cmdRename;
-
-    // What we don't need:
-    // A map of commands to be renameded, with next argument merged into a plusArg format
-    // (because all slang's plusargs have non plusargs aliases, so no reason to create a
-    // plusArg) What we MIGHT need: Split a vendor's plusArg into regular arguments
 
     std::string programName;
     std::vector<std::string> errors;
--- a/include/slang/util/CommandLine.h
+++ b/include/slang/util/CommandLine.h
@@ -189,6 +189,32 @@ public:
     void add(string_view name, std::vector<std::string>& value, string_view desc,
              string_view valueName = {}, bool isFileName = false);
 
+    /// Register an option with @a name that will be parsed as a 2-tuple of string,int.
+    /// The tuple is stored in a map, with tuple[0] being the key, and tuple[1] the value.
+    /// The option can be invoked multiple times, to add multiple entries to the map.
+    ///
+    /// @name is a comma separated list of long form and short form names
+    /// (including the dashes) that are accepted for this option.
+    /// @a desc is a human-friendly description for printing help text.
+    /// @a valueName is an example name for the value when printing help text.
+    /// @a isFileName indicates whether the parsed string is a filename that
+    ///               might be relative to the current directory.
+    void add(string_view name, std::map<std::string, int>& value, string_view desc,
+             string_view valueName);
+
+    /// Register an option with @a name that will be parsed as a 2-tuple of string,string.
+    /// The tuple is stored in a map, with tuple[0] being the key, and tuple[1] the value.
+    /// The option can be invoked multiple times, to add multiple entries to the map.
+    ///
+    /// @name is a comma separated list of long form and short form names
+    /// (including the dashes) that are accepted for this option.
+    /// @a desc is a human-friendly description for printing help text.
+    /// @a valueName is an example name for the value when printing help text.
+    /// @a isFileName indicates whether the parsed string is a filename that
+    ///               might be relative to the current directory.
+    void add(string_view name, std::map<std::string, std::string>& value, string_view desc,
+             string_view valueName);
+
     using OptionCallback = std::function<std::string(string_view)>;
 
     /// Register an option with @a name that will be parsed as a string and
@@ -288,7 +314,7 @@ private:
                      optional<uint64_t>*, optional<double>*, optional<std::string>*,
                      std::vector<int32_t>*, std::vector<uint32_t>*, std::vector<int64_t>*,
                      std::vector<uint64_t>*, std::vector<double>*, std::vector<std::string>*,
-                     OptionCallback>;
+                     OptionCallback, std::map<std::string, int>*, std::map<std::string, std::string>* >;
 
     class Option {
     public:
@@ -317,6 +343,8 @@ private:
         std::string set(std::vector<double>& target, string_view name, string_view value);
         std::string set(std::vector<std::string>& target, string_view name, string_view value);
         std::string set(OptionCallback& target, string_view name, string_view value);
+        std::string set(std::map<std::string, int>& target, string_view name, string_view value);
+        std::string set(std::map<std::string, std::string>& target, string_view name, string_view value);
 
         template<typename T>
         static constexpr bool allowValue(const optional<T>& target) {
@@ -325,6 +353,11 @@ private:
 
         template<typename T>
         static constexpr bool allowValue(const std::vector<T>&) {
+            return true;
+        }
+
+        template<typename T1, typename T2>
+        static constexpr bool allowValue(const std::map<T1, T2>&) {
             return true;
         }
     };
@@ -339,6 +372,7 @@ private:
 
     void handlePlusArg(string_view arg, ParseOptions options, bool& hadUnknowns);
 
+    Option* findOption(string_view arg) const;
     Option* findOption(string_view arg, string_view& value) const;
     Option* tryGroupOrPrefix(string_view& arg, string_view& value, ParseOptions options);
     std::string findNearestMatch(string_view arg) const;
--- a/include/slang/util/CommandLine.h
+++ b/include/slang/util/CommandLine.h
@@ -189,28 +189,6 @@ public:
     void add(string_view name, std::vector<std::string>& value, string_view desc,
              string_view valueName = {}, bool isFileName = false);
 
-    /// Register an option with @a name that will be parsed as a 2-tuple of string,int.
-    /// The tuple is stored in a map, with tuple[0] being the key, and tuple[1] the value.
-    /// The option can be invoked multiple times, to add multiple entries to the map.
-    ///
-    /// @name is a comma separated list of long form and short form names
-    /// (including the dashes) that are accepted for this option.
-    /// @a desc is a human-friendly description for printing help text.
-    /// @a valueName is an example name for the value when printing help text.
-    void add(string_view name, std::map<std::string, int>& value, string_view desc,
-             string_view valueName);
-
-    /// Register an option with @a name that will be parsed as a 2-tuple of string,string.
-    /// The tuple is stored in a map, with tuple[0] being the key, and tuple[1] the value.
-    /// The option can be invoked multiple times, to add multiple entries to the map.
-    ///
-    /// @name is a comma separated list of long form and short form names
-    /// (including the dashes) that are accepted for this option.
-    /// @a desc is a human-friendly description for printing help text.
-    /// @a valueName is an example name for the value when printing help text.
-    void add(string_view name, std::map<std::string, std::string>& value, string_view desc,
-             string_view valueName);
-
     using OptionCallback = std::function<std::string(string_view)>;
 
     /// Register an option with @a name that will be parsed as a string and
@@ -248,6 +226,8 @@ public:
     /// @return true on success, false if an errors occurs.
     bool parse(int argc, const char* const argv[]);
 
+    std::string setIgnoreCommand(string_view value);
+    std::string setRenameCommand(string_view value);
 #if defined(_MSC_VER)
     /// Parse the provided command line (MSVC wchar-style).
     /// @return true on success, false if an errors occurs.
@@ -310,8 +290,7 @@ private:
                      optional<uint64_t>*, optional<double>*, optional<std::string>*,
                      std::vector<int32_t>*, std::vector<uint32_t>*, std::vector<int64_t>*,
                      std::vector<uint64_t>*, std::vector<double>*, std::vector<std::string>*,
-                     OptionCallback, std::map<std::string, int>*,
-                     std::map<std::string, std::string>*>;
+                     OptionCallback>;
 
     class Option {
     public:
@@ -340,9 +319,6 @@ private:
         std::string set(std::vector<double>& target, string_view name, string_view value);
         std::string set(std::vector<std::string>& target, string_view name, string_view value);
         std::string set(OptionCallback& target, string_view name, string_view value);
-        std::string set(std::map<std::string, int>& target, string_view name, string_view value);
-        std::string set(std::map<std::string, std::string>& target, string_view name,
-                        string_view value);
 
         template<typename T>
         static constexpr bool allowValue(const optional<T>& target) {
@@ -351,11 +327,6 @@ private:
 
         template<typename T>
         static constexpr bool allowValue(const std::vector<T>&) {
-            return true;
-        }
-
-        template<typename T1, typename T2>
-        static constexpr bool allowValue(const std::map<T1, T2>&) {
             return true;
         }
     };
@@ -370,7 +341,6 @@ private:
 
     void handlePlusArg(string_view arg, ParseOptions options, bool& hadUnknowns);
 
-    Option* findOption(string_view arg) const;
     Option* findOption(string_view arg, string_view& value) const;
     Option* tryGroupOrPrefix(string_view& arg, string_view& value, ParseOptions options);
     std::string findNearestMatch(string_view arg) const;
@@ -378,6 +348,23 @@ private:
     std::shared_ptr<Option> positional;
     std::map<std::string, std::shared_ptr<Option>> optionMap;
     std::vector<std::shared_ptr<Option>> orderedOptions;
+
+    /// A map of commands to be ignored.
+    /// key is the command name (including any leading +/- symbols)
+    /// value is an the number of arguments to be skipped (int)
+    /// If argument begins with a '+' then matching will ignore anything after a 2nd '+'
+    /// so that +vendorXYZ+vendorARG can be ignored by matching against +vendorXYZ
+    std::map<std::string, int> cmdIgnore;
+
+    /// A map of commands to be renamed, pointing to new name
+    /// key is the vendor command name (including any leading +/- symbols)
+    /// value is the slang command name to be used instead
+    std::map<std::string, std::string> cmdRename;
+
+    // What we don't need:
+    // A map of commands to be renameded, with next argument merged into a plusArg format
+    // (because all slang's plusargs have non plusargs aliases, so no reason to create a
+    // plusArg) What we MIGHT need: Split a vendor's plusArg into regular arguments
 
     std::string programName;
     std::vector<std::string> errors;
--- a/include/slang/util/CommandLine.h
+++ b/include/slang/util/CommandLine.h
@@ -310,7 +310,8 @@ private:
                      optional<uint64_t>*, optional<double>*, optional<std::string>*,
                      std::vector<int32_t>*, std::vector<uint32_t>*, std::vector<int64_t>*,
                      std::vector<uint64_t>*, std::vector<double>*, std::vector<std::string>*,
-                     OptionCallback, std::map<std::string, int>*, std::map<std::string, std::string>* >;
+                     OptionCallback, std::map<std::string, int>*,
+                     std::map<std::string, std::string>*>;
 
     class Option {
     public:
@@ -340,7 +341,8 @@ private:
         std::string set(std::vector<std::string>& target, string_view name, string_view value);
         std::string set(OptionCallback& target, string_view name, string_view value);
         std::string set(std::map<std::string, int>& target, string_view name, string_view value);
-        std::string set(std::map<std::string, std::string>& target, string_view name, string_view value);
+        std::string set(std::map<std::string, std::string>& target, string_view name,
+                        string_view value);
 
         template<typename T>
         static constexpr bool allowValue(const optional<T>& target) {
--- a/include/slang/util/CommandLine.h
+++ b/include/slang/util/CommandLine.h
@@ -197,8 +197,6 @@ public:
     /// (including the dashes) that are accepted for this option.
     /// @a desc is a human-friendly description for printing help text.
     /// @a valueName is an example name for the value when printing help text.
-    /// @a isFileName indicates whether the parsed string is a filename that
-    ///               might be relative to the current directory.
     void add(string_view name, std::map<std::string, int>& value, string_view desc,
              string_view valueName);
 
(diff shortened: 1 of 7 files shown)


MikePopoloski:
Hello, thank you for the PR. I think this is a worthy goal and your end state looks fairly nice -- I have some quibbles with the implementation that I'll share in a more detailed review but wanted to start out by saying that the broad gist makes sense to me.

Some side notes:
- Yeah, the project is currently limited to C++17. I'm hoping to make GCC 9 my lowest supported version very soon, and then probably make that GCC 10 early next year, at which point I can start requiring C++20 support.
- The missing `define might be because your project requires macros to bleed through across files, the so-called "single compilation unit" mode. slang supports that if you pass `--single-unit`. I think it's a very lamentable part of the Verilog ecosystem but everyone does it so it is what it is I guess.
- Definitely interested if you have any bugs to file, or if you want to upvote existing issues for missing language features that you'd like prioritized. No promises of course, but it's helpful to me when choosing what to work on next.

include/slang/util/CommandLine.h
```diff
@@ -189,6 +189,32 @@ class CommandLine {
...
+
+    /// Register an option with @a name that will be parsed as a 2-tuple of string,string.
+    /// The tuple is stored in a map, with tuple[0] being the key, and tuple[1] the value.
+    /// The option can be invoked multiple times, to add multiple entries to the map.
+    ///
+    /// @name is a comma separated list of long form and short form names
+    /// (including the dashes) that are accepted for this option.
+    /// @a desc is a human-friendly description for printing help text.
+    /// @a valueName is an example name for the value when printing help text.
+    /// @a isFileName indicates whether the parsed string is a filename that
```
MikePopoloski, line 213: Same here.

source/driver/Driver.cpp
```diff
@@ -143,6 +154,10 @@ void Driver::addStandardArgs() {
 
     cmdLine.setPositional(
         [this](string_view fileName) {
+            if (fileName.rfind(".e"sv) == fileName.size() - 2)
```
MikePopoloski, line 157: I get what you're trying to do here but it's a little too ad-hoc and project-specific for my liking. Perhaps we can back this part of the change out for now and in a separate diff we can add a configureable filter to exclude files, maybe based on a glob or regex pattern or something.

source/util/CommandLine.cpp
```diff
@@ -397,6 +413,34 @@ bool CommandLine::parse(span<const string_view> args, ParseOptions options) {
             continue;
         }
 
+        // check if arg is in the list of commands to skip
+        auto cmd_ignore_option = findOption("cmd_ignore");
```
MikePopoloski, line 417: The CommandLine class should remain project-independent, meaning we could always extract it out and use it somewhere else without needing to change much, if anything. I think this gets a little too close to hardcoding slang-specific details inside this class. I think instead you could make this feature explicit in the interface of the class instead of implicit based on a magic name -- instead of a general add() for a new option, make a function like setIgnoreCommand() and setRenameCommand() and store those as separate members instead of the general list of options. Then you can avoid all this extra machinery and it will be a tiny bit more efficient to boot.

source/util/CommandLine.cpp
```diff
@@ -397,6 +413,34 @@ bool CommandLine::parse(span<const string_view> args, ParseOptions options) {
...
+        string_view ignore_arg = arg;
+        // if we ignore a vendor command of the form +xx ,
+        // we match on any +xx+yyy command as +yy is the command's argument
+        if (arg[0] == '+') {
+            size_t plusIndex = arg.substr(1).find_first_of('+');
+            if (plusIndex != string_view::npos)
+                ignore_arg =
+                    arg.substr(0, plusIndex + 1); // +1 because we started from arg.substr(1)
+        }
+        if (auto it{ cmdIgnore->find(std::string(ignore_arg)) }; it != cmdIgnore->end()) {
```
MikePopoloski, line 429: I would generally just use the normal `=` initialization for simple variables; the braces aren't buying you anything.

source/util/CommandLine.cpp
```diff
@@ -492,11 +536,24 @@ std::string CommandLine::getHelpText(string_view overview) const {
...
         result += fmt::format("  {:{}}"sv, key, maxLen);
-        if (!opt->desc.empty())
-            result += fmt::format("  {}", opt->desc);
-        result += "\n";
+        if (!opt->desc.empty()) {
+            bool first = true;
+            std::istringstream iss(opt->desc);
+            std::string item;
+            while (std::getline(iss, item, '\n')) {
+                if (!first)
```
MikePopoloski, line 547: I think this drops the spacing between options and their descriptions? I think if you run the unit tests they should catch this. It would be good to update them for the new features too.

source/util/CommandLine.cpp
```diff
@@ -801,4 +865,32 @@ std::string CommandLine::Option::set(OptionCallback& target, string_view, string
     return target(value);
 }
 
+std::string CommandLine::Option::set(std::map<std::string, int>& target, string_view name,
+                                     string_view value) {
+    size_t equalsIndex = value.find_first_of(',');
+    string_view num_args;
+    if (equalsIndex != string_view::npos) {
+        num_args = value.substr(equalsIndex + 1);
```
MikePopoloski, line 873: Noticed this in a few places in this diff, but please be consistent about naming style. Local variables are camelCase, no underscores.

source/util/CommandLine.cpp
```diff
@@ -801,4 +865,32 @@ std::string CommandLine::Option::set(OptionCallback& target, string_view, string
...
+        value = value.substr(0, equalsIndex);
+        std::string error;
+        auto result = parseInt<uint64_t>(name, num_args, error);
+        if (result)
+            target[std::string(value)] = *result;
+        return error;
+    }
+    return {};
```
MikePopoloski, line 881: If there is no comma should that be considered an error? Seems like it to me, because the option on its own is otherwise useless.

MikePopoloski (commented):
Also note that there are some CI build failures that should be addressed.

udif:
Thanks for your comments!
I will try to fix all your remarks.
I think the CI is failing because of a missing `#include <sstream>` . For some reason it worked fine on Linux/gcc, so I assumed the file was already being included by some other include file. I'll add it explicitly.

The problem is that I don't have a good methodology to run these integration tests before submitting a PR. They don't run on my fork because the PR lives in a dedicated branch, and the github action runs only on pushes to the master branch. Modifying the github actions yaml file to change this will introduce an extra unrelated file into the PR.
I even tried the[ `netkos/act`](https://github.com/nektos/act) tool, but for some reason I can't get it to work.

One of my failures could definitely be due to missing `--single-unit`. I know for a fact we rely on this.
OI just reran `slang` with this flag, and that error is now gone!
What I do have is thousands of warnings due to lines inside `specify` blocks:
```
if( XXX==1'b1 )( YYY => ZZZ ) = ( `DDD1,`DDD2 ); 
```
`XXX`, `YYY`, `ZZZ` are I/O ports in a library cell, and `` `DDD1 `` and `` `DDD2 `` are timing constants. These are not the real names since it's part of a commercial library.

After the `--single-unit` fix, the run now fails due to a new error: `error: exceeded max include depth` on some of my include file.
I have no idea what might have caused this, I'll need to add some debugging prints for that.

I'll start cleaning up according to your comments.

include/slang/util/CommandLine.h
```diff
@@ -189,6 +189,32 @@ class CommandLine {
...
 
+    /// Register an option with @a name that will be parsed as a 2-tuple of string,int.
+    /// The tuple is stored in a map, with tuple[0] being the key, and tuple[1] the value.
+    /// The option can be invoked multiple times, to add multiple entries to the map.
+    ///
+    /// @name is a comma separated list of long form and short form names
+    /// (including the dashes) that are accepted for this option.
+    /// @a desc is a human-friendly description for printing help text.
+    /// @a valueName is an example name for the value when printing help text.
+    /// @a isFileName indicates whether the parsed string is a filename that
```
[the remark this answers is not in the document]
udif: Done (to be pushed soon).

include/slang/util/CommandLine.h
udif: Done (to be pushed soon).

MikePopoloski:
You can suppress the warnings about the specify blocks, something like `-Wno-not-supported` to clean that up, or just comment out that line in the source.

The max include file depth is set to some arbitrary value (I think like 128), which is perhaps too low if you have a very nested set of includes. You can bump that limit with a command line arg. If slang is actually in an include loop though, maybe due to some macro trickery that isn't working correctly in slang to stop the recursive include, that won't help.

No worries about the CI failures, you can just bounce your changes off the PR here and let Github actions run them for you.

source/util/CommandLine.cpp
udif: OK, I understand, you suggest to make command renaming and ignoring an integral part of the CommandLine class.

MikePopoloski: Yeah that's right, you basically already have done that here, so there's no reason not to make life easier on the implementation.

MikePopoloski: Also just to say it, I think it's still an optional thing. We don't need to hardcode the name, the user of the class can enable it and tell you what name that command should have by doing something like enableIgnoreCommand("cmd_ignore"), enableReplaceCommand("cmd_replace")

source/util/CommandLine.cpp
udif: I have to admit I took this construct straight out of stackoverflow :-) There are just too many nuances with the different C++ initializer types (direct vs copy vs value), and I didn't want to introduce any subtle bug.
I could also break it into separate lines for assigning the iterator and for testing its value, but the current form keeps the iterator scope inside the `if` block, and I could reuse the short temporary `it` variable.

MikePopoloski: No, using the if statement initialization syntax is totally fine. Mostly I'm just saying to use the straightforward equals sign unless there's a good reason to do otherwise; yes C++ has made a hash of initialization syntax but for the most part the straightforward thing will do what you want so don't worry too much about the braced initialization stuff.

source/util/CommandLine.cpp
udif: It doesn't (see quote in the original PR description, it's copied directly from the help screen).
Here is a larger snippet from the `--help` output:
```
  --libraries-inherit-macros               If true, library files will inherit macro definitions from the primary source files. --single-unit must also be passed when this option is used.
  --cmd_ignore <vendor_cmd>,<N>            Define rule to ignore vendor command <vendor_cmd> with its following <N> parameters.
                                           A command of the form +xyz will also match any vendor command of the form +xyz+abc,
                                           as +abc is the command's argument, and doesn't need to be matched.
  --cmd_rename <vendor_cmd>,<slang_cmd>    Define rule to rename vendor command <vendor_cmd> into existing <slang_cmd>
  --max-parse-depth <depth>                Maximum depth of nested language constructs allowed
  --max-lexer-errors <count>               Maximum number of errors that can occur during lexing before the rest of the file is skipped
  --max-hierarchy-depth <depth>            Maximum depth of the design hierarchy
  --max-generate-steps <steps>             Maximum number of steps that can occur during generate block evaluation before giving up
  --max-constexpr-depth <depth>            Maximum depth of a constant evaluation call stack
  --max-constexpr-steps <steps>            Maximum number of steps that can occur during constant evaluation before giving up
  --constexpr-backtrace-limit <limit>      Maximum number of frames to show when printing a constant evaluation backtrace; the rest will be abbreviated
  --max-instance-array <limit>             Maximum number of instances allowed in a single instance array
 ```

Unless you meant something else, which I missed.

This should have been split to a standalone PR, but by the time I realized I needed it, I was already knee deep in the main PR, so I left it as part of the main PR.

File: source/util/CommandLine.cpp (hunk 6 above)
Comment (udif) [reply source/util/CommandLine.cpp:873]: I'll fix variable naming.

File: source/util/CommandLine.cpp (hunk 7 above)
Comment (udif) [reply source/util/CommandLine.cpp:881]: It is considered an error, but it is handled silently by not accepting the option.
I should have added an explicit error message.

File: source/util/CommandLine.cpp (hunk 5 above)
Comment (MikePopoloski) [reply source/util/CommandLine.cpp:547]: Oh hmm, previously the code would do += format("  {}", desc), so there was two spaces added. Now the desc doesn't get those two spaces added before it (unless I'm missing something) so I would expect that the very longest option name would no longer have the space. All of the shorter names of course still benefit from the maxLen indentation.

**Comment by udif** [comment]
I'm pretty sure the `` `include`` error is not due to a real limit of 128 nested includes. It must be kind of infinite recursive `` `include`` loop, which I'll have to debug. It didn't happen elsewhere (our vendor tools, as well as `Surelog`, which also acquired a similar but not identical set of "vendor" commands by me last week).

File: source/driver/Driver.cpp (hunk 2 above)
Comment (udif) [reply source/driver/Driver.cpp:157]: I would like to do this now, because the design I'm test-parsing includes vendor VHDL and DV files in Specman, and without filtering those, I won't be able to complete it.
What about something like `--exclude-ext e,vhd` (or maybe `--noext` , to be more in-line with `--libext`)  and use the `std::vector<std::string>&` option type that already exists to store all the extensions to be skipped.

Comment (MikePopoloski) [reply source/driver/Driver.cpp:157]: Yeah, I like --exclude-ext

File: source/util/CommandLine.cpp (hunk 5 above)
Comment (udif) [reply source/util/CommandLine.cpp:547]: Yup, you're right. I see what you mean and upon close examination, the longest one is indeed touching the description. I'll just change the alignment width factor to `maxLen + 1` or `maxLen + 2`.

**Comment by udif** [comment]
At this point, I think I implemented everything above, except the following taskss:

1. implement `--exclude-ext` instead of hardwired checks.
2. remove CommandLine class dependency on Driver commands `--cmd_ignore` and `--cmd_rename` by moving their std::map tables into CommandLine as described above.
3. Implement unit tests to cover the new features of this PR in `slang/tests/unittests/UtilTests.cpp`

I plan to complete these in the next few days.

**Comment by udif** [comment]
Just pushed a fix for `--exclude-ext` implementation.
`bin/unittests` is passing, but I haven't added specific unit tests yet (will do later for the whole PR).

I'm running the command below and getting expected output.
```
localhost:~/git/slang/build$ bin/slang --exclude-ext vhd --exclude-ext e --exclude-ext "" --cmd_ignore --xx,1 --cmd_ignore --ww,0 --cmd_ignore +abc,0 --cmd_rename --yy,--version --cmd_rename --zz,--help --xx 12121 --ww +abc+xyz a.vhd a.e xxx a.v
error: no such file or directory: 'a.v'
```
For some reason I was expecting `--exclude-ext vhd,e` to work, but apparently this style is supported only for plusargs (`+flag+arg1+arg2+...` ?).

**Comment by MikePopoloski** [comment]
For list / vector options using long form --, you can pass the command multiple times and the provided values are added to the list. It makes sense to add support for the comma form as well since it looks natural, perhaps on an opt-in basis for string options since you may have a case where you want to pass a string actually containing commas and not have it imply a list.

Also it occurred to me that maybe we don't actually want the cmd_ignore functionality, if instead you had an option (or maybe just the default) to ignore all unknown commands in .f command files?

**Comment by udif** [comment]
> Also it occurred to me that maybe we don't actually want the cmd_ignore functionality, if instead you had an option (or maybe just the default) to ignore all unknown commands in .f command files?

I see two reasons against it:
1. The command also specifies how many arguments to ignore. Some vendor commands have no arguments, but others may have one or more arguments.  `--cmd-ignore` specifies the number of arguments to skip for the specified command.
2. If you silently ignore unknown commands, you might miss typos in commands you do want to execute.

**Comment by MikePopoloski** [comment]
Yeah, those are valid reasons. Carry on then.

**Comment by udif** [comment]
I think I've fixed everything mentioned above.
As I wrote, I added unit tests for all new features, but for the help message feature (adding `\n` to break a long description) I simply modified an existing test.

**Comment by udif** [comment]
OK, everything passes!

File: source/driver/Driver.cpp [hunk 9]
```diff
@@ -143,6 +158,18 @@ void Driver::addStandardArgs() {
 
     cmdLine.setPositional(
         [this](string_view fileName) {
+            if (!options.excludeExtDone) {
```
MikePopoloski, line 161: This boolean flag is kind of clunky, and furthermore relies on additional --exclude-ext args never coming after a positional argument, which nothing will currently prevent. I don't think you need it though; instead of storing the exclude-ext args in a vector, make that option take a callback that adds the values directly to the set instead and then you won't need a separate data structure.

include/slang/util/CommandLine.h
```diff
@@ -226,6 +226,8 @@ class CommandLine {
     /// @return true on success, false if an errors occurs.
     bool parse(int argc, const char* const argv[]);
 
+    std::string setIgnoreCommand(string_view value);
```
MikePopoloski, line 229: When I recommended the set* names for these I was thinking of a slightly different implementation; now that I see what you have here I would recommend these be add* instead because you can call them multiple times and their effects add to each other.

include/slang/util/CommandLine.h
```diff
@@ -347,6 +349,23 @@ class CommandLine {
...
+    /// If argument begins with a '+' then matching will ignore anything after a 2nd '+'
+    /// so that +vendorXYZ+vendorARG can be ignored by matching against +vendorXYZ
+    std::map<std::string, int> cmdIgnore;
+
+    /// A map of commands to be renamed, pointing to new name
+    /// key is the vendor command name (including any leading +/- symbols)
+    /// value is the slang command name to be used instead
+    std::map<std::string, std::string> cmdRename;
+
+    // What we don't need:
```
MikePopoloski, line 364: You can remove this comment here, I don't think it's adding much at this point.

MikePopoloski (commented):
Thanks, this looks really good now. Just a couple comments left and then we can merge this.

source/driver/Driver.cpp
udif: I didn't realize you could mix positional arguments and flags, because that never happens on standard Linux tools, but after you mentioned this, I realized you are right. Each '-f' file usually contains both filenames and incdir/defines/libdir flags, so the end result is that they are really mixed.

I'll modify the implementation to add members directly to the set, similar to the std::map code I had before.

udif:
I'm having difficulties using `flat_hash_set` in `CommandLine.h` .

Trying : `#include "slang/util/Hash.h"` is bringing up hundreds of lines of template errors, that all originate just from including that file:
```
[  1%] Building CXX object source/CMakeFiles/slangcore.dir/util/CommandLine.cpp.o
In file included from /usr/include/c++/9/bits/move.h:55,
                 from /usr/include/c++/9/bits/stl_function.h:60,
                 from /usr/include/c++/9/functional:49,
                 from /home/udif/git/slang/source/../include/slang/util/CommandLine.h:9,
                 from /home/udif/git/slang/source/util/CommandLine.cpp:7:
/usr/include/c++/9/type_traits: In substitution of ...
...
```
and eventually boild to:
```
...
/usr/include/c++/9/type_traits:2330:26: error: request for member ‘operator()’ is ambiguous
 2330 |       std::declval<_Fn>()(std::declval<_Args>()...)
      |       ~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /usr/include/c++/9/functional:49,
                 from /home/udif/git/slang/source/../include/slang/util/CommandLine.h:9,
                 from /home/udif/git/slang/source/util/CommandLine.cpp:7:
/usr/include/c++/9/bits/stl_function.h:355:7: note: candidates are: ‘constexpr bool std::equal_to<_Tp>::operator()(const _Tp&, const _Tp&) const [with _Tp = std::__cxx11::basic_string<char>]’
  355 |       operator()(const _Tp& __x, const _Tp& __y) const
      |       ^~~~~~~~
In file included from /home/udif/git/slang/source/../include/slang/util/Hash.h:9,
                 from /home/udif/git/slang/source/../include/slang/util/CommandLine.h:15,
                 from /home/udif/git/slang/source/util/CommandLine.cpp:7:
/home/udif/git/slang/external/flat_hash_map.hpp:45:12: note:                 ‘template<class ... Args> Result ska::detailv3::functor_storage<Result, Functor>::operator()(Args&& ...) const [with Args = {Args ...}; Result = bool; Functor = std::equal_to<std::__cxx11::basic_string<char> >]’
   45 |     Result operator()(Args &&... args) const
      |            ^~~~~~~~
/home/udif/git/slang/external/flat_hash_map.hpp:40:12: note:                 ‘template<class ... Args> Result ska::detailv3::functor_storage<Result, Functor>::operator()(Args&& ...) [with Args = {Args ...}; Result = bool; Functor = std::equal_to<std::__cxx11::basic_string<char> >]’
   40 |     Result operator()(Args &&... args)
      |            ^~~~~~~~
/home/udif/git/slang/external/flat_hash_map.hpp:45:12: note:                 ‘template<class ... Args> Result ska::detailv3::functor_storage<Result, Functor>::operator()(Args&& ...) const [with Args = {Args ...}; Result = long unsigned int; Functor = slang::Hasher<std::__cxx11::basic_string<char> >]’
   45 |     Result operator()(Args &&... args) const
      |            ^~~~~~~~
/home/udif/git/slang/external/flat_hash_map.hpp:40:12: note:                 ‘template<class ... Args> Result ska::detailv3::functor_storage<Result, Functor>::operator()(Args&& ...) [with Args = {Args ...}; Result = long unsigned int; Functor = slang::Hasher<std::__cxx11::basic_string<char> >]’
   40 |     Result operator()(Args &&... args)
      |            ^~~~~~~~
make[2]: *** [source/CMakeFiles/slangcore.dir/build.make:266: source/CMakeFiles/slangcore.dir/util/CommandLine.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:1123: source/CMakeFiles/slangcore.dir/all] Error 2
make: *** [Makefile:141: all] Error 2
```
I'm commenting in and out my changes. but so far I'm not advancing.

include/slang/util/CommandLine.h
udif: ok, I'll take care of that.

udif:
ok, I hope everything is OK now. CI is passing.

source/driver/Driver.cpp
udif: as you said, the callback method was the simplest.

MikePopoloski:
Looks good, thanks!

Issue #600, opened 2022-08-30
lexerStack() management bug?


udif:
I've been debugging why I'm getting `exceeded max include depth` on my large design, and I think I found the reason:
In `Driver.cpp`:
```
bool Driver::runPreprocessor(bool includeComments, bool includeDirectives) {
    BumpAllocator alloc;
    Diagnostics diagnostics;
    Preprocessor preprocessor(sourceManager, alloc, diagnostics, createOptionBag());

    for (auto it = buffers.rbegin(); it != buffers.rend(); it++)
        preprocessor.pushSource(*it);
```

But `pushSource` is doing this:
```
void Preprocessor::pushSource(SourceBuffer buffer) {
    ASSERT(buffer.id);

    std::cout << "pushSource:  lexerStack.size(): " << lexerStack.size() << std::endl;
    lexerStack.emplace_back(std::make_unique<Lexer>(buffer, alloc, diagnostics, lexerOptions));
}
```

It is creating a lexer stack entry, even when the buffer comes from a list of buffers on the command line (or `-f` which is effectively the same), so my design begins with a lexerStack.size() of more than 3600.

If this is intentional, and lexerStack is not only for nesting `` `include`` files, then why is this check?
```
        else if (lexerStack.size() >= options.maxIncludeDepth)
            addDiag(diag::ExceededMaxIncludeDepth, fileName.range());
```

I see that `options.maxIncludeDepth` is set to 1024, which is what kills my design.

MikePopoloski:
Ah yes, good debugging. The lexer stack is also used to implement the "single compilation unit" functionality by pushing all of the sources onto the stack, but of course those should not count toward the include file limit because they aren't include files (and aren't actually nested so there's nothing to protect against). I can fix that.

udif:
While we're at it, I was looking at the same area , and saw this:
```
        else if (includeOnceHeaders.find(buffer.data.data()) == includeOnceHeaders.end()) {
            pushSource(buffer);
        }
```
From the little I guess (Haven't done a serious search for this in the code yet), you probably implement something like `` `pragma once`` .
1. I was not aware of such a thing in other Verilog tools
2. What I learn from Wikipedia regarding `#pragma once` for C/C++, they typically match the filename while you match against actual file data?!

MikePopoloski:
[...]

PR #601, opened 2022-08-30
Allow redefining modules using a new --allow-redefinition flag

udif:
Here is my first trial of a PR to handle https://github.com/MikePopoloski/slang/issues/600#issuecomment-1232131424
Other than the boilerplate flags code, I'm not sure this code is 100%, even though it works for my toy example:

Given the 3 test files from #600 :
```
// a.v
module a;
b b_inst (.x(1'b0));
endmodule

// b0.v
module b;
endmodule

// b1.v
module b(input x);
endmodule
```

Without the new flag:
```
:~/git/slang/build$ bin/slang  b0.v a.v b1.v
Top level design units:
    a

a.v:2:12: error: port 'x' does not exist in 'b'
b b_inst (.x(1'b0));
           ^
b1.v:1:8: error: redefinition of 'b'
module b(input x);
       ^
b0.v:1:8: note: previous definition here
module b;
       ^

Build failed: 2 errors, 0 warnings
```

With the new flag:
```
~/git/slang/build$ bin/slang --allow-redefinition b0.v a.v b1.v
Top level design units:
    a

```

Comparing AST's reveals the output from `bin/slang --allow-redefinition b0.v a.v b1.v` and `bin/slang a.v b1.v` is almost identical:
Other than address changes, we have 1 extra `compilationUnit` , but other that, the JSONs are identical.

### DIFF
--- a/scripts/diagnostics.txt
+++ b/scripts/diagnostics.txt
@@ -91,6 +91,16 @@ warning extra-pragma-args ExtraPragmaArgs "too many arguments provided for pragm
 warning expected-diag-arg ExpectedDiagPragmaArg "expected diagnostic pragma argument"
 warning unknown-diag-arg UnknownDiagPragmaArg "unknown diagnostic pragma argument '{}'"
 warning pragma-diag-level ExpectedDiagPragmaLevel "expected diagnostic severity (ignore,warn,error,fatal)"
+warning expected-protect-keyword ExpectedProtectKeyword "expected pragma protect keyword"
+warning unknown-protect-keyword UnknownProtectKeyword "unknown pragma protect keyword '{}'"
+warning expected-protect-arg ExpectedProtectArg "'{}' pragma expects a single string argument"
+warning nested-protect-begin NestedProtectBegin "nested pragma protect begin-end blocks are not allowed"
+warning extra-protect-end ExtraProtectEnd "pragma protect '{}' seen without corresponding start of block"
+warning protect-arglist ProtectArgList "protect {} directive expects a parenthesized list of key=value options"
+warning unknown-protect-encoding UnknownProtectEncoding "unknown protect encoding '{}', treating as 'raw' -- source text may not be processed correctly"
+warning unknown-protect-option UnknownProtectOption "unknown protect {} option '{}'"
+warning invalid-pragma-number InvalidPragmaNumber "expected a 32-bit integer value for pragma option"
+warning invalid-pragma-viewport InvalidPragmaViewport "pragma protect viewport requires options to be specified in the form (object = <string>, access = <string>)"
 
 subsystem Parser
 error ExpectedIdentifier "expected identifier"
@@ -442,8 +452,6 @@ error BadOpenRange "invalid bounds in open range ({} and {})"
 error BadCastType "invalid casting type {}"
 error BadSetMembershipType "invalid type {} for {} expression"
 error ConcatWithStringInt "cannot mix strings and integers in a concatenation (use a cast if this is desired)"
-error BadRangeExpression "cannot select range of [{}:{}] from {}"
-error RangeWidthTooLarge "cannot select range of {} elements from {}"
 error RangeWidthOverflow "select range overflows the bounds of a signed 32-bit integer"
 error RangeSelectAssociative "cannot take a slice of an associative array"
 error CannotIndexScalar "scalar type cannot be indexed"
@@ -594,6 +602,9 @@ warning nonblocking-final NonblockingInFinal "nonblocking assignments in final b
 warning port-coercion InputPortCoercion "input net port '{}' coerced to 'inout'"
 warning port-coercion OutputPortCoercion "output net port '{}' coerced to 'inout'"
 warning reversed-range ReversedOpenRange "reversed range will behave as if it were empty"
+warning index-oob IndexOOB "cannot refer to element {} of {}"
+warning range-oob RangeOOB "cannot select range of [{}:{}] from {}"
+warning range-width-oob RangeWidthOOB "cannot select range of {} elements from {}"
 
 subsystem Statements
 error ReturnNotInSubroutine "return statement is only valid inside task and function blocks"
@@ -902,7 +913,10 @@ group default = { real-underflow real-overflow vector-overflow int-overflow unco
                   unknown-escape-code missing-top warning-task unsized-concat finish-num implicit-conv
                   nonblocking-final constraint-missing missing-format dpi-spec enum-range udp-port-empty
                   lifetime-prototype net-inconsistent port-coercion empty-body reversed-range
-                  invalid-source-encoding }
+                  invalid-source-encoding index-oob range-oob range-width-oob expected-protect-arg
+                  unknown-protect-keyword expected-protect-keyword nested-protect-begin extra-protect-end
+                  protect-arglist unknown-protect-encoding unknown-protect-option invalid-pragma-number
+                  invalid-pragma-viewport }
 
 group extra = { empty-member empty-stmt dup-import pointless-void-cast case-gen-none case-gen-dup
                 unused-result format-real ignored-slice task-ignored width-trunc dup-attr event-const
--- a/scripts/diagnostics.txt
+++ b/scripts/diagnostics.txt
@@ -725,7 +725,8 @@ error NTResolveArgModify "resolution function for nettype '{}' cannot modify its
 warning enum-range EnumRangeLiteral "enum element ranges must be integer literals"
 
 subsystem Lookup
-warning redefinition Redefinition "redefinition of '{}'"
+warning duplicate-definition DuplicateDefinition "duplicate definition of '{}'"
+error Redefinition "redefinition of '{}'"
 error RedefinitionDifferentType "redefinition of '{}' with a different type: {} vs {}"
 error AmbiguousWildcardImport "multiple imports found for identifier '{}'"
 note NoteImportedFrom "imported from here"
--- a/scripts/diagnostics.txt
+++ b/scripts/diagnostics.txt
@@ -714,7 +714,7 @@ error NTResolveArgModify "resolution function for nettype '{}' cannot modify its
 warning enum-range EnumRangeLiteral "enum element ranges must be integer literals"
 
 subsystem Lookup
-error Redefinition "redefinition of '{}'"
+warning redefinition Redefinition "redefinition of '{}'"
 error RedefinitionDifferentType "redefinition of '{}' with a different type: {} vs {}"
 error AmbiguousWildcardImport "multiple imports found for identifier '{}'"
 note NoteImportedFrom "imported from here"
--- a/scripts/diagnostics.txt
+++ b/scripts/diagnostics.txt
@@ -725,7 +725,6 @@ error NTResolveArgModify "resolution function for nettype '{}' cannot modify its
 warning enum-range EnumRangeLiteral "enum element ranges must be integer literals"
 
 subsystem Lookup
-warning duplicate-definition DuplicateDefinition "duplicate definition of '{}'"
 error Redefinition "redefinition of '{}'"
 error RedefinitionDifferentType "redefinition of '{}' with a different type: {} vs {}"
 error AmbiguousWildcardImport "multiple imports found for identifier '{}'"
@@ -785,6 +784,7 @@ error InstanceArrayEndianMismatch "endianness of selection [{}:{}] must match de
 error IfaceImportExportTarget "could not find target for {} of '{}'"
 error UnexpectedNameToken "unexpected use of '{}' keyword"
 warning dup-import DuplicateImport "duplicate import declaration is redundant"
+warning duplicate-definition DuplicateDefinition "duplicate definition of '{}'"
 
 subsystem SysFuncs
 error BadSystemSubroutineArg "invalid argument type {} to system {}"
@@ -917,7 +917,7 @@ group default = { real-underflow real-overflow vector-overflow int-overflow unco
                   invalid-source-encoding index-oob range-oob range-width-oob expected-protect-arg
                   unknown-protect-keyword expected-protect-keyword nested-protect-begin extra-protect-end
                   protect-arglist unknown-protect-encoding unknown-protect-option invalid-pragma-number
-                  invalid-pragma-viewport }
+                  invalid-pragma-viewport duplicate-definition }
 
 group extra = { empty-member empty-stmt dup-import pointless-void-cast case-gen-none case-gen-dup
                 unused-result format-real ignored-slice task-ignored width-trunc dup-attr event-const
--- a/source/compilation/Compilation.cpp
+++ b/source/compilation/Compilation.cpp
@@ -440,7 +440,8 @@ const Definition* Compilation::getDefinition(const ModuleDeclarationSyntax& synt
 }
 
 template<typename T, typename U>
-static void reportRedefinition(const Scope& scope, const T& newSym, const U& oldSym, DiagCode code = diag::Redefinition) {
+static void reportRedefinition(const Scope& scope, const T& newSym, const U& oldSym,
+                               DiagCode code = diag::Redefinition) {
     if (!newSym.name.empty()) {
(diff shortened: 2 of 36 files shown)


udif:
I left this as draft PR because I feel less confident about this PR. Too many places where this might break things by mistake.

udif:
The new commit changes the behavior so that while redefinition is allowed on `--allow-redefinition` , it is still issued as a warning, and not completely silent.
This warning can be turned off by `-Wno-redefinition`, but no hint about that is given.

In order to achieve this commit , I had to remove the const on `optionMap` generated by `diagnostic_gen.py` .
You probably won't like it, but I wanted to have a message that is an error if redefinition is not allowed (default), a warning if redefinition is allowed, and if it is a warning, it can also be turned off completely.

You will probably have comments on this PR, which is why it was issued as a DRAFT in the first place, and no tests were written yet.

source/compilation/Compilation.cpp
```diff
@@ -458,14 +458,23 @@ void Compilation::createDefinition(const Scope& scope, LookupLocation location,
     // Record that the given scope contains this definition. If the scope is a compilation unit, add
     // it to the root scope instead so that lookups from other compilation units will find it.
     auto targetScope = scope.asSymbol().kind == SymbolKind::CompilationUnit ? root.get() : &scope;
+    bool redefined = false;
     std::tuple key(def->name, targetScope);
     if (auto it = definitionMap.find(key); it != definitionMap.end()) {
         reportRedefinition(scope, *def, *it->second);
-        return;
+        if (options.allowRedefinition) {
```
MikePopoloski, line 465: This should probably only allow redefinition at the top level? i.e. not allow redefining nested modules.

udif:
> You're right, I don't love the changes to the optionsMap. I think you should do what I recently did in [84060b2](https://github.com/MikePopoloski/slang/commit/84060b2aa63a184d93065c0f23ff4b9aa4a51e76) for out of bounds selections. You make the diagnostic a warning in diagnostics.txt, inside Driver you promote the severity to an Error unless some other option (in this case allow-redefinition) says otherwise.

I did this. changed the error to warning in `diagnostics.txt`, and it works . However, the error is accompanied by a warning flag option ( `[-Wredefinition]` ):
```
:~/git/slang/build$ bin/slang  b0.v a.v b1.v
Top level design units:
    a

a.v:2:12: error: port 'x' does not exist in 'b'
b b_inst (.x(1'b0));
           ^
b1.v:1:8: error: redefinition of 'b' [-Wredefinition]
module b(input x);
       ^
b0.v:1:8: note: previous definition here
module b;
       ^

Build failed: 2 errors, 0 warnings
```

Adding `-Wno-redefinition` removes the warning even though this is an error and not a warning.
```
~/git/slang/build$ bin/slang -Wno-redefinition b0.v a.v b1.v
Top level design units:
    a

a.v:2:12: error: port 'x' does not exist in 'b'
b b_inst (.x(1'b0));
           ^

Build failed: 1 error, 0 warnings
```
I would expect an error class message not to have a `-W` option.

source/compilation/Compilation.cpp
udif: I removed any references to `topDefinitions`. It was added at an early stage due to unexplained failures on `ASSERT(!topDef);` , but it is not needed now.

> This should probably only allow redefinition at the top level? i.e. not allow redefining nested modules.

What exactly do you imply by the term "nested modules"? Do you mean any module which is not top level? Is there any special issue with those?

Th case I'm trying to resolve involves reading two non-identical copies of library files, definitely not top level modules. Each (large) library file contains a different set of dozens of small modules, some are common between the 2 files, but some are unique.

MikePopoloski:
Warnings can be promoted to error severity with `-Werror=` and demoted from error severity with `-Wno-error=` so it makes sense to me that it would show the warning option name, as it's a valid target for changing the severity. It does occur to me though that we should separate this out to a separate warning specifically for module/interface/program redefinition, as that error is used for other kinds of redefinition that we don't want suppressible. The naming here is tough as I call these things "Definitions" internally but there isn't a good standard name for them.

MikePopoloski:
In fact if you make the handling in createDefinition always take the newer rather than the older definition in the case of a conflict then you don't even need a separate option. It's sufficient for the "warning" to be an error by default and downgradable by the user if desired.

source/compilation/Compilation.cpp
MikePopoloski: A nested module is a module definition inside another module. Like so:
```
module m;
  module n;
  endmodule
endmodule
```
Such a module can only be instantiated within its parent definition, and is automatically instantiated once if not explicitly instantiated there. I don't think your change should apply to duplicate nested modules, since nobody would duplicate the definitions on purpose in the same file.

udif:
> Warnings can be promoted to error severity with `-Werror=` and demoted from error severity with `-Wno-error=` so it makes sense to me that it would show the warning option name, as it's a valid target for changing the severity

The problem above is that an error is removed with `-Wno-redefinition`, **not** with `-Wno-error=` . I think `-Wno-XXX` should not demote an error, only a warning. maybe change `findAndSet()` to take an additional parameter specifying the existing severity.

Also, I think `Wno-error` might be broken:
```
        else if (startsWith(arg, "error=")) {
            findAndSet(arg.substr(6), DiagnosticSeverity::Error, "-Werror=");
        }
        else if (startsWith(arg, "no-error=")) {
            findAndSet(arg.substr(9), DiagnosticSeverity::Error, "-Wno-error=");
        }
        else if (startsWith(arg, "no-")) {
            findAndSet(arg.substr(3), DiagnosticSeverity::Ignored, "-Wno-");
        }
```

I would expect:

| | |
|---|---
|`-Wxxx` | set `xxx` to be a warning. |
| `-Werror=xxx` | set `xxx` to be an error. |
|`-Wno-error=xxx` | set `xxx` to be ignored.|
|`-Wno-xxx` | sets `xxx` to be ignored **if it is already a warning** |

MikePopoloski:
Yes, I discovered and fixed that issue yesterday. Please pull the latest changes.

I guess it's arguable whether a command line like `-Werror=foo -Wno-foo` should still report the foo error or not. C++ compilers (i.e. gcc and clang) take the opinion that it should work the way it does with slang, i.e. the no-foo will suppress the warning even if its severity has been upgraded.

udif:
> In fact if you make the handling in createDefinition always take the newer rather than the older definition in the case of a conflict then you don't even need a separate option. It's sufficient for the "warning" to be an error by default and downgradable by the user if desired.

ok, let's try it.

> Yes, I discovered and fixed that issue yesterday. Please pull the latest changes.

I took a look at the new code and it makes sense now. You can guess I'm not playing much with GCC errors and warnings, as I wasn't aware of their convention.

BTW, clang-tidy has some warnings on the autogenerated `DiagCode.cpp`, and I even managed to totally break clang-tidy (core dump or internal error) in one of my earlier changes. something about \<eof\>, followed by a few hundred lines of clang-tidy stack trace.

source/compilation/Compilation.cpp
udif: Oh, I totally forgot about this SV thing, and I even didn't know it can be automatically instantiated.
I agree that this is irrelevant because in the example above, the scope of `n` is purely within `m`, and it cannot be defined outside `m`.
The only use for this for this feature that comes into my mind is when you incorporate multiple IP's in a SoC where each IP might contain modules with the same name, but of a different revision. you can use this to wrap it:

```
module external_IP1 (/* add ports here, matching IP_file top level */);
`include "IP1_file.v"
`include "IP1/module.v"
IP1 IP1_inst(.*); // connect IP to outside ports
endmodule
```
Now `module.v` will not be visible outside, and you can use a multiple versions of it .

udif:
Whooops... A failed unit test.

MikePopoloski:
Looks like you figured out the topDef assert -- basically the code was saying we should never have an existing entry in topDefinitions because if there was a redefinition we would have returned early from the function above, but after your change that's no longer true.

I think my comment above about defining a new warning for this case got missed above. The Redefinition diagnostic gets used in a bunch of other places that should not be able to be demoted, so we want a separate warning that applies to just to duplicate Definitions (modules / interfaces / programs). Otherwise your code seems like it's on the right track.

source/compilation/Compilation.cpp
```diff
@@ -458,19 +458,23 @@ void Compilation::createDefinition(const Scope& scope, LookupLocation location,
...
     if (auto it = definitionMap.find(key); it != definitionMap.end()) {
         reportRedefinition(scope, *def, *it->second);
-        return;
+        redefined = true;
     }
 
-    auto [it, inserted] = definitionMap.emplace(key, std::move(def));
-    ASSERT(inserted);
+    auto [it, inserted] = definitionMap.insert_or_assign(key, std::move(def));
+    if (!redefined) {
```
MikePopoloski, line 469: I think this ASSERT is no longer doing anything useful; either the entry is inserted or it's not because of an existing entry -- you're really only checking that the hashmap implements its own interface correctly which can be safely assumed I think.

source/compilation/Compilation.cpp
```diff
@@ -458,19 +458,23 @@ void Compilation::createDefinition(const Scope& scope, LookupLocation location,
...
+        ASSERT(inserted);
+    }
 
     auto result = it->second.get();
     if (targetScope == root.get()) {
         auto& topDef = topDefinitions[result->name].first;
```
MikePopoloski, line 475: Same here -- I would just rewrite this as `topDefinitions[result->name].first = result;` now that it's allowed for there to be duplicates.

MikePopoloski:
Maybe the new warning could be called something like `-Wduplicate-definition`

source/compilation/Compilation.cpp
udif: > This should probably only allow redefinition at the top level? i.e. not allow redefining nested modules.

Coming back to this again after  a fresh look at the code, a bit more knowledge about implicit creation of std::map entries when a key is not found, and generally reading the relevant sections of Compilation.cpp and Compilation.h over and over:

As far as Iunderstand, the code in its current form will also work for nested modules, not that it is useful or that I've even tested it, but I think that:
```
module x;
  module y;
  endmodule
  module y;
  endmodule
endmodule
```

Would behave this way:

* For module x, it would create an entry using `definitionMap.insert_or_emplace()` in line 468.
* For module y, it would create an entry using `definitionMap.insert_or_emplace()` in line 468. Since `targetScope` is not root, it would create a `topDefinition` entry by setting its bool flag to `true` (line 485). Looking at `topDefinitions[result->name].second = true;` indicates will have no legal `Definition` pointer, since there is no top level module y.
* For the 2nd module y, it would replace the old entry using `definitionMap.insert_or_emplace()` in line 468, as well as setting `redefined` to `true`. Since `targetScope` is not root, it would overwrite the old `topDefinition` entry, writing the 2nd tuple element to `true` again.

source/compilation/Compilation.cpp
udif: I was trying to replace:
```
    auto [it, inserted] = definitionMap.insert_or_assign(key, std::move(def));
    //if (!redefined) {
    //    ASSERT(inserted);
    //}

    auto result = it->second.get();
    if (targetScope == root.get()) {
        topDefinitions[result->name].first = result;
```
with something simpler:
```
    const auto& result = (definitionMap[key] = std::move(def));
    if (targetScope == root.get()) {
        topDefinitions[result->name].first = &(*result);
```
It compiles cleanly, passes unittest, but the ugly `&(*result)` syntax, and the fact that `std::uniq_ptr` , x-values (`std::move`) 
 is pretty new to me, scares me a bit.
I switched today to Ubuntu 22.04 (from 20.04) and clang-14 (from g++-9) since I never got g++ failures without clang, but got clang failures when g++ passed.

source/compilation/Compilation.cpp
udif: I agree, 

> Maybe the new warning could be called something like `-Wduplicate-definition`

I'm doing that, duplicating `reportRedefinition()` into `reportDuplicateDefinition()`, and while searching for all the calls of the former, it brings up the question of what happens if I have a module redefining a UDP, or vice versa. It could happen in theory (although I specifically don't need this specific case).

I also returned `redefinition` to be an error as before.

source/compilation/Compilation.cpp
MikePopoloski: You should be able to do a `result.get()` instead in your example, which is a bit nicer. I think you still want the insert_or_assign though, and then use the `inserted` result to run your `reportRedefinition`. Then you can get rid of the extra `find()` call.

source/compilation/Compilation.cpp
MikePopoloski: I think if nobody wants it right now then we should leave it an error. Always better to start out with tight restrictions and relax them if needed, IMO.

source/compilation/Compilation.cpp
MikePopoloski: Hmm actually, that suggestion to collapse the find() call doesn't work because we do want to see the old element for purposes of error / warning reporting before we overwrite it.

source/compilation/Compilation.cpp
udif: The question is whether to leave the old `-Wredefinition` as a warning that gets promoted to an error by default (so it can be changed back to warning/ignore by the user), or return it to its original state, where it was a real error that couldn't be overridden.
The new `-Wduplicate-definition` is created as a warning and promoted to an error by default, as was  `-Wredefinition`  in this PR until now.

MikePopoloski: You should return the old `Redefinition` diagnostic to an error.

source/compilation/Compilation.cpp
udif: Aaaand indeed it also triggers for nested modules (in a test named "Nested Modules", that now fails because it checks for the old diag code) - fixed ofcourse.

scripts/diagnostics.txt
```diff
@@ -725,6 +725,7 @@ error NTResolveArgModify "resolution function for nettype '{}' cannot modify its
 warning enum-range EnumRangeLiteral "enum element ranges must be integer literals"
 
 subsystem Lookup
+warning duplicate-definition DuplicateDefinition "duplicate definition of '{}'"
```
MikePopoloski, line 728: Can you move this to the bottom of this section (the convention is to put warnings at the end of their respective sections) and also add it to the `default` group of warnings that are on by default down at the bottom of the file?

source/driver/Driver.cpp
```diff
@@ -346,6 +346,10 @@ bool Driver::processOptions() {
     diagEngine.setErrorLimit((int)options.errorLimit.value_or(20));
     diagEngine.setDefaultWarnings();
 
+    // Demoted to warning just to get the -W flag options, but we really
```
MikePopoloski, line 349: Can you mention that this is to support compatibility with other tools? Something like "Some tools allow ignoring duplicate module/interface/program definitions, so this is a suppressible warning that we promote to an error by default."

source/compilation/Compilation.cpp
```diff
@@ -448,6 +448,15 @@ static void reportRedefinition(const Scope& scope, const T& newSym, const U& old
     }
 }
 
+template<typename T, typename U>
+static void reportDuplicateDefinition(const Scope& scope, const T& newSym, const U& oldSym) {
```
MikePopoloski, line 452: Since this basically just copies the reportRedefinition method, how about reusing that and making it take a DiagCode argument that you pass on to addDiag(). Then each call site can specify whether it's the warning or error version just by changing the DiagCode it passes.

scripts/diagnostics.txt
MikePopoloski: Also add documentation and an example to warning_docs.txt. When the docs are built it will run the slang tool on your snippet and ensure that it produces the specified warning, and then include the output in the docs. See https://sv-lang.com/warning-ref.html for the current output.

MikePopoloski (commented):
Ok, this is looking good now. Just some cleanup tasks and we'll be good.

MikePopoloski:
Perfect, thanks!

PR #614, opened 2022-09-14
Enhance SyntaxVisitor functionality, let it support non-const lvalue object

HungMingWu:
I  want this functionality.
``` cpp
struct Visitor {
   // ignored
};
MemberSyntax *ptr = ....;
MemberSyntax *clone_ptr = ptr->clone(alloc); // It should be deepClone, but assume it works
clone_ptr->visit(Visitor{});
```
I want to reuse to logic on `SyntaxVisitor`, like this.
``` cpp
struct Visitor  : public slang::SyntaxVisitor<Visitor> {
    void handle(IdentifierNameSyntax& decl) {
           // ignored
    }
    void handle(DeclaratorSyntax& decl) {
           // ignored
    }
};
```
it modified the cloned_object directly, but the current implementation doesn't support it.

### DIFF
--- a/include/slang/syntax/SyntaxNode.h
+++ b/include/slang/syntax/SyntaxNode.h
@@ -40,10 +40,6 @@ struct TokenOrSyntaxBase : public std::variant<Token, TNode> {
 
     /// Gets access to the object as a syntax node (throwing an exception
     /// if it's not actually a syntax node).
-    TNode node() { return std::get<1>(*this); }
-
-    /// Gets access to the object as a syntax node (throwing an exception
-    /// if it's not actually a syntax node).
     TNode node() const { return std::get<1>(*this); }
 
 protected:
--- a/include/slang/syntax/SyntaxNode.h
+++ b/include/slang/syntax/SyntaxNode.h
@@ -40,6 +40,10 @@ struct TokenOrSyntaxBase : public std::variant<Token, TNode> {
 
     /// Gets access to the object as a syntax node (throwing an exception
     /// if it's not actually a syntax node).
+    TNode node() { return std::get<1>(*this); }
+
+    /// Gets access to the object as a syntax node (throwing an exception
+    /// if it's not actually a syntax node).
     TNode node() const { return std::get<1>(*this); }
 
 protected:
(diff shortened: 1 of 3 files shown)


include/slang/syntax/SyntaxNode.h
```diff
@@ -38,6 +38,10 @@ struct TokenOrSyntaxBase : public std::variant<Token, TNode> {
     /// if it's not actually a token).
     Token token() const { return std::get<0>(*this); }
 
+    /// Gets access to the object as a syntax node (throwing an exception
```
MikePopoloski, line 41: I don't think you need this node() overload, it has the same return type as the const version so it's not really doing anything.

include/slang/syntax/SyntaxNode.h
```diff
@@ -232,18 +241,15 @@ class SeparatedSyntaxList : public SyntaxListBase {
     public:
         using ParentList = std::conditional_t<std::is_const_v<std::remove_pointer_t<U>>,
                                               const SeparatedSyntaxList, SeparatedSyntaxList>;
-        iterator_base(ParentList& list, size_t index) : list(&list), index(index) {}
-        iterator_base(const iterator_base& other) = default;
-
-        iterator_base& operator=(const iterator_base& other) = default;
```
MikePopoloski, line 238: Why did you change this? iterators are required to be copy assignable so removing this doesn't seem like a good idea.

HungMingWu: It seems identical to master branch.

MikePopoloski: Why do you say it's identical? There's a diff showing here of removing the copy assignment operator. Perhaps you're working off an old checkout and need to rebase?

HungMingWu: That's my mistake, I would fix it.

include/slang/syntax/SyntaxNode.h
HungMingWu: I want to modify inplace.
``` cpp
class TestRewriter : public SyntaxRewriter<TestRewriter> {
    void handle(TypedefDeclarationSyntax& decl) {
    };
};
auto clone_obj = obj->clone(alloc);
clone_obj->visit(TestRewriter {});
```
Icannot archive without overload node function due to constness

MikePopoloski: But they both return TNode, so how does this help? The `const` version will be called by even on non-const objects if there isn't this other overload, since it's always safe to call a const method on a non-const object.

HungMingWu: You are right, my visual C++ have success compile it. maybe my environment have some flawed.
It compiles ok on VC, GCC, clang

PR #617, opened 2022-09-17
Enhance syntax clone functionality, include deepClone

HungMingWu:
Here is my attempt to implement deepClone semantics in discussion #599 
Try to uniform API usage, any suggestion is welcome


include/slang/syntax/SyntaxNode.h
```diff
@@ -277,6 +310,9 @@ class SLANG_EXPORT SeparatedSyntaxList : public SyntaxListBase {
     SeparatedSyntaxList(span<TokenOrSyntax> elements) :
         SyntaxListBase(SyntaxKind::SeparatedList, elements.size()), elements(elements) {}
 
+    /// @return the elements of nodes in the list
+    [[nodiscard]] span<TokenOrSyntax> elems() const { return elements; }
```
MikePopoloski, line 314: I don't think this should return a span<TokenOrSyntax> if elems() is const, since that would allow modification of the underlying elements. I think there should be two overloads, the const one should return a span<const ConstTokenOrSyntax> and the non-const one can return what you have here.

include/slang/syntax/SyntaxNode.h
```diff
@@ -320,4 +356,18 @@ class SLANG_EXPORT SeparatedSyntaxList : public SyntaxListBase {
...
 
+template<typename T>
+SeparatedSyntaxList<T>* deepClone(const SeparatedSyntaxList<T>& node, BumpAllocator& alloc) {
+    SmallVectorSized<TokenOrSyntax, 8> buffer(node.size());
+    for (const auto& ele : node.elems()) {
+        if (ele.isToken()) {
+            buffer.append(ele);
+        }
+        else {
+            buffer.append(slang::syntax::deepClone(ele.node()->template as<T>(), alloc));
```
MikePopoloski, line 367: I don't think you need the as<T> here, since you have an overload that will take the SyntaxNode and do the correct thing anyway?

scripts/syntax_gen.py
```diff
@@ -729,6 +721,126 @@ class SyntaxNode;
     outf.write("\n}\n")
 
 
+def generateSyntaxClone(builddir, alltypes, kindmap):
+    # Start the clone source file.
+    clonef = open(os.path.join(builddir, "AllSyntaxClone.cpp"), "w")
```
MikePopoloski, line 726: Minor nit, but can you name this just SyntaxClone.cpp? The "All" part is unnecessary I feel.

scripts/syntax_gen.py
```diff
@@ -493,14 +493,6 @@ def generateSyntax(builddir, alltypes, kindmap):
 
             cppf.write("}\n\n")
 
-            cppf.write("{0}* {0}::clone(BumpAllocator& alloc) const {{\n".format(k))
-            if v.final:
-                cppf.write("    return alloc.emplace<{}>(*this);\n".format(k))
-            else:
-                cppf.write("    CloneVisitor<{0}> visitor;\n".format(k))
```
MikePopoloski, line 500: Can you also remove the CloneVisitor type from this file since it's not used now?

source/syntax/SyntaxNode.cpp
```diff
@@ -33,6 +33,25 @@ struct GetChildVisitor {
     TokenOrSyntax visitInvalid(SyntaxNode&, size_t) { return nullptr; }
 };
 
+template<typename T>
```
MikePopoloski, line 36: Why is this class duplicated here?

scripts/syntax_gen.py
```diff
@@ -729,6 +721,126 @@ class SyntaxNode;
...
+        """namespace slang::syntax {
+
+struct CloneVisitor {
+    template<typename U>
+    using clone_t = decltype(std::declval<U>().clone(std::declval<BumpAllocator&>()));
+
+    template<typename T>
+    SyntaxNode* visit(const T& node, BumpAllocator& alloc) {
+        if constexpr (is_detected_v<clone_t, T>) {
+            return node.clone(alloc);
```
MikePopoloski, line 806: Which types still have a .clone() member in them? It seems like you removed them all so this branch would be unnecessary. Maybe just the SyntaxListBase types? Perhaps those should be moved out into standalone methods too for consistency.

scripts/syntax_gen.py
```diff
@@ -729,6 +721,126 @@ class SyntaxNode;
...
+};
+
+struct DeepCloneVisitor {
+    template<typename U>
+    using deepClone_t = decltype(std::declval<U>().deepClone(std::declval<BumpAllocator&>()));
+
+    template<typename T>
+    SyntaxNode* visit(const T& node, BumpAllocator& alloc) {
+        if constexpr (is_detected_v<deepClone_t, T>) {
+            return node.deepClone(alloc);
```
MikePopoloski, line 821: I don't think any of the syntax type have a deepClone member method?

include/slang/syntax/SyntaxNode.h
```diff
@@ -137,6 +138,23 @@ class SLANG_EXPORT SyntaxNode {
     TokenOrSyntax getChild(size_t index);
 };
 
+SyntaxNode* clone(const SyntaxNode&, BumpAllocator&);
```
MikePopoloski, line 141: These functions need to be marked SLANG_EXPORT to work with shared library builds.

MikePopoloski (commented):
This approach seems ok to me. I left some comments to fix up. Also it looks like you need to run the pre-commit lint on the syntax_gen.py file to get that failed build cleared up. Note that you can run `pre-commit install` in the checked out repo and it will automatically run the linter every time you commit, which is convenient.

scripts/syntax_gen.py
HungMingWu: I try to implement on this way, but it fails.
Due to the SyntaxNode is cast into SyntaxListBase and it lost type information on runtime.
Maybe std::variant can solve the issue, but the code change is much bigger.

HungMingWu:
I have another thought about deepClone
I tried to implement the second version.
``` cpp
struct deepCloneVisitor {
    template<typename T>
    SyntaxNode* visit(const T& node, ) {
        auto newNode = clone(node, alloc); // Shallow Clone first
        for (uint32_t i = 0; i < node.getChildCount(); i++) {
            auto oldChild = node.childNode(i);
            auto newChild = node.childNode(i);
            if (child)
                oldChild = newChild;
        }
        return newNode;
    }
    SyntaxNode* visitInvalid(const SyntaxNode&) { return nullptr; }
};
```
But it meets different problem.
Do you think it is a good idea? Or keep the original implementation.

source/syntax/SyntaxNode.cpp
```diff
@@ -32,7 +32,6 @@ struct GetChildVisitor {
 
     TokenOrSyntax visitInvalid(SyntaxNode&, size_t) { return nullptr; }
 };
-
```
MikePopoloski, line 35: This seems like an errant change.

scripts/syntax_gen.py
MikePopoloski: Why doesn't deepClone have the same problem?

source/syntax/SyntaxNode.cpp
HungMingWu: Sorry, I don't know whats wrong about this?

MikePopoloski: It's just a random newline deletion in an otherwise unchanged file.

MikePopoloski:
Please also rebase your change to fix the build (and make sure pre-commit passes).

PR #620, opened 2022-09-19
Add a new '--ignore-directive <directive>' command line option.

udif:
Some tools have non-standard compiler directives.
In order to be able to filter those specific directives in legacy code, but not to miss undefined macros or spelling mistakes, a new `--ignore-directive <directive>` command line option was added.

Here is an example:

Test code:
```
~/git/slang/build$ cat ~/v/d.v
`dummy_directive
`dummy_directive 1

module top;
`dummy_directive2
endmodule
```
No flags:
```
~/git/slang/build$ bin/slang  ~/v/d.v
Top level design units:
    top

../../../v/d.v:1:1: error: unknown macro or compiler directive '`dummy_directive'
`dummy_directive
^
../../../v/d.v:2:1: error: unknown macro or compiler directive '`dummy_directive'
`dummy_directive 1
^
../../../v/d.v:2:18: error: expected member
`dummy_directive 1
                 ^
../../../v/d.v:5:1: error: unknown macro or compiler directive '`dummy_directive2'
`dummy_directive2
^

Build failed: 4 errors, 0 warnings
```
New flag:
```
~/git/slang/build$ bin/slang --ignore-directive dummy_directive  --ignore-directive dummy_directive2 ~/v/d.v
Top level design units:
    top

```

### DIFF
--- a/include/slang/driver/Driver.h
+++ b/include/slang/driver/Driver.h
@@ -73,6 +73,9 @@ public:
         /// A set of extensions that will be used to exclude files.
         flat_hash_set<std::string> excludeExts;
 
+        /// A set of preprocessor directives to be ignored
+        std::vector<std::string> ignoreDirectives;
+
         /// @}
         /// @name Preprocessing
         /// @{
--- a/include/slang/driver/Driver.h
+++ b/include/slang/driver/Driver.h
@@ -73,7 +73,7 @@ public:
         /// A set of extensions that will be used to exclude files.
         flat_hash_set<std::string> excludeExts;
 
-        /// A set of preprocessor directives to be ignored
+        /// A set of preprocessor directives to be ignored.
         std::vector<std::string> ignoreDirectives;
 
         /// @}
--- a/include/slang/parsing/Preprocessor.h
+++ b/include/slang/parsing/Preprocessor.h
@@ -50,7 +50,7 @@ struct SLANG_EXPORT PreprocessorOptions {
     /// A set of macro names to undefine at the start of file preprocessing.
     std::vector<std::string> undefines;
 
-    /// A set of preprocessor directives to be ignored
+    /// A set of preprocessor directives to be ignored.
     flat_hash_set<string_view> ignoreDirectives;
 };
 
@@ -148,7 +148,7 @@ private:
     Trivia handleIncludeDirective(Token directive);
     Trivia handleResetAllDirective(Token directive);
     Trivia handleDefineDirective(Token directive);
-    Trivia handleMacroUsage(Token directive);
+    std::pair<Trivia, Trivia> handleMacroUsage(Token directive);
     Trivia handleIfDefDirective(Token directive, bool inverted);
     Trivia handleElsIfDirective(Token directive);
     Trivia handleElseDirective(Token directive);
@@ -272,7 +272,7 @@ private:
 
     // Macro handling methods
     MacroDef findMacro(Token directive);
-    syntax::MacroActualArgumentListSyntax* handleTopLevelMacro(Token directive);
+    std::pair<syntax::MacroActualArgumentListSyntax*, Trivia> handleTopLevelMacro(Token directive);
     bool expandMacro(MacroDef macro, MacroExpansion& expansion,
                      syntax::MacroActualArgumentListSyntax* actualArgs);
     bool expandIntrinsic(MacroIntrinsic intrinsic, MacroExpansion& expansion);
--- a/include/slang/parsing/Preprocessor.h
+++ b/include/slang/parsing/Preprocessor.h
@@ -49,6 +49,9 @@ struct SLANG_EXPORT PreprocessorOptions {
 
     /// A set of macro names to undefine at the start of file preprocessing.
     std::vector<std::string> undefines;
+
+    /// A set of preprocessor directives to be ignored
+    flat_hash_set<string_view> ignoreDirectives;
 };
 
 /// Preprocessor - Interface between lexer and parser
--- a/source/driver/Driver.cpp
+++ b/source/driver/Driver.cpp
@@ -18,10 +18,8 @@
 #include "slang/diagnostics/ParserDiags.h"
 #include "slang/diagnostics/SysFuncsDiags.h"
 #include "slang/diagnostics/TextDiagnosticClient.h"
-#include "slang/parsing/Lexer.h"
 #include "slang/parsing/Parser.h"
 #include "slang/parsing/Preprocessor.h"
-#include "slang/syntax/SyntaxKind.h"
 #include "slang/syntax/SyntaxPrinter.h"
 #include "slang/syntax/SyntaxTree.h"
 
@@ -614,13 +612,12 @@ Bag Driver::createOptionBag() const {
     ppoptions.predefineSource = "<command-line>";
     if (options.maxIncludeDepth.has_value())
         ppoptions.maxIncludeDepth = *options.maxIncludeDepth;
+    for (const auto& d : options.ignoreDirectives)
+        ppoptions.ignoreDirectives.emplace(std::string_view(d));
 
     LexerOptions loptions;
     if (options.maxLexerErrors.has_value())
         loptions.maxErrors = *options.maxLexerErrors;
-    for (const auto& d : options.ignoreDirectives)
-        loptions.ignoreDirectives.push_back(
-            std::make_pair(std::string_view(d), syntax::SyntaxKind::UnknownDirective));
 
     ParserOptions poptions;
     if (options.maxParseDepth.has_value())
--- a/source/driver/Driver.cpp
+++ b/source/driver/Driver.cpp
@@ -18,12 +18,12 @@
 #include "slang/diagnostics/ParserDiags.h"
 #include "slang/diagnostics/SysFuncsDiags.h"
 #include "slang/diagnostics/TextDiagnosticClient.h"
-#include "slang/parsing/Parser.h"
 #include "slang/parsing/Lexer.h"
+#include "slang/parsing/Parser.h"
 #include "slang/parsing/Preprocessor.h"
+#include "slang/syntax/SyntaxKind.h"
 #include "slang/syntax/SyntaxPrinter.h"
 #include "slang/syntax/SyntaxTree.h"
-#include "slang/syntax/SyntaxKind.h"
 
 namespace slang::driver {
 
@@ -79,10 +79,8 @@ void Driver::addStandardArgs() {
         "--cmd-rename", [this](string_view value) { return cmdLine.addRenameCommand(value); },
         "Define rule to rename vendor command <vendor_cmd> into existing <slang_cmd>",
         "<vendor_cmd>,<slang_cmd>");
-    cmdLine.add(
-        "--ignore-directive", options.ignoreDirectives,
-        "Ignore preprocessor directive and all its arguments until EOL", "<directive>");
-
+    cmdLine.add("--ignore-directive", options.ignoreDirectives,
+                "Ignore preprocessor directive and all its arguments until EOL", "<directive>");
 
     // Parsing
     cmdLine.add("--max-parse-depth", options.maxParseDepth,
@@ -622,8 +620,7 @@ Bag Driver::createOptionBag() const {
         loptions.maxErrors = *options.maxLexerErrors;
     for (const auto& d : options.ignoreDirectives)
         loptions.ignoreDirectives.push_back(
-            std::make_pair(std::string_view(d),
-                           syntax::SyntaxKind::UnknownDirective));
+            std::make_pair(std::string_view(d), syntax::SyntaxKind::UnknownDirective));
 
     ParserOptions poptions;
     if (options.maxParseDepth.has_value())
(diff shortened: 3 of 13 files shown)


udif:
Regrading the code, I added a new constructor for `StringTable<>` and merged the lookup of the static directives table in LexerFacts.cpp with a dynamic table built by the command line options. 
I did this in order to minimize the runtime impact of looking at two lists, one of legal directives, and one of directives to be ignored. As a result, not using this function should have zero impact on runtime.

udif:
rebase due to unrelated CI failure on macos (maybe due to not being on master when releasing the PR).

udif:
According to the github action logs, build actually broke due to https://github.com/MikePopoloski/slang/commit/642ef985fae186818b8bfce4ba1d5ddfab8a4955

MikePopoloski:
The build error should be fixed now, please rebase.

I think this could be done much more simply by having the ignore list in the Preprocessor instead, and looking at it separately in handleTopLevelMacro() if we fail to look up a macro usage. You could then check if the "macro usage" is instead one of your ignored directives and consume tokens if so. That way you don't need to modify StringTable, create new syntax kinds, etc. This doesn't add any overhead unless a macro usage is not found, which is either an error case and so not expected in valid code or it's a usage of this new feature.

udif:
rebased due to CI failure fix.

udif:
I've pushed a new version that works at the preprocessor level as suggested. It passes the same manual tests appearing in the top post.
The only remaining issue is how to handle the ignored directives. My intent was to save them as trivia, but I wasn't sure exactly how to do this, as the original code returned a `nullptr` instead of a trivia in a token when the macro was not recognized.

source/parsing/Preprocessor_macros.cpp
```diff
@@ -59,6 +59,17 @@ void Preprocessor::createBuiltInMacro(string_view name, int value, string_view v
 MacroActualArgumentListSyntax* Preprocessor::handleTopLevelMacro(Token directive) {
```
MikePopoloski, line 59: Probably you need to make this function return a std::pair<MacroActualArgumentListSyntax*, Trivia> and return the skipped tokens as a Trivia(TriviaKind::SkippedTokens...). handleMacroUsage then gets changed to return a std::pair<Trivia, Trivia> and returns this extra trivia result as the second member, and then the usage inside handleDirective checks for that trivia and adds it to the list if needed.

source/parsing/Preprocessor_macros.cpp
```diff
@@ -59,6 +59,17 @@ void Preprocessor::createBuiltInMacro(string_view name, int value, string_view v
...
     auto macro = findMacro(directive);
     if (!macro.valid()) {
+        if (options.ignoreDirectives.find(directive.valueText().substr(1)) !=
+            options.ignoreDirectives.end()) {
+            SmallVectorSized<Trivia, 16> trivia;
+
+            trivia.append(createSimpleDirective(directive));
```
MikePopoloski, line 66: If you follow the approach I outlined above then you don't need to (and really shouldn't) create a simple directive. The directive token will be taken by the MacroUsage syntax in the parent method.

source/parsing/Preprocessor_macros.cpp
```diff
@@ -59,6 +59,17 @@ void Preprocessor::createBuiltInMacro(string_view name, int value, string_view v
...
+            while (currentToken.kind != TokenKind::EndOfFile && peekSameLine())
+                consume();
```
MikePopoloski, line 68: These tokens should be added to the trivia array here.

source/parsing/Preprocessor_macros.cpp
```diff
@@ -59,6 +59,17 @@ void Preprocessor::createBuiltInMacro(string_view name, int value, string_view v
...
+            trivia.appendRange(directive.trivia());
```
MikePopoloski, line 69: No need to take the directive's trivia. Trivia needs to be listed in-order such that when you traverse tokens in-order and print them and their trivia you get back the original source file, so the ordering actually matters quite a bit.

include/slang/driver/Driver.h
```diff
@@ -73,6 +73,9 @@ class SLANG_EXPORT Driver {
         /// A set of extensions that will be used to exclude files.
         flat_hash_set<std::string> excludeExts;
 
+        /// A set of preprocessor directives to be ignored
```
MikePopoloski, line 76: Small nit, can you end these comments with a period to match the other comments in this file and elsewhere?

include/slang/parsing/Preprocessor.h
```diff
@@ -49,6 +49,9 @@ struct SLANG_EXPORT PreprocessorOptions {
 
     /// A set of macro names to undefine at the start of file preprocessing.
     std::vector<std::string> undefines;
+
+    /// A set of preprocessor directives to be ignored
```
MikePopoloski, line 53: Same here.

source/driver/Driver.cpp
```diff
@@ -610,6 +612,8 @@ Bag Driver::createOptionBag() const {
     ppoptions.predefineSource = "<command-line>";
     if (options.maxIncludeDepth.has_value())
         ppoptions.maxIncludeDepth = *options.maxIncludeDepth;
+    for (const auto& d : options.ignoreDirectives)
+        ppoptions.ignoreDirectives.emplace(std::string_view(d));
```
MikePopoloski, line 616: I don't think the cast is necessary?

udif:
I fixed locally all the trivial changes (comments, casts), but I still need to go over the Trivia and Token classes and member functions, to better understand them.

MikePopoloski:
rawData() should return a pointer into the source buffer (not a temporary). Possibly you're not looking at the right buffer start pointer?

Regardless though you shouldn't need this; for SkippedToken trivia you can use the Trivia() constructor that takes a list of tokens skipped instead of trying to compute the raw text (which doesn't in general work anyway since the tokens that come from, say, a macro expansion will reside in a different buffer).

udif:
What is the simplest way to check if the `Trivia` nodes are added correctly? I noticed the `rewriter` tool doesn't take any command line arguments so I can't pass any `--ignore-directive` flag there. The slang JSON dump don't list those either.

include/slang/driver/Driver.h
udif: Done.

source/driver/Driver.cpp
udif: Done.

tests/unittests/PreprocessorTests.cpp
```diff
@@ -2189,3 +2189,27 @@ endmodule // secret
...
+)";
+
+    auto result = preprocess(text);
+    CHECK(result == "\n");
+
+    REQUIRE(diagnostics.size() == 1);
+    CHECK(diagnostics.back().code == diag::UnknownDirective);
+}
+
+TEST_CASE("Unknown but ignored directive") {
```
MikePopoloski, line 2205: Your test should include the "consuming extra tokens" feature as well.

MikePopoloski (commented):
You can test that you got the trivia correct by round tripping some source code back to itself in the PreprocessorTests. Basically you can create a SyntaxTree from some source code and then run SyntaxPrinter::printFile() on it and the result should exactly equal the original source.

MikePopoloski:
Other than the testing stuff though this looks good to me.

udif:
OK, I have this test:
```
TEST_CASE("Unknown but ignored directive") {
    auto& text = R"(
`unknown_pragma xyz abc 123
)";
    PreprocessorOptions ppOptions;
    ppOptions.ignoreDirectives.emplace("unknown_pragma");

    Bag options;
    options.set(ppOptions);

    auto result = preprocess(text, options);
    CHECK(result == "\n");

    REQUIRE(diagnostics.size() == 0);

    auto tree = SyntaxTree::fromText(text, SyntaxTree::getDefaultSourceManager(), "", "", options);
    CHECK(SyntaxPrinter::printFile(*tree) == text);
```
It is passing as long as the skipped preprocessor directive has no arguments.
When it has arguments (like in this example), I get a SIGSEGV.
```
slang::parsing::Token::location(const slang::parsing::Token * this) (\home\udif\git\slang\source\parsing\Token.cpp:274)
slang::parsing::Trivia::getExplicitLocation(const slang::parsing::Trivia * this) (\home\udif\git\slang\source\parsing\Token.cpp:122)
slang::syntax::SyntaxPrinter::print(slang::syntax::SyntaxPrinter * this, slang::parsing::Token token) (\home\udif\git\slang\source\syntax\SyntaxPrinter.cpp:70)
slang::syntax::SyntaxPrinter::print(slang::syntax::SyntaxPrinter * this, const slang::syntax::SyntaxNode & node) (\home\udif\git\slang\source\syntax\SyntaxPrinter.cpp:99)
slang::syntax::SyntaxPrinter::print(slang::syntax::SyntaxPrinter * this, const slang::syntax::SyntaxTree & tree) (\home\udif\git\slang\source\syntax\SyntaxPrinter.cpp:105)
slang::syntax::SyntaxPrinter::printFile[abi:cxx11](slang::syntax::SyntaxTree const&)(const slang::syntax::SyntaxTree & tree) (\home\udif\git\slang\source\syntax\SyntaxPrinter.cpp:117)
CATCH2_INTERNAL_TEST_244() (\home\udif\git\slang\tests\unittests\PreprocessorTests.cpp:2221)
...
```
Apparently, it is looking for an "explicit location" of the trivia in `info->location`.

source/parsing/Preprocessor_macros.cpp
```diff
@@ -56,16 +56,29 @@ void Preprocessor::createBuiltInMacro(string_view name, int value, string_view v
...
     if (!macro.valid()) {
+        if (options.ignoreDirectives.find(directive.valueText().substr(1)) !=
+            options.ignoreDirectives.end()) {
+            std::vector<Token> ignore_tokens;
+            while (currentToken.kind != TokenKind::EndOfFile && peekSameLine())
+                ignore_tokens.push_back(consume());
+            if (ignore_tokens.empty())
+                return std::make_pair(nullptr, emptyTrivia);
+            else
+                return std::make_pair(nullptr, Trivia(TriviaKind::SkippedTokens, ignore_tokens));
```
MikePopoloski, line 72: I missed this on previous review. The ignore_tokens vector is on the stack; once you return you're referring to freed memory. You need to make a copy on the heap in order to refer to it like this.

I would rewrite this to use a SmallVector instead and use its copy() method to make a heap copy via the local allocator. You can find examples of this pattern all over the place in the codebase.

udif:
Bottom line, your suggestion works.
Before trying that, I spent an hour writing down why this shouldn't matter.

_1 Hour earlier..._
I made a small change to the code above, replacing `ignore_tokens` with a different `std::vector<Token>` built using a copy constructor, as well as breaking it into two lines so I can more easily examine the return value in the debugger.
```
            else {
                auto rval = std::make_pair(nullptr, Trivia(TriviaKind::SkippedTokens, std::vector<Token>(ignore_tokens)));
                return rval;
            }
```
I then checked `rval.second.tokens.ptr` and `rval.second.tokens.len` then compared it to  `actualArgs.second.tokens.ptr` and `actualArgs.second.tokens.len` in the calling function,  `handleMacroUsage`:
```
    const std::pair<MacroActualArgumentListSyntax*, Trivia>& actualArgs = handleTopLevelMacro(
        directive);
```

The result was that the ptr & len in both cases was identical, and was heap allocated (stack addresses start at 0x8000000000000000 grow downwards, while for heap I see 0x555555... )
I then removed the copy-constructor I added above, and used `ignore_tokens` directly as before.
what I see is that `ignore_tokens` is built on the stack, but it's data array , `&ignore_tokens[0]` is allocated on the heap (which is expected, given that it needs to be able to grow). Following this back one level up, `actualArgs.second.tokens.ptr` is holding the same heap address as `&ignore_tokens[0]`.

_Back to present time..._
So even in the original case, the `Token*` pointer in `Trivia` is pointing to a heap-allocated object, so why did replacing it with `SmallVectorSized` and calling it's `copy` method made a difference?

MikePopoloski:
To be more clear about my previous comment and why this doesn't work -- yes, the vector allocates its memory on the heap, but when the vector itself goes out of scope its destructor is called and the memory is freed. The Trivia object takes a span<> pointing to the memory containing the skipped tokens -- once the memory is freed that pointer is no longer valid. Running a copy constructor doesn't help you because the copied object will be destroyed at the end of that line of code and again the pointed to memory will be invalid.

SmallVector works here because the copy() method copies the memory into the provided allocator object, which has an appropriate lifetime and will keep it around for as long as needed. This is achievable without SmallVector of course by just using the allocator object directly and doing the copy yourself but this way is more convenient.

udif:
OK, I missed that.

When is the memory allocated by the bump allocated freed? Looking around at the code, I assume that all memory allocated inside the preprocessor using the `alloc` object, is bulk-freed at the end of preprocessing when called by `driver::runPreprocessor`  or at the end of parse tree creation if called by `SyntaxTree::create` ?

I've also tried to do the same with `new` (which means the object would be lost since no one would free it), only to find out there is no way to do copy initialization as part of object allocation by `new`.

Given that this is already fixed and committed, Is there anything else blocking this PR?

MikePopoloski:
The BumpAllocator lives inside the SyntaxTree that owns all the memory for that tree. When it gets destroyed all memory related to that tree gets freed (including preprocessor stuff, tokens, syntax nodes, etc).

Issue #588, opened 2022-08-18
Rewriting syntax tree should also update parser metadata


MikePopoloski:
Updating the metadata should happen automatically if possible. Right now there's no easy way to do it.

### Discussed in https://github.com/MikePopoloski/slang/discussions/572

<div type='discussions-op-text'>

<sup>Originally posted by **HungMingWu** August  2, 2022</sup>
Hi Mike,
I have one question.
Can I operator like this?
``` cpp
class TestRewriter : public SyntaxRewriter<TestRewriter> {
public:
    Compilation compilation;
    BumpAllocator alloc;
    SemanticModel model;

    TestRewriter(const std::shared_ptr<SyntaxTree>& tree) : model(compilation) {
        compilation.addSyntaxTree(tree);
    }

    ~TestRewriter() { compilation.steal(std::move(alloc)); }

    void handle(const HierarchyInstantiationSyntax& decl) {
        auto &new_decl = factory.hierarchyInstantiation(decl.attributes, decl.type, decl.parameters, decl.instances, decl.semi);
        auto type = decl.type.withRawText(alloc, "abcde");
        new_decl.type = type;
        replace(decl, new_decl);
    }
};
auto tree = SyntaxTree::fromBuffer(buffer, sourceManager, options);
tree = TestRewriter(tree).transform(tree);
compilation.addSyntaxTree(tree);
```
In my brief test, The metadata have been changed.
So I need to serialize to string and  construct it again?
</div>

MikePopoloski:
Fixed in 2eac3711f5ef07b6f238194f635d75c70219a4ee

PR #621, opened 2022-09-21
Fix metadata invalid issue via SyntaxRewriter

HungMingWu:
Issue #588 
Discussion #572 
Here is my patch, but I don't have enough confidence on verilog processing.

### DIFF
--- a/include/slang/syntax/SyntaxTree.h
+++ b/include/slang/syntax/SyntaxTree.h
@@ -37,6 +37,7 @@ public:
     bool isLibrary = false;
 
     SyntaxTree(SyntaxNode* root, SourceManager& sourceManager, BumpAllocator&& alloc,
+               std::unique_ptr<slang::parsing::ParserMetadata> metadata,
                std::shared_ptr<SyntaxTree> parent = nullptr);
 
     SyntaxTree(SyntaxTree&& other) = default;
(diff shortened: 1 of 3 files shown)


MikePopoloski:
Hmm, I'm not sure about this re-parsing approach. I think how I would approach it is to have a method that can build all of the ParserMetadata from scratch given a SyntaxTree. Right now it's built in the parser on the fly as an optimization but for the rewriting case I think it should just start with an empty metadata, traverse the syntax tree, and build it all up again. That seems less error prone than trying to catch all the ways a rewrite operation might affect the metadata and applying the fixups in place.

HungMingWu:
> Hmm, I'm not sure about this re-parsing approach. I think how I would approach it is to have a method that can build all of the ParserMetadata from scratch given a SyntaxTree. Right now it's built in the parser on the fly as an optimization but for the rewriting case I think it should just start with an empty metadata, traverse the syntax tree, and build it all up again. That seems less error prone than trying to catch all the ways a rewrite operation might affect the metadata and applying the fixups in place.

Can you give some pseudo code to me?
I am not familiar with this part.

MikePopoloski:
I mean to write a SyntaxVisitor that runs on the new tree and has handlers for all the places inside Parser that currently add metadata, and do a similar operation in the visitor. I don't have time to write it out right now, but if you wait I'll get around to implementing it eventually.

HungMingWu:
> if you wait I'll get around to implementing it eventually.
Ok, I will wait for your solution.

MikePopoloski:
Superseded by 2eac3711f5ef07b6f238194f635d75c70219a4ee

PR #624, opened 2022-09-29
Bugfix, fix deepClone crash issue on token copy

HungMingWu:
This is a bugfix for deepClone.
In the previous implementation, token would adopt shallow copy.
So with different life cycle BumpAllocators, it would crash.
Here is my test case.
``` cpp
        auto tree = SyntaxTree::fromText(R"(module m; reg tmp; endmodule)");
        class CloneRewriter : public SyntaxVisitor<CloneRewriter> {
            std::shared_ptr<SyntaxTree>& tree;

        public:
            CloneRewriter(std::shared_ptr<SyntaxTree>& tree) : tree(tree) {}
            void handle(const ModuleDeclarationSyntax& decl) {
                BumpAllocator newAlloc;
                auto newModule = slang::syntax::deepClone(decl, newAlloc);
                tree = std::make_shared<SyntaxTree>(newModule, tree->sourceManager(),
                                                    std::move(newAlloc));
            }
        };
        CloneRewriter visitor(tree);
        tree->root().visit(visitor);
```
Without modification, the original BumpAllocator would be destroyed, and it would meet use-after-free issue.

### DIFF
--- a/source/parsing/Token.cpp
+++ b/source/parsing/Token.cpp
@@ -153,6 +153,34 @@ span<Token const> Trivia::getSkippedTokens() const {
     return {tokens.ptr, tokens.len};
 }
 
+Trivia Trivia::clone(BumpAllocator& alloc) const {
+    Trivia result;
+    result.kind = kind;
+    result.hasFullLocation = hasFullLocation;
+
+    switch (kind) {
+        case TriviaKind::Directive:
+        case TriviaKind::SkippedSyntax:
+            result.syntaxNode = syntaxNode;
+            break;
+        case TriviaKind::SkippedTokens:
+            result.tokens = tokens;
+            break;
+        default:
+            if (hasFullLocation) {
+                result.fullLocation = alloc.emplace<FullLocation>();
+                result.fullLocation->text = fullLocation->text;
+                result.fullLocation->location = fullLocation->location;
+            }
+            else {
+                result.rawText = rawText;
+            }
+            break;
+    }
+
+    return result;
+}
+
 Token::Token() :
     kind(TokenKind::Unknown), missing(false), triviaCountSmall(0), reserved(0), numFlags() {
 }
@@ -271,6 +299,8 @@ SourceRange Token::range() const {
 }
 
 SourceLocation Token::location() const {
+    if (!info)
+        return SourceLocation::NoLocation;
     return info->location;
 }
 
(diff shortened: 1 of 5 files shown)


source/parsing/Token.cpp
```diff
@@ -370,6 +372,13 @@ Token Token::clone(BumpAllocator& alloc, span<Trivia const> trivia, string_view
     return result;
 }
 
+Token Token::deepClone(BumpAllocator& alloc) const {
+    SmallVectorSized<Trivia, 32> triviaBuffer;
+    for (const auto& t : trivia())
+        triviaBuffer.append(t.withLocation(alloc, location()));
```
MikePopoloski, line 378: You should not use withLocation for this, since it results in different trivia than what we originally had here. You might instead need a deepClone method on the Trivia object as well, since a trivia might point to an allocation in the old allocator (like spans of tokens for example).

source/parsing/Token.cpp
```diff
@@ -153,6 +153,25 @@ span<Token const> Trivia::getSkippedTokens() const {
     return {tokens.ptr, tokens.len};
 }
 
+Trivia Trivia::deepClone(BumpAllocator& alloc, SourceLocation location) const {
```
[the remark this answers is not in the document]
HungMingWu: I don't know how to handle deepClone properly without location semantics.
Maybe you can give me some hints.

HungMingWu:
I have update deepClone implementation, but is it allright?
``` cpp
    for (const auto& t : trivia()) {
        Trivia cloned_trivia = t.deepClone(alloc);
        cloned_trivia.withLocation(location());
        triviaBuffer.append(cloned_trivia);
    }
```

source/parsing/Token.cpp
```diff
@@ -153,6 +153,22 @@ span<Token const> Trivia::getSkippedTokens() const {
     return {tokens.ptr, tokens.len};
 }
 
+Trivia Trivia::deepClone(BumpAllocator& alloc) const {
```
MikePopoloski, line 156: I am confused about this function. The point is to clone the Trivia so that it's exactly the same, just in a new object yes? You need to check what kind of trivia it is and clone it appropriately, not always force it to be a full location trivia. The result should only be a full location trivia if the source trivia also has a full location (which is pretty rare btw). 

Basically you want to clone whichever member of the union is valid, which you can know by looking at the `kind` member.

MikePopoloski: So something like this: 
```
Trivia Trivia::clone(BumpAllocator& alloc) const {
    Trivia result;
    result.kind = kind;

    switch (kind) {
        case TriviaKind::Directive:
        case TriviaKind::SkippedSyntax:
            result.syntaxNode = syntaxNode;
            break;
        case TriviaKind::SkippedTokens:
            result.tokens = tokens;
            break;
        default:
            if (hasFullLocation) {
                result.fullLocation = alloc.emplace<FullLocation>();
                result.fullLocation->text = fullLocation->text;
                result.fullLocation->location = fullLocation->location;
                result.hasFullLocation = true;
            }
            else {
                result.rawText = rawText;
            }
            break;
    }

    return result;
}
```

PR #636, opened 2022-10-14
added "implements" and "extends" fields to ast-json output for ClassType elements

pistoletpierre:


MikePopoloski:
Thanks for the PR. This logic should just live in ClassType::serializeTo instead of inside the general serialize method. There's already some other logic there to serialize other class fields besides implements / extends.

source/ast/symbols/ClassSymbols.cpp
```diff
@@ -711,6 +711,13 @@ void ClassType::serializeTo(ASTSerializer& serializer) const {
         serializer.write("forward", *firstForward);
     if (genericClass)
         serializer.writeLink("genericClass", *genericClass);
+    auto syntax = getSyntax();
+    ASSERT(syntax);
+    auto& classSyntax = syntax->as<syntax::ClassDeclarationSyntax>();
```
MikePopoloski, line 716: I don't think you want to be writing the stringified syntax here -- it means all consumers have to reimplement SystemVerilog parsing in order to understand the output. We've already done the work to get this into AST format, so you might as well write the rich structured information here instead. For example, call the getBaseClass() method and write that type (if non-null). Call getImplementedInterfaces() and serialize that as a JSON array, etc.

source/ast/symbols/ClassSymbols.cpp
```diff
@@ -699,18 +703,45 @@ void ClassType::handleImplements(const ImplementsClauseSyntax& implementsClause,
...
         serializer.write("forward", *firstForward);
     if (genericClass)
         serializer.writeLink("genericClass", *genericClass);
+
+    serializer.writeProperty("extends");
+    serializer.startObject();
+    if (getBaseClass()) {
+        serializer.write("baseClassName", getBaseClass()->getCanonicalType().toString());
+
+        auto extendsClause = getSyntax()->as<syntax::ClassDeclarationSyntax>().extendsClause;
```
MikePopoloski, line 726: Again, you don't want to be serializing syntax nodes here. Use getBaseConstructorCall() to serialize the AST-form of the expression.

source/ast/symbols/ClassSymbols.cpp
```diff
@@ -699,18 +703,45 @@ void ClassType::handleImplements(const ImplementsClauseSyntax& implementsClause,
...
 void ClassType::serializeTo(ASTSerializer& serializer) const {
     if (firstForward)
         serializer.write("forward", *firstForward);
     if (genericClass)
         serializer.writeLink("genericClass", *genericClass);
+
+    serializer.writeProperty("extends");
+    serializer.startObject();
+    if (getBaseClass()) {
+        serializer.write("baseClassName", getBaseClass()->getCanonicalType().toString());
```
MikePopoloski, line 724: You should just write the type instead of diving down and toString()ing it yourself.

source/ast/symbols/ClassSymbols.cpp
```diff
@@ -540,22 +540,23 @@ const Expression* ClassType::getBaseConstructorCall() const {
...
 static void findIfaces(const ClassType& type, SmallVectorBase<const Type*>& ifaces,
-                       SmallSet<const Symbol*, 4>& visited) {
+                       SmallSet<const Symbol*, 4>& visited, bool recurse) {
     if (type.isInterface) {
         if (visited.emplace(&type).second)
             ifaces.push_back(&type);
     }
 
-    for (auto iface : type.getImplementedInterfaces()) {
+    for (auto iface :
```
MikePopoloski, line 552: This seems wrong; it's still recursive either way -- you're pulling in each implemented interface's first level of parent interfaces as well. You don't need to call findIfaces() at all for the direct list, just add the interfaces in the loop below as you encounter them.

include/slang/ast/symbols/ClassSymbols.h
```diff
@@ -71,6 +71,18 @@ class SLANG_EXPORT ClassType : public Type, public Scope {
...
     }
 
+    /// Gets the list of interface classes that this class implements.
+    /// If this class is itself an interface class, this is instead the list of
+    /// interface classes that it extends from, if any.
+    ///
+    /// Note that this list is *not* flattened from the full set of all interfaces implemented
+    /// by any base classes or interface class parents, up the inheritance hierarchy.
+    /// It only includes those mentioned as being implemented in the class definition.
+    span<const Type* const> getImplementedInterfacesImmediate() const {
```
MikePopoloski, line 81: I would call this "getDirectlyImplementedInterfaces" instead of using the "immediate" term.

MikePopoloski:
This PR has been superseded by 7addb9d810ea7aca3a5c85e82030e47a3460429e

Issue #637, opened 2022-10-17
Null dereference introduced by recent SyntaxRewriter changes


HungMingWu:
**Describe the bug**
Null dereference introduced by recent SyntaxRewriter changes
Since you append a commit, it still happens at these case.

``` cpp
template<typename T>
class SLANG_EXPORT SyntaxList : public SyntaxListBase, public span<T*> {
    void resetAll(BumpAllocator& alloc, span<const TokenOrSyntax> children) final {
        SmallVector<T*> buffer(children.size(), UninitializedTag());
        for (auto& t : children)
            buffer.push_back(&t.node()->as<T>());

        *this = buffer.copy(alloc);
        childCount = buffer.size();
    }
};
```
Since now t.node() may be nullptr.
I provide another thought to solve the issue.

PR #638, opened 2022-10-17
Fix null dereference on at SyntaxRewriter remove operation

HungMingWu:
See bug #637


MikePopoloski:
Thanks, this is a much better way of handling this.

PR #644, opened 2022-10-29
Append missing fmt headers on perform "make install"

HungMingWu:


MikePopoloski:
fmt is used as a private dependency of slang, so it's always statically linked and the headers aren't made available (and aren't needed by any public slang headers). Your change here is not sufficient if the user builds with, say, a shared lib because the fmt symbols won't be exported. I think if you want to use fmtlib in your own project you should just add it as a dependency yourself?

HungMingWu:
But `slang/diagnostics/TextDiagnosticClient.h` is public header, it needs `fmt` involved
So I need add decadency my self?

MikePopoloski:
Oh I see, the terminal colors are exposed as fmt enum members. I'll have to think about this some more.

HungMingWu:
I think a alternative solution is keep mapping notation on slang
``` cpp
namespace slang {
    enum class terminal_color : uint8_t {
         // Ignore
    };
}
```
Is it a good idea?

MikePopoloski:
Yeah, it's a good idea. I just need to decide whether it makes sense to make fmtlib a public dependency instead and make sure to install it alongside slang, or whether it's better to keep it private.

MikePopoloski:
I'm going to merge this since I think it's an improvement. There's probably additional work to do in the future if people try to use the fmt headers and call a method that's privately implemented but I'll wait until it becomes a real problem.

PR #649, opened 2022-11-03
Improve parseability of type and constant string representations

paulsc96:
## Overview
Enable `TypePrinter` and various `toString()`s to produce more parseable output by optionally emitting well-formed SystemVerilog expressions in more situations.

All changes below except 3. keep existing stringifications unchanged by default. All existing code remains compatible through optional fields and parameters.

These changes do not guarantee type or constant parseability, but cover some simple-to-fix gaps in Slang's already quite parseable stringifications.

### `TypePrinter`
1. Add the `skipScopedTypeNames` option to `TypePrinter` to omit suffixed type names from container types.
2. Add the `fullEnumType`option to `TypePrinter` to include enum base types.

### `ConstantValue` and `SVInt`
3. Use assignment patterns (`'{...}` ) instead of simple brackets (`[...]`) for container constants.
4. Expose `abbreviateThresholdBits` to the end user of `ConstantValue.toString()` and callees to control or suppress integer vector abbreviation as needed.
5. Add the optional `exactUnknowns` parameter to `ConstantValue.toString()` and callees to ensure a reversible representation of integer vectors with unknown bits.

### DIFF
--- a/include/slang/ast/types/TypePrinter.h
+++ b/include/slang/ast/types/TypePrinter.h
@@ -25,6 +25,7 @@ struct SLANG_EXPORT TypePrintingOptions {
     bool addSingleQuotes = false;
     bool elideScopeNames = false;
     bool printAKA = false;
+    bool skipScopedTypeNames = false;
 
     enum AnonymousTypeStyle { SystemName, FriendlyName } anonymousTypeStyle = SystemName;
 };
--- a/include/slang/ast/types/TypePrinter.h
+++ b/include/slang/ast/types/TypePrinter.h
@@ -26,6 +26,7 @@ struct SLANG_EXPORT TypePrintingOptions {
     bool elideScopeNames = false;
     bool printAKA = false;
     bool skipScopedTypeNames = false;
+    bool fullEnumType = false;
 
     enum AnonymousTypeStyle { SystemName, FriendlyName } anonymousTypeStyle = SystemName;
 };
--- a/source/numeric/ConstantValue.cpp
+++ b/source/numeric/ConstantValue.cpp
@@ -19,14 +19,14 @@ struct always_false : std::false_type {};
 
 const ConstantValue ConstantValue::Invalid;
 
-std::string ConstantValue::toString(bitwidth_t abbreviateThresholdBits) const {
+std::string ConstantValue::toString(bitwidth_t abbreviateThresholdBits, bool exactUnknowns) const {
     return std::visit(
-        [abbreviateThresholdBits](auto&& arg) {
+        [abbreviateThresholdBits, exactUnknowns](auto&& arg) {
             using T = std::decay_t<decltype(arg)>;
             if constexpr (std::is_same_v<T, std::monostate>)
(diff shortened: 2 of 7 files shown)


source/numeric/ConstantValue.cpp
```diff
@@ -37,51 +37,54 @@ std::string ConstantValue::toString() const {
                 return "$"s;
             else if constexpr (std::is_same_v<T, Elements>) {
                 FormatBuffer buffer;
-                buffer.append("[");
+                buffer.append("'{");
                 for (auto& element : arg) {
-                    buffer.append(element.toString());
+                    buffer.append(element.toString(abbreviateThresholdBits));
```
MikePopoloski, line 42: Did you want to pass the exactUnknowns flag down into these other toString() calls as well?

include/slang/ast/types/TypePrinter.h
```diff
@@ -25,6 +25,8 @@ struct SLANG_EXPORT TypePrintingOptions {
     bool addSingleQuotes = false;
     bool elideScopeNames = false;
     bool printAKA = false;
+    bool skipScopedTypeNames = false;
```
MikePopoloski, line 28: Can you make sure these are exposed in the Python bindings as well? Can be easily added in the bindings/python/types.cpp file.

source/numeric/SVInt.cpp
```diff
@@ -637,25 +637,27 @@ std::ostream& operator<<(std::ostream& os, const SVInt& rhs) {
     return os;
 }
 
-std::string SVInt::toString() const {
+std::string SVInt::toString(bitwidth_t abbreviateThresholdBits, bool exactUnknowns) const {
     // guess the base to use
     LiteralBase base;
-    if (bitWidth < 8 || (unknownFlag && bitWidth <= 64))
+    // unless the int is wholly or digit/hexit-wise X or Z (TODO: worth checking?),
```
MikePopoloski, line 643: Probably not worth the effort. I'd just remove the TODO.

MikePopoloski (commented):
Thanks for the diff. I have a few minor comments but happy to land this once you address them.

MikePopoloski:
Also as you can probably tell from the failed tests, there are a bunch of eval tests that need to be updated for your ConstantValue toString change to use `'{}` but I think it's fine to not make that an option and just update the tests instead.

source/numeric/SVInt.cpp
paulsc96: Done

include/slang/ast/types/TypePrinter.h
paulsc96: Done

source/numeric/ConstantValue.cpp
paulsc96: Yes, I did. Thanks for catching that; I propagated it.

paulsc96:
I chose to make 3. optional now and keep the existing default behavior, maintaining full compatibility.

Issue #646, opened 2022-11-02
library name collision when upgrade to slang-v2.0


vowstar:
**Describe the bug**

Before I used the source code of [v1.0](https://github.com/MikePopoloski/slang/releases/tag/v1.0) to compile and install without problems.

Later, when [v2.0](https://github.com/MikePopoloski/slang/releases/tag/v2.0) was released, I tried to compile and install it with the source code of [v2.0](https://github.com/MikePopoloski/slang/releases/tag/v2.0), and found that two paths were generated that conflicted with the existing ``libslang.so`` of the system:

```bash
 * Detected file collision(s):
 *
 * /usr/lib64/libslang.so
 * /usr/lib64/libslang.so.2
```

**To Reproduce**

This problem is easy to reproduce on CentOS 7. For example, you can install any package that depends on ``slang``(not this repo), such as ``newt``, ``libcaca``, ``nettop``, ``util-linux``, etc.

Fastest way to reproduce:

```bash
yum install slang
```

Then CentOS 7 will have:

```bash
/usr/lib64/libslang.so.2
/usr/lib64/libslang.so.2.2.4
```

And then download https://github.com/MikePopoloski/slang/releases/tag/v2.0 source code, compile and install:

```bash
git clone -b v2.0 https://github.com/MikePopoloski/slang.git
mkdir build && cd build && cmake .. && make -j && make install
```
Then the two files will conflict, resulting in corruption.

**Additional context**

Except on CentOS, on other operating systems, library name conflicts will still cause program running errors.

For example debian/ubuntu based OS:

https://packages.debian.org/search?arch=any&mode=filename&searchon=contents&keywords=libslang.so.2

For example:
[AlmaLinux 9](https://pkgs.org/download/libslang.so.2()(64bit)#)
[AlmaLinux 8](https://pkgs.org/download/libslang.so.2()(64bit)#)
[ALT Linux P10](https://pkgs.org/download/libslang.so.2()(64bit)#)
[ALT Linux P9](https://pkgs.org/download/libslang.so.2()(64bit)#)
[ALT Linux Sisyphus](https://pkgs.org/download/libslang.so.2()(64bit)#)
[Amazon Linux 2](https://pkgs.org/download/libslang.so.2()(64bit)#)
[Amazon Linux 1](https://pkgs.org/download/libslang.so.2()(64bit)#)
[CentOS 9 Stream](https://pkgs.org/download/libslang.so.2()(64bit)#)
[...]

PR #650, opened 2022-11-05
Provide cmake override for shared-lib name

jrudess:
As discussed in #646, this PR  provides a solution to the packaging issue, by adding a new variable SLANG_SHARED_LIB_NAME allowing a command line method for renaming the shared-lib during distribution packaging.

Since "s-lang" is in the dependency tree for NetworkManager most distros will have libslang.so installed already.   I'll be using "libsvlang.so" for Arch Linux packaging.

### DIFF
--- a/source/CMakeLists.txt
+++ b/source/CMakeLists.txt
@@ -86,6 +86,7 @@ add_library(
 
 add_subdirectory(ast)
 
+set(SLANG_SHARED_LIB_NAME slang)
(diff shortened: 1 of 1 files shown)


MikePopoloski:
Thanks, libsvlang seems like a reasonable workaround. I'll probably just make that the default but having the option doesn't hurt.

vowstar:
Thanks

PR #656, opened 2022-11-13
Verbose details on system/remote libraries.

cbalint13:
(x ) This PR enhances the following:

1. ~~Enable providing system wide optional (and/or) ```fmt```, ```catch2```, ```pybind11``` libraries.~~
2. ~~Use python module provided standard installation path in the system.~~
3. Verbose detail the process of finding / downloading external dependencies.

(x) This PR preserves and **does not affect current** build behaviour.

---

(x) See the introduced advertisements:

```
{...}
-- CMake version: 3.25.0-rc4
-- slang version: 2.0.0+c04fd52
-- The CXX compiler identification is GNU 12.2.1
{...}
-- Found system fmt library: /usr/lib64/libfmt.so.9.1.0
-- Using system fmt include: /usr/include
{...}
-- Using remote Catch2 library
{...}
-- Using system pybind11 library: 2.10.1
-- Using system pybind11 include: /usr/include;/usr/include/python3.11
{...}
```

(x) Builds pass on *{x86_64, ppc64le, aarch64} x {rhel8, rhel9, fc35, fc36, fc37, fc38}*, see:
https://copr.fedorainfracloud.org/coprs/rezso/HDL/build/5035110/

---

Cc @MikePopoloski

Thanks,
~Cristian.


MikePopoloski:
Hmm, this should not be necessary. FetchContent will already use find_package under the hood first, if FIND_PACKAGE_ARGS is used (which I do set). This requires a relatively new CMake but that's not that hard to satisfy given that there are readily available, dependency-less binary distributions for all major platforms, if your local version is not recent enough.

bindings/CMakeLists.txt
```diff
@@ -44,7 +49,7 @@ target_compile_definitions(pyslang PRIVATE VERSION_INFO=${PROJECT_VERSION})
 install(
   TARGETS pyslang
   COMPONENT pylib
-  DESTINATION .)
+  DESTINATION ${Python_SITEARCH})
```
MikePopoloski, line 52: I'm not sure about this change. The build here on its own is not enough to produce a python package you can place into the Python lib folder, you need the other pieces defined in https://github.com/MikePopoloski/pyslang -- which depends on this installing into a very particular location. Have you actually tested this change and does it work correctly for your use case?

cbalint13:
> Hmm, this should not be necessary. FetchContent will already use find_package under the hood first, if FIND_PACKAGE_ARGS is used (which I do set). This requires a relatively new CMake but that's not that hard to satisfy given that there are readily available, dependency-less binary distributions for all major platforms, if your local version is not recent enough.

* It is not possible to satisfy on more than half (don't have numbers) of rhel / fedora products.
* Same impossibility would be with other distros too (don't know in depths now), pretty sure.

If don't accept it is fine, there are other releng workarounds (not elegant).
I am thinking here on supporting (as compatibility) as much span of platforms as it is possible.

Please confirm YES/NO and i'll drop this part as requested.

bindings/CMakeLists.txt
cbalint13: > I'm not sure about this change. The build here on its own is not enough to produce a python package you can place into the Python lib folder, you need the other pieces defined in https://github.com/MikePopoloski/pyslang -- which depends on this installing into a very particular location. Have you actually tested this change and does it work correctly for your use case?



(x) Let's see a plain ```make install``` test rounds:
 * ```before``` it is putting in something like ```/usr/./pyslang.cpython-311-{...}.so``` (very bad).
 * ```after``` is ```/usr/lib64/python3.11/site-packages/usr/./pyslang.cpython-311-{...}.so``` (the legit one).
 * User can prepend a prefix {/opt, /usr/local} to this ```make install DESTDIR=/opt```

(x) The ```${Python_SITEARCH}``` comes from ```find(python develoment.Module)```.

(x) To my knowledge, the standard python lookups:
   ```*/lib64/*/site-packages``` is the right place for binary (having some ```.so```) inclusion
   ```*/lib/*/site-packages``` is the right place for pure python (noarch) inclusion

(x) Also if you will ever publish it it via ```pip/pypi``` platform you will also need this properly done.

Let me know YES/NO and will drop it, but on this one just enumerated some strong arguments.

cbalint13: Furthermore,

(x) Looking at ```pyslang``` pip:
```
$ pip3 install pyslang
  Created wheel for pyslang: filename=pyslang-2.0.0-cp311-cp311-linux_x86_64.whl size=2938144 
  Successfully installed pyslang-2.0.0
$ ls -l ~/.local/lib/python3.11/site-packages/pyslang
pyslang-2.0.0.dist-info/                 pyslang.cpython-311-x86_64-linux-gnu.so  
```
* Reveals the ```.so``` part reachable  */lib/python3.11/site-packages/pyslang* path.

Now, since ```pyslang``` and ```slang``` are separated (even in different repos) it is still a good to **match legit paths**.
Not sure how ```pip``` build flow finally **just moves** ```.so``` in the proper path, if want we can drop for now this proposal.

cbalint13:
@MikePopoloski ,

(x) I've repurposed this PR to **just print details** about find (first trial) download (final step).
(x) Dropped touching python install path (will see in time) and all other previous conditional switches.

(x) I think this way is usefull too for relengs, see sample details:

```
{...}
-- CMake version: 3.25.0-rc4
-- slang version: 2.0.0+c04fd52
-- The CXX compiler identification is GNU 12.2.1
{...}
-- Found system fmt library: /usr/lib64/libfmt.so.9.1.0
-- Using system fmt include: /usr/include
{...}
-- Using remote Catch2 library
{...}
-- Using system pybind11 library: 2.10.1
-- Using system pybind11 include: /usr/include;/usr/include/python3.11
{...}
```

Can re-review ?

Thanks.

MikePopoloski:
Sure, this seems fine. 

Regarding the Python path, it's not recommended that you include the pylib in a typical configuration. That target is really only intended to be used when building the Python package, which you need the scikit-build pieces in the other repo. The package is already distributed on pypi so probably no need for you to include it in your distro?

For the cmake changes, are you saying that an updated cmake version can't be used when building RHEL 8 packages? As mentioned cmake has readily available binary distributions (there's even a cmake version available on PyPI) so I guess it's more of a policy issue that disallows using extra tools during package build? If so I can allow your proposed changes to make your life easier, but we should plan on it being temporary -- once systems can advance their minimum supported cmake version to 3.24 we can remove the hack.

cbalint13:
> Sure, this seems fine.

* This is very useful, thank you much again for the patience !

> Regarding the Python path, it's not recommended that you include the pylib in a typical configuration. That target is really only intended to be used when building the Python package, which you need the scikit-build pieces in the other repo. The package is already distributed on pypi so probably no need for you to include it in your distro?

*Let's drop this for now.*

* R: My idea was a seamless "make install" using only  ```slang``` github repo, without any other additional ```pyslang``` package.
 
> For the cmake changes, are you saying that an updated cmake version can't be used when building RHEL 8 packages? 

*Let's drop this for now.*

* R: Didn't checked it yet, anyway rhel8, rhel9 suffer of having old native ```fmt``` & ```pybind```.

As mentioned cmake has readily available binary distributions (there's even a cmake version available on PyPI) so I guess it's more of a policy issue that disallows using extra tools during package build? If so I can allow your proposed changes to make your life easier, but we should plan on it being temporary -- once systems can advance their minimum supported cmake version to 3.24 we can remove the hack.

* R: Yes, cmake can be upgraded, but proper releng (emitting native .rpm) forbid any ABI/API breakage on package dependencies level, so can't just distribute a newer "cmake" within a repository with that ease. There are ways to do like "z-streams" (rhel/fedora only) or redundant "cmake-compat" module but really not worth so much effort.


Thank you again !
~Cristian.

PR #660, opened 2022-11-19
Add --obfuscate-ids option to driver

Sustrak:
With this option the preprocessed output will have all of its identifiers randomized.

It might come useful in order to share SV code without having to use a commercial encryption tool.

### DIFF
--- a/source/driver/Driver.cpp
+++ b/source/driver/Driver.cpp
@@ -13,6 +13,7 @@
 #include <fmt/color.h>
 #include <functional>
 #include <random>
+#include <unordered_map>
 
 #include "slang/ast/symbols/CompilationUnitSymbols.h"
 #include "slang/ast/symbols/InstanceSymbols.h"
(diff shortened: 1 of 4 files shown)


source/driver/Driver.cpp
```diff
@@ -401,8 +427,26 @@ bool Driver::runPreprocessor(bool includeComments, bool includeDirectives) {
...
+    std::unordered_map<std::string, std::string> translations;
+
     while (true) {
         Token token = preprocessor.next();
+        if (fuzzIds && token.kind == TokenKind::Identifier) {
+            std::string name = std::string(token.valueText());
+            auto translation = translations.find(name);
+            if (translation == translations.end()) {
+                auto new_name = generateRandomAlphanumericString();
+                auto ret = translations.insert({name, new_name});
```
MikePopoloski, line 439: I don't see how this insert could fail, you only enter this branch if the name is not in the map already.

MikePopoloski (commented):
Hello, thanks for the PR. It looks like it fails to build under clang due to a missing header; can you please fix that?

Also as a general comment I would prefer if this were named "obfuscate" instead of "fuzz". The latter term is overloaded already.

Thanks again!

source/driver/Driver.cpp
Sustrak: True, removed the check. Only putted there in order to completely use the result of the insert function.

PR #661, opened 2022-11-20
Fix library name in sv-lang.pc.in

nickg:
Not sure if the .pc file should also be named `svlang.pc.in` to match the library rename.


MikePopoloski:
Yes, thanks for fixing.

PR #665, opened 2022-11-23
Enhance ReplicatedAssignmentPatternExpression

HungMingWu:
I need the following semantics
```
byte data[4] = '{4{1}}; 
```
ReplicatedAssignmentPatternExpression would expanded into
```
{1, 1, 1, 1} 
```
but I want to keep the original semantics.


MikePopoloski:
I think it's better to only actually store the direct list of expressions and then evaluate them in a loop when needed, instead of binding the expressions over and over, so I did that in 0f5933dc511dac83ace48822ab758cb2e42e65e3 which should make your PR unnecessary.

PR #669, opened 2022-11-25
Update Parser_expressions.cpp

godblesszhouzhou:
For 'always@( *)' lexed as '@' '(' '*,' , there will be a error for the parser. The commit can correct the error.

### DIFF
--- a/source/parsing/Parser_expressions.cpp
+++ b/source/parsing/Parser_expressions.cpp
@@ -1135,6 +1135,13 @@ TimingControlSyntax* Parser::parseTimingControl() {
                                                              expect(TokenKind::CloseParenthesis));
                     }
 
+                    // Special case since @(*) will be lexed as '@' '(' '*)'
+                    if (peek(TokenKind::StarCloseParenthesis)) {
+                        auto startclosePara = consume();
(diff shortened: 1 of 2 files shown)


source/parsing/Parser_expressions.cpp
```diff
@@ -1135,6 +1135,13 @@ TimingControlSyntax* Parser::parseTimingControl() {
                                                              expect(TokenKind::CloseParenthesis));
                     }
 
+                    // Special case since @(*) will be lexed as '@' '(' '*)'
+                    if (peek(TokenKind::StarCloseParenthesis)) {
+                        auto startclosePara = consume();
```
MikePopoloski, line 1140: Typo here, should be something like `starCloseParen` not "start"

MikePopoloski (commented):
Thanks for the PR. The logic looks good to me, just one small fix to make.

MikePopoloski:
Also it would be good to add a test that exercises this. Probably something simple in ExpressionParsingTests.cpp

godblesszhouzhou:
Thanks for suggestions. I have corrected the typo and added a test.

PR #687, opened 2022-12-01
Improve pybind arg

Kuree:
See https://github.com/MikePopoloski/slang/discussions/686

Also fixed one function definition: `SystemSubroutinePublicist::unevaluatedContext` should be `def_static`.


MikePopoloski:
Thanks!

PR #695, opened 2022-12-04
Fix implicit function call issue

HungMingWu:


MikePopoloski:
This does not seem like the right fix. For one thing, all of the tools I tested your example on resolve the name `func` to the return value var instead of the function, so this diff would make slang do the wrong thing even for your test case. For another, checking for a semicolon to trigger the new logic means this only works with very particular orderings of expressions, which doesn't make sense to me.

HungMingWu:
All right,  my thought is too riugh, it is hard to do it right due to the ambiguous syntax.

PR #715, opened 2022-12-27
New tool: slang-tidy

Sustrak:
Adding a new tool, slang-tidy to collect different lint rules that are not strictly related to the language, and so don't make much sense to have them in the slang library.


tools/tidy/tests/OnlyAssignedOnResetTest.cpp
```diff
@@ -0,0 +1,219 @@
...
+    logic [2:0] data;
+    logic [4:0][2:0] fifo;
+
+    always_ff @(posedge clk or negedge rst_n) begin
+        if (~rst_n) begin
+            fifo <= '0;
+        end
+        else if (push) begin
+            for (int i = 0; i <= 4; i++) begin
+                // if (ptr == i) fifo[i] <= data;
```
MikePopoloski, line 136: And here

tools/tidy/tidy.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+    if (showHelp) {
+        slang::OS::print(
+            fmt::format("{}", driver.cmdLine.getHelpText("slang SystemVerilog linter")));
+        return 0;
+    }
+
+    if (showVersion) {
+        OS::print(
+            fmt::format("slang-tidy version {}.{}.{}+{}\nCompiled with slang version {}.{}.{}+{}\n",
```
MikePopoloski, line 34: Probably easiest to just use the slang version for slang-tidy as well. Hardcoding the version number here isn't that useful.

tools/tidy/tidy.cpp
```diff
@@ -0,0 +1,84 @@
...
+        return 1;
+    }
+
+    Registry checkRegistry;
+    DiagnosticEngine diagEngine(*compilation->getSourceManager());
+    auto textDiagClient = std::make_shared<TextDiagnosticClient>();
+    textDiagClient->showColors(true);
+    diagEngine.addClient(textDiagClient);
+
+    for (const auto& check_name : checkRegistry.get_registered()) {
```
MikePopoloski, line 66: I don't think we'll want all checks that exist to be run every time. Probably the user wants to control which ones, maybe just using the normal -W warning flag interface. Maybe the registry should be keyed on the DiagCode that controls the check, or alternatively we add a new command line flag with a name for each check (clang-tiday uses a --check flag).

tools/tidy/tests/OnlyAssignedOnResetTest.cpp
```diff
@@ -0,0 +1,219 @@
+#include "../synthesis/OnlyAssignedOnReset.cpp"
```
MikePopoloski, line 1: Should probably have an easy way to write tests against a particular check without needing to cross include the source file.

MikePopoloski (commented):
Thanks a lot for the diff, this is a good start. In order to pass lint checks all files will need to have a comment header added, see other files for examples, and then the pre-commit checks and auto-formatting need to be run too.

tools/tidy/include/ASTHelperVisitors.h
```diff
@@ -0,0 +1,53 @@
...
+        if (auto* symbol = expression.getSymbolReference(); symbol) {
+            identifiers.push_back(symbol->name);
+        }
+    }
+    std::vector<std::string_view> identifiers;
+};
+
+/// ASTVisitor that will try to find the provided name in the LHS of an assignment
+struct LookupLhsIdentifier : public slang::ast::ASTVisitor<LookupLhsIdentifier, true, true> {
+    LookupLhsIdentifier() = delete;
```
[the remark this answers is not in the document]
Sustrak: I want  to enforce that all instances of `LookupLhsIdentifier` have a `name` set, so the visitor works as intended.

scripts/diagnostics.txt
```diff
@@ -972,6 +972,9 @@ subsystem Meta
 error TooManyErrors "too many errors emitted, stopping now [--error-limit=]"
 warning unknown-warning-option UnknownWarningOption "unknown warning option '{}'"
 
+subsystem Tidy
```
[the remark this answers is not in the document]
Sustrak: Do you mean to have a `diagnostics.txt` in the slang-tidy, or have a whole diagnostic system for the slang-tidy tool?

tools/tidy/tidy.cpp
Sustrak: Agree, also `slang-tidy` could read from a `.slang-tidy` file as `clang-tidy` does in order to select which checks to enable.

tools/tidy/tests/OnlyAssignedOnResetTest.cpp
Sustrak: Not an elegant solution, but how can I do that without having a header for each check?

scripts/diagnostics.txt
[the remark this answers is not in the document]
Sustrak: The fact that those new diagnostics are not checked using slang, but with slang-tidy, makes the docs CI to fail. Since it tries to trigger the warning with the diagnostics docs example using slang.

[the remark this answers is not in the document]
MikePopoloski: Yeah, that's part of why I think it should be a separate file (like diagnostics.txt), but I'm not entirely sure what it should look like. Perhaps analysis "warnings" in slang-tidy don't need to be in a separate file at all but could just put their message inline with the source code.

tools/tidy/include/ASTHelperVisitors.h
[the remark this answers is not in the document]
MikePopoloski: Sure, I'm just saying that providing a user-defined constructor means that there *is* no default constructor anyway, so this line actually does nothing.

tools/tidy/tests/OnlyAssignedOnResetTest.cpp
MikePopoloski: Perhaps you can enable the check using the registry and its string name? You don't actually need to see the class for the check, you just need to turn it on.

scripts/diagnostics.txt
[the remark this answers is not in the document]
Sustrak: I have moved the diagnostics to be inline in the visitor code, and removed them from the `diagnostics.txt` file.

MikePopoloski:
Apologies for the delay on this, I want to do another review pass but I've been away for the holidays.

MikePopoloski:
OK, I'm going to merge this as-is since I think it's in fine shape. Probably lots of things to improve on it as well but we can do that in followup commits.

PR #717, opened 2023-01-09
Backport slang to clang9

HungMingWu:
Since the minimum gcc version is 9, its release date is May 3, 2019
I hope clang 9 should support either. (19 September 2019)


MikePopoloski:
This is fine for now, but be aware that I'm going to switch to gcc 10 relatively soon so that I can start using C++20 functionality, which will likely not work with clang9.

PR #729, opened 2023-03-12
fix addOne/subOne for SVInt

adream307:
fix addOne/subOne for SVInt


MikePopoloski:
Nice, thanks for finding this.

PR #734, opened 2023-03-29
Remove reference to old std span compatibility header

daglem:


MikePopoloski:
Nice, thanks for fixing.

PR #749, opened 2023-05-05
Expose Symbol::getDeclaredType in python bindings

tdp2110:


bindings/python/symbols.cpp
```diff
@@ -101,6 +101,7 @@ void registerSymbols(py::module_& m) {
         .def_property_readonly("isScope", &Symbol::isScope)
         .def_property_readonly("isType", &Symbol::isType)
         .def_property_readonly("isValue", &Symbol::isValue)
+        .def_property_readonly("getDeclaredType", &Symbol::getDeclaredType)
```
MikePopoloski, line 104: Since this is a property it should be named as a noun not a verb, i.e. `declaredType`

PR #752, opened 2023-05-08
search for interface files when using -y

AndrewNolte:


include/slang/parsing/ParserMetadata.h
```diff
@@ -44,6 +44,8 @@ struct SLANG_EXPORT ParserMetadata {
     /// A list of all class declarations parsed.
     std::vector<const syntax::ClassDeclarationSyntax*> classDecls;
 
+    std::vector<const syntax::InterfacePortHeaderSyntax*> interfacePorts;
```
MikePopoloski, line 47: Please add a simple doc string so this shows up in the reference docs.

source/driver/Driver.cpp
```diff
@@ -593,7 +593,7 @@ bool Driver::parseAllSources() {
 
         auto findMissingNames = [&](const std::shared_ptr<SyntaxTree>& tree,
                                     flat_hash_set<std::string_view>& missing) {
-            auto& meta = tree->getMetadata();
+            const auto& meta = ParserMetadata::fromSyntax(tree->root());
```
MikePopoloski, line 596: You should not recreate the metadata here, it's an expensive operation. The parser already builds up the metadata incrementally, you should modify the parser to also include your new field.

source/driver/Driver.cpp
```diff
@@ -612,6 +612,12 @@ bool Driver::parseAllSources() {
                         missing.emplace(name);
                 }
             }
+
+            for (auto bd : meta.interfacePorts) {
```
MikePopoloski, line 616: What does 'bd' stand for here? Can you either use 'ip' for interface port or use something more descriptive like 'portSyntax' or whatever.

source/parsing/Parser.cpp
```diff
@@ -82,7 +82,23 @@ AnsiPortListSyntax& Parser::parseAnsiPortList(Token openParen) {
...
+
+        if (ansiport->header->kind != SyntaxKind::InterfacePortHeader)
+            continue;
+        InterfacePortHeaderSyntax* goodheader = &ansiport->header->as<InterfacePortHeaderSyntax>();
+
+        meta.interfacePorts.push_back(goodheader);
+    }
+
+    return *portList;
 }
```
AndrewNolte, line 102: This seems like a lot for the parser, should I maybe put this logic in the driver and use the nodemap?

MikePopoloski: Why not just add to the interfacePorts list from the parsePortHeader function in the branch that creates InterfacePortHeaders directly so that you don't need this second pass through the list of ports?

source/driver/Driver.cpp
```diff
@@ -612,6 +612,12 @@ bool Driver::parseAllSources() {
...
+                std::string_view name = bd->nameOrKeyword.valueText();
+                if (knownNames.find(name) == knownNames.end())
```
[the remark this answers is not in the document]
AndrewNolte: I took out the interface keyword check bc of the Parser change. I don't think the name will every be empty right?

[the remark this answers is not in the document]
MikePopoloski: I think you still need those checks because of the reconstruction case in ParserMetadata.cpp, which will apply to all InterfacePortHeaders and not just the specific code path in parsePortHeader. That check can either be here or in ParserMetadata.cpp.

Issue #765, opened 2023-05-26
RFC: expose visitors to python


tdp2110:
**Is your feature request related to a problem? Please describe.**
Expose visitor pattern to python for richness of pybindings.

**Describe the solution you'd like**
Given a `pyslang.Expression`, `pyslang.Statement`, or `pyslang.Symbol`, `t`, and a python object `f`, 
expose a function
```
  t.visit(f)
```
to recursively visit `t`. To emulate the `handle` pattern used in C++, something closer to the [MLIR walk pattern](https://github.com/MikePopoloski/slang/compare/master...tdp2110:slang:py-visitors?expand=1) could be used, for example:

```
def visit_fn(n):
   if isinstance(n, pyslang.NamedValueExpression):
    # "handle" a NVE
   if isinstance(n, pyslang.BinaryExpression):
     # "handle" a binary Exp
  if should_continue:
    return pyslang.Continue # continue traversing in pre-order
  return pyslang.Skip # skip children

v.visit(visit_fn)
```

A basic sample implementation is available [here](https://github.com/MikePopoloski/slang/compare/master...tdp2110:slang:py-visitors?expand=1), which doesn't fully implement the continue/skip pattern as above -- only using the noneness of the return value.

**Describe alternatives you've considered**
Re-implement the visitor pattern in pure python (error prone and likely incomplete).

I would find this useful. If you like the idea, I'd be happy to flush it out and submit a PR.

MikePopoloski:
Sure, this makes sense to me. Happy to take a PR.

PR #771, opened 2023-05-29
Python bindings: expose a visitor pattern

tdp2110:
This addresses #765. Commit messages are reproduced below:

1af24651 
pyslang.h: correct file reference in leading comment

7d650283
Python bindings: expose a visitor pattern

Expose visit methods on pyslang Expressions, Statements, and Symbols.
This method is modelled after MLIR's visitor pattern wherein
callbacks can signal certain actions on the walk, such as
interrupting (i.e., stopping entirely), skipping child nodes,
or to advance.

This commit includes no tests, as pyslang tests are not included
in the slang repo, but in the pyslang repo. Once this commit
is visible to pyslang, tests will be added there.

Here is an example test case that could be added, modelled off
slang/tests/unittests/VisitorTests.cpp:

```python
    from pyslang import *

    tree = SyntaxTree.fromText(
      """
    module m;
      int j;
      initial begin : asdf
        j = j + 3;
        if (1) begin : baz
          static int i;
          i = i + 2;
          if (1) begin : boz
            i = i + 4
          end
        end
      end
    endmodule
    """)

    class Visitor:
      def __init__(self):
        self.stmt_ct = 0
      def visit(self, node):
        if isinstance(node, Statement):
          self.stmt_ct += 1

    c = Compilation()
    c.addSyntaxTree(tree)
    v = Visitor()
    c.getRoot().visit(v.visit)
    assert v.stmt_ct == 11
```


MikePopoloski:
Nice, looks good.

PR #776, opened 2023-06-22
Add new features to slang-tidy

Sustrak:
This PR adds new features to slang-tidy:

- slang-tidy can read its configuration from a .slang-tidy file, closely similar to what .clang-tidy offers.
- Adds new checks to the linter.
- Small refactor in the code of the checks.
- Add option to skip linting checks from specific files.

This PR also adds a new tool: `include/slang/util/CppTypePrinter.h`


MikePopoloski:
The shared lib builds require that symbols be exported / imported correctly across the shared lib boundary (with the use of the SLANG_EXPORT macro decoration). It's not immediately obvious to me why those particular symbols are complaining though. I can look more though.

PR #777, opened 2023-06-25
Add new features to slang-tidy

Sustrak:
This PR adds new features to slang-tidy:

- slang-tidy can read its configuration from a .slang-tidy file, closely similar to what .clang-tidy offers.
- Adds new checks to the linter.
- Small refactor in the code of the checks.
- Add option to skip linting checks from specific files.

This PR also adds a new tool: `include/slang/util/CppTypePrinter.h`

### DIFF
--- a/tools/tidy/src/style/NoOldAlwaysSyntax.cpp
+++ b/tools/tidy/src/style/NoOldAlwaysSyntax.cpp
@@ -1,0 +1,58 @@
+//! @file NoOldAlwaysSyntax.h
+//! @brief Checks for use of old verilog always syntax
+//
+// SPDX-FileCopyrightText: Michael Popoloski
+// SPDX-License-Identifier: MIT
+//------------------------------------------------------------------------------
+#include "ASTHelperVisitors.h"
+#include "TidyDiags.h"
+
+#include "slang/syntax/AllSyntax.h"
+
+using namespace slang;
+using namespace slang::ast;
+
+namespace no_old_always_syntax {
+struct MainVisitor : public TidyVisitor, ASTVisitor<MainVisitor, true, true> {
+    explicit MainVisitor(Diagnostics& diagnostics) : TidyVisitor(diagnostics) {}
+
+    void handle(const ast::ProceduralBlockSymbol& symbol) {
+        NEEDS_SKIP_SYMBOL(symbol)
+
+        if (symbol.procedureKind == ProceduralBlockKind::Always) {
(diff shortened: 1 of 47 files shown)


MikePopoloski:
You shouldn't need the SLANG_EXPORT tag on anything in the tidy binary, it's only for library symbols. I suspect the problem might be from not_null being tagged even though it's a template class. I'm going to experiment with removing it and seeing if that helps.

Sustrak:
Removed, thanks for taking a look at it

MikePopoloski:
Ok, I committed a potential fix, try rebasing.

Sustrak:
It did the trick, thanks. Ready to be merged if you don't see anything sketchy on the code

tools/tidy/src/style/NoOldAlwaysSyntax.cpp
```diff
@@ -0,0 +1,58 @@
+//! @file NoOldAlwaysSyntax.h
```
MikePopoloski, line 1: Looks like some of these files are missing the top comment line, and also have the wrong file name listed. You don't need the doxygen @file style comments here either for source files since nothing in them needs to be documented.

tools/tidy/tests/CMakeLists.txt
```diff
@@ -5,9 +5,26 @@
 
 add_executable(
   tidy_unittests
-  ../../../tests/unittests/main.cpp ../../../tests/unittests/Test.cpp
-  ../synthesis/OnlyAssignedOnReset.cpp OnlyAssignedOnResetTest.cpp
-  ../synthesis/RegisterHasNoReset.cpp RegisterHasNoResetTest.cpp)
+  ../../../tests/unittests/main.cpp
+  ../../../tests/unittests/Test.cpp
+  ../src/ASTHelperVisitors.cpp
```
MikePopoloski, line 10: It's probably worth compiling these into an object library at least, so that you aren't compiling each file twice in the build (once for the tool and once for the tests).

Sustrak: If you are referring to the checks `cpp` files, I'm not sure I can do that since then the static variable that registers the checks gets optimized out by the linker. I'm talking about the `REGISTER` macro I use to register the different checks into the factory

MikePopoloski: I think if you use an OBJECT library that shouldn't happen? It's just a collection of compiled object files that get pulled into each build.

PR #787, opened 2023-07-10
Support multiple slang-tidy port suffixes for EnforcePortSuffix check

VJSchneid:
Hi folks,
this PR makes the following modifications in slang-tidy:
 * Adds support to provide an array of configuration values in the slang-tidy config file (``[_i, _ni]``).
 * Multiple (or none) port suffixes can be specified now. This enables e.g. negated ports to be specified as ``_ni`` when both ``_i`` and ``_ni`` are configured as port suffixes.
 * Parsing of configValues was slightly rewritten to allow flexible parsing depending on targeted type.


MikePopoloski:
Thanks for the PR. I skimmed this and it looks ok to me so I'm going to merge. @Sustrak should reply as well if he thinks there are things worth changing here.

PR #801, opened 2023-07-27
New tool: slang-reflect

Sustrak:
`slang-reflect` is a tool that aims to help developers to remove the burden of creating `C/C++` representations of SystemVerilog structs, enums and parameters. Its behavior is inspired from the [verilator public](https://verilator.org/guide/latest/extensions.html#cmdoption-2) option provided by `verilator`, and it is of a great use when building testbenches using `C++/SystemC`.


tools/type_extract/include/CppEmitter.h
```diff
@@ -0,0 +1,101 @@
...
+        includes.emplace_back("ostream");
+        includes.emplace_back("cstddef");
+        includes.emplace_back("cstdint");
+        includes.emplace_back("string");
+        includes.emplace_back("sstream");
+        if (!noSystemC)
+            includes.emplace_back("systemc.h");
+    }
+
+    inline void add(std::string&& code) { hpp << code; }
```
MikePopoloski, line 25: Note that the `inline` keyword for methods that are already defined inline is redundant.

tools/type_extract/include/CppEmitter.h
```diff
@@ -0,0 +1,101 @@
...
+    inline void addWithIndent(std::string&& code) { hpp << indent(currentIndent) << code; }
+    inline void increaseIndent() { currentIndent++; }
+    inline void decreaseIndent() {
+        SLANG_ASSERT(currentIndent != 0);
+        currentIndent--;
+    }
+
+    std::string emit() {
+        std::stringstream ss;
+        ss << "// " << fileName << std::endl;
```
MikePopoloski, line 43: It's fine if you want to keep what you have but the fmt library can make this kind of code a lot more straightforward.

tools/type_extract/include/SvGeneric.h
```diff
@@ -0,0 +1,133 @@
...
+            return alias->second;
+        return typeName;
+    }
+
+    static bool isCppReserved(std::string_view name) {
+        return std::find(cppReservedKeywords.begin(), cppReservedKeywords.end(), name) !=
+               cppReservedKeywords.end();
+    }
+
+    static constexpr std::array<std::string_view, 97> cppReservedKeywords = {"alignas",
```
MikePopoloski, line 36: Might be nice to use CTAD here so you don't have to explicitly count the entries: `static constexpr std::array cppReservedKeyword =`

tools/type_extract/include/SvLocalParam.h
```diff
@@ -0,0 +1,25 @@
...
+#include "SvType.h"
+#include <fmt/format.h>
+
+class SvLocalParam : public SvGeneric {
+public:
+    explicit SvLocalParam(const slang::ast::ParameterSymbol& parameter) :
+        SvGeneric(SvGeneric::Kind::LocalParam), parameter(parameter) {}
+
+    void toCpp(HppFile& hppFile, std::string_view, const SvAliases&, bool) const override {
+        //** DECLARATION **//
```
MikePopoloski, line 13: Not sure what this comment means.

tools/type_extract/include/SvType.h
```diff
@@ -0,0 +1,33 @@
+#pragma once
+
+#include <ostream>
+
+#include "slang/ast/types/Type.h"
+
+namespace CppType {
+enum Type { BOOL, U32, U64, SC_BV, STRUCT, ENUM };
+
+std::string toString(const Type cppType);
```
MikePopoloski, line 10: Note that the top-level `const` on value parameters does nothing, so it's a little confusing to include it (because it makes me think that maybe you meant it to be a const reference or something).

tools/type_extract/src/TypeExtract.cpp
```diff
@@ -0,0 +1,82 @@
...
+    }
+    SLANG_CATCH(const std::exception& e) {
+#if __cpp_exceptions
+        OS::printE(fmt::format("internal compiler error: {}\n", e.what()));
+#endif
+        return 1;
+    }
+
+    if (!compilation_ok) {
+        OS::print("slang-tidy: errors found during compilation\n");
```
MikePopoloski, line 66: Should update the name "slang-tidy" here

MikePopoloski (commented):
Thanks, this is really nice. As far as the name goes, maybe something with "reflect" in it? That's kind of what you're doing, reflecting / using reflection to get the type information. So like type-reflect or just slang-reflect even. type-extract isn't terrible though if you want to just stick with that.

Also something for a followup diff might be to support other kinds of parameter types like strings, not just integers.

tools/type_extract/include/ASTVisitors.h
```diff
@@ -0,0 +1,39 @@
...
+
+            for (auto& blockComment : blockComments) {
+                isPublic = std::find(publicDirectives.begin(), publicDirectives.end(),
+                                     blockComment.getRawText()) != publicDirectives.end();
+            }
+        }
+    }
+
+    bool operator()() {
+        bool ret = isPublic;
```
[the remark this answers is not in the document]
Sustrak: Neat 👀

tools/type_extract/include/SvType.h
Sustrak: Forgot the `&`

Sustrak:
I like the `slang-reflect` name, shorter and with better meaning over what the tools does 😄

PR #805, opened 2023-08-01
Expose symbol on ParameterSymbolBase in Python bindings

Cyberwizzard:
As discussed in MikePopoloski/pyslang#172, I was looking for a way to get to the parameter symbol from the ParameterSymbolBase class, and I just exposed the `symbol`.

I know you proposed a different solution, but I compiled and tested this and am using this now myself in the meantime so I figured I would just open a pull request.

### DIFF
--- a/bindings/python/SymbolBindings.cpp
+++ b/bindings/python/SymbolBindings.cpp
@@ -211,7 +211,8 @@ void registerSymbols(py::module_& m) {
                                [](const EnumValueSymbol& self) { return self.getValue(); });
 
     py::class_<ParameterSymbolBase>(m, "ParameterSymbolBase")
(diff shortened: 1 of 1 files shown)


MikePopoloski:
Thanks for the PR. It's not wrong, but I think the automatic downcaster is a better experience. I committed that change in 7d37f2591cce5fce149cbbef91f92d3f9ce193b0 so I think this PR becomes unnecessary.

Cyberwizzard:
Sorry for the late reply as I was travelling - I agree, any path to reach the parameter details is appreciated.

@MikePopoloski when do you think you'll release a new tag and a new build to pypi.org? It looks like quite some changes have accumulated between version 3 and master.

MikePopoloski:
Should be soon, I'm working on fixing issues and getting the release in shape.

PR #894, opened 2024-02-14
Reflect more complex local parameters

### LINKED ISSUE #825: Wrong index variant on slang-reflect
The following construct will cause a:
```
terminate called after throwing an instance of 'std::bad_variant_access'
  what():  std::get: wrong index for variant
Aborted (core dumped)
```

the construct:

```systemverilog
localparam [CORE_ID_BITS - 1:0] CORE_ID_0 [pkg::NR_CORES] = '{2, 3, 0, 0, 0}; 
localparam [CORE_ID_BITS - 1:0] CORE_ID_1 [pkg::NR_CORES] = '{4, 5, 0, 0, 0}; 
localparam [CORE_ID_BITS - 1:0] CORE_ID [2][pkg::NR_CORES] /* verilator public */ = '{CORE_ID_0, CORE_ID_1};  
```

Sustrak:
With this commit slang-reflect is able to reflect more complex local parameters. Also, it will throw an error if it doesn't know how to reflect the parameter instead of crashing.

Fixes #825

### DIFF
--- a/tools/reflect/src/SvLocalParam.cpp
+++ b/tools/reflect/src/SvLocalParam.cpp
@@ -1,0 +1,77 @@
+// SPDX-FileCopyrightText: Michael Popoloski
+// SPDX-License-Identifier: MIT
+
+#include "SvLocalParam.h"
+
+void unwrapUnpackedArray(const std::span<const slang::ConstantValue> constantValues,
+                         std::vector<std::vector<uint64_t>>& values, uint64_t& biggestElementSize) {
+    if (constantValues.front().isUnpacked())
+        for (const auto& unpackedArray : constantValues)
+            unwrapUnpackedArray(unpackedArray.elements(), values, biggestElementSize);
+    else if (constantValues.front().isInteger()) {
+        std::vector<uint64_t> collectedValues;
+        for (const auto& value : constantValues) {
+            if (!value.isInteger())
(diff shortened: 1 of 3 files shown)


tools/reflect/src/SvLocalParam.cpp
```diff
@@ -0,0 +1,77 @@
...
+void SvLocalParam::toCpp(HppFile& hppFile, std::string_view, const SvAliases&, bool) const {
+    std::string parameterName = isCppReserved(parameter.name) ? fmt::format("_{}", parameter.name)
+                                                              : std::string(parameter.name);
+    if (parameter.getValue().isUnpacked()) {
+        std::vector<std::vector<uint64_t>> unpackedArrays;
+        uint64_t biggestSize = 0;
+        SLANG_TRY {
+            unwrapUnpackedArray(parameter.getValue().elements(), unpackedArrays, biggestSize);
+        }
+        SLANG_CATCH(const std::runtime_error& error) {
```
MikePopoloski, line 36: The exception object gets removed when exceptions are disabled, so you then can't refer to it in the format() message below. You would need to put that call between a `#if __cpp_exceptions` so that the no-exception build will pass.

Sustrak: Right, done so in the last commit :+1:

PR #915, opened 2024-03-14
[slang] Fix virtual interface comparison

likeamahoney:
`systemverilog` LRM (for example IEEE 1800-2017 or IEEE 1800-2023) 25.9 clause says:

> — Equality ( == ) and inequality ( != ) with the following:
• Another virtual interface of the same type
• An interface instance of the same type
• The special constant null

It follows that virtual interfaces can be compared with instances of the same type which is not supported by `slang` for now. 

I fixed this at that PR.


MikePopoloski:
Thanks for the PR!

PR #928, opened 2024-04-01
SVInt operator< fix.

Krym4s:
Hello, SVInt compares incorrectly with minimum signed int value. These test cases are failed on current main branch
```
CHECK(SVInt(32, -2147483648, 1) < SVInt(32, -1, 1));
CHECK(SVInt(32, -1, 1) > SVInt(32, -2147483648, 1));
```

It is because of negation of minimum signed int value is impossible (INT_MIN = -INT_MAX-1). This fact is used [here](https://github.com/MikePopoloski/slang/blob/3c5e5e0106f53a505c8b23d0db4e23a2ffa58a6b/source/numeric/SVInt.cpp#L1402). It is better to compare bit inversions of negative values as 
```
-var = ~var + 1
``` 
so every value can be represented.

This patch also includes function for representing minimum and maximum integer values.

### DIFF
--- a/include/slang/numeric/SVInt.h
+++ b/include/slang/numeric/SVInt.h
@@ -684,6 +684,14 @@ inline SLANG_EXPORT uint32_t clog2(const SVInt& v) {
     return v.getBitWidth() - (v - SVInt::One).countLeadingZeros();
 }
 
+// Returns maximum value of signed or unsugned int
+SVInt getMinValue(const SVInt& Int);
+SVInt getMinValue(bitwidth_t bitwidth, bool isSigned);
+
+// Returns minimum value of signed or unsugned int
+SVInt getMaxValue(const SVInt& Int);
+SVInt getMaxValue(bitwidth_t bitwidth, bool isSigned);
+
 inline namespace literals {
 
 inline SVInt operator""_si(const char* str, size_t size) {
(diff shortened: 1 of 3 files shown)


Krym4s:
Added new commit with shared library build fix. Also added some tests.

MikePopoloski:
Thanks for the PR; I definitely should have had test coverage of this case. I think the extra codepath here is unnecessary though; if both sides are negative then we can fall through to the normal unsigned handling case, since negative numbers will compare correctly when both are negative and treated as unsigned. In fact I can probably clean up these comparison functions a bit more to make them more efficient anyway.

I'm not entirely convinced of the need for the getMinValue / getMaxValue functions added here. They seem unrelated to the bug? At the very least the overloads that accept another SVInt are confusing, since the value of that parameter doesn't contribute to the result.

Krym4s:
Thank you for your answer. Yes, getMinValue and getMaxValue are not related to this bug, so it is better to remove it from this PR, but I think that it is useful methods, so maybe it is possible to add them later. 

In new commit I handled case when only one of two signed numbers is negative. Also I added tests for this case. In case two values are negative we fall through to the normal unsigned handling case.

MikePopoloski:
Thanks, lgtm

Issue #927, opened 2024-04-01
slang allows Verilog code to redefine `define's from the command line


udif:
**Describe the bug**
While not part of the spec, Most of the tools (or at least the 2 that I have checked), treat preprocessor defines that were defined by the command line args (-define or `+define+<NAME>=<val>`) to be "stronger" than those defined within Verilog files.

**To Reproduce**
```Verilog
`define N 1
module t;
generate
  $info("%d", `N);
endgenerate
endmodule
```
Call this with:

`slang +define+N=2 t.v`

And you would get:
```
Top level design units:
    t

t2.v:4:3: note: $info encountered:           1
  $info("%d", `N);
  ^

Build succeeded: 0 errors, 0 warnings
```

Run the equivalent code on one tool and you would get:
```
...
	Elaborating the design hierarchy:
	Top level design units:
		t
          2 
  $info("%d", `N);
...
```
Another tool yields:
```
# KERNEL: Info: testbench.sv (6):           2
```

**Additional context**
You may consider this behavior to be optional (using a new flag), if you think changing this would break past usage of slang, in the same way you do with `--single-unit`, which as I have just noticed,  conforms the IEEE spec to the letter (3.12.1):

_The exact mechanism for defining which files constitute a compilation unit is tool-specific. However, compliant tools shall provide use models that allow **both** of the following cases:_
_a) All files on a given compilation command line make a single compilation unit (in which case the declarations within those files are accessible following normal visibility rules throughout the entire set of files)._
_b) Each file is a separate compilation unit (in which case the declarations in each compilation-unit scope are accessible only within its 
 corresponding file)._

udif:
Two more data points (from my tests with these 2 simulators):
1. If your code undefines the macro in your verilog file, it **will** get undefined, regardless of its origin (command line or within your code).
2. If you redefine it again after undefining it, the new value overrides the command line value (e.g. once it is undefined, it is really gone).

MikePopoloski:
All four tools I tried also exhibit the same behavior, so it makes sense for slang to just have the same behavior as well (not gated behind a flag).

PR #929, opened 2024-04-02
PR: Make command-line defines stronger than `define's inside Verilog files

udif:
Here is a simple PR that solves #927 , although it doesn't add a command line flag to keep the older behavior optionally.
I haven't added more unit tests yet, but the existing unittests all pass.
Here are 3 manual cases I've tried, all ran with `slang +define+N=2 t.v` :

``` Verilog
`define N 1
module t;
generate
  $info("%d", `N);
endgenerate
endmodule
```
This one yields 2, as the command line flag overrides the `define` inside the file.

```Verilog
`define N 1
`undef N
module t;
generate
  $info("%d", `N);
endgenerate
endmodule
```
This one yields an undefined macro error, since the `undef` inside the file removed the command-line `define`.

```Verilog
`define N 1
`undef N
`define N 1
module t;
generate
  $info("%d", `N);
endgenerate
endmodule
```
This one yields 1 because once `N` is undefined, it can be redefined inside the file.

### DIFF
--- a/source/parsing/Preprocessor.cpp
+++ b/source/parsing/Preprocessor.cpp
@@ -643,7 +643,7 @@ Trivia Preprocessor::handleDefineDirective(Token directive) {
             bad = true;
         }
         else if (it->second.commandLine)
-            bad = true;
+            bad = true; // not really bad, but commandLine args has precedence so we skip this
         else if (!bad && it->second.valid() && !isSameMacro(*result, *it->second.syntax)) {
             auto& diag = addDiag(diag::RedefiningMacro, name.range());
             diag << name.valueText();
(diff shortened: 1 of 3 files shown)


source/parsing/Preprocessor.cpp
```diff
@@ -640,6 +642,8 @@ Trivia Preprocessor::handleDefineDirective(Token directive) {
             addDiag(diag::InvalidMacroName, name.range());
             bad = true;
         }
+        else if (it->second.commandLine)
```
MikePopoloski, line 645: Can you add a comment here that it's not really "bad", it's just that the command line value takes precedence over the one found in the source?

MikePopoloski (commented):
This approach looks good to me, just need to add the test cases in your summary to PreprocessorTests.cpp

source/parsing/Preprocessor.cpp
udif: Done.

MikePopoloski:
All syntax node types are defined by a generated file called slang/syntax/AllSyntax.h.

PR #931, opened 2024-04-02
Don't error if multiple UDP row inputs match, unless their output differ

### LINKED ISSUE #926: primitive elaboration too strict
**Describe the bug**
When multiple table rows in a primitive overlap but don't contradict, slang issues an error.

**To Reproduce**
The test below was extracted from a library:
```
 primitive D (q, clk, d, _c, _s);
          output reg q;
          input clk, d, _c, _s;
          table
          // clk in  _c  _s  : Qt  : Qt+1                             
            r    0    1   1  : ?   :  0;
            *    0    ?   1  : 0   :  -;
          endtable
 endprimitive
```
When run, I get:
```
t.v:7:13: error: primitive table row duplicates a set of inputs with a different specified output value
            *    0    ?   1  : 0   :  -;
            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
t.v:6:13: note: previous definition here
            r    0    1   1  : ?   :  0;
            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
**Additional context**

Section 29.3.4 of IEEE1800-2017 says:

`It shall be illegal to have the same combination of inputs, including edges, specify different output values.`

It doesn't say inputs can't overlap (and when using wildcards it is hard to avoid).  
In the case above, the inputs partially overlap, but don't contradict where they do.  
Regardless of whether this should be reported or now, I think it should be a warning in the worst case.

udif:
This fixes #926 .
It takes into account the 'keep previous state' output value when comparing UDP table rows.
Also includes a new unit test.

### DIFF
--- a/source/ast/symbols/MemberSymbols.cpp
+++ b/source/ast/symbols/MemberSymbols.cpp
@@ -1015,9 +1015,12 @@ static void createTableRow(const Scope& scope, const UdpEntrySyntax& syntax,
         return 0;
     };
 
-    auto stateChar = getStateChar(syntax.current);
-    if (!stateChar)
-        return;
+    char stateChar = 0;
+    if (syntax.current) {
+        stateChar = getStateChar(syntax.current);
+        if (!stateChar)
+            return;
+    }
 
     auto getOutputChar = [](const UdpFieldBaseSyntax* base) -> char {
         if (base && base->kind == SyntaxKind::UdpSimpleField) {
--- a/source/ast/symbols/MemberSymbols.cpp
+++ b/source/ast/symbols/MemberSymbols.cpp
@@ -1042,6 +1042,23 @@ static void createTableRow(const Scope& scope, const UdpEntrySyntax& syntax,
     if (!outputChar)
         return;
 
+    auto matchOutput = [](char state1, char output1, char output2) -> bool {
+        if (output1 != '-')
+            return false;
+        switch (state1) {
+            case '0':
+            case '1':
+            case 'x':
+                return (output2 == state1);
+            case 'b':
+                return (output2 == '0' || output2 == '1');
+            case '?':
+                return (output2 == '0' || output2 == '1' || output2 == 'x');
+            default:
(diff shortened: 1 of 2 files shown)


source/ast/symbols/MemberSymbols.cpp
```diff
@@ -995,10 +995,9 @@ static void createTableRow(const Scope& scope, const UdpEntrySyntax& syntax,
         return;
     }
 
-    char stateChar = 0;
-    if (syntax.current) {
-        if (syntax.current->kind == SyntaxKind::UdpSimpleField) {
-            auto raw = syntax.current->as<UdpSimpleFieldSyntax>().field.rawText();
+    auto getstateChar = [](const UdpFieldBaseSyntax* base) -> char {
```
MikePopoloski, line 998: `getstateChar` should be `getStateChar`

source/ast/symbols/MemberSymbols.cpp
```diff
@@ -1007,17 +1006,18 @@ static void createTableRow(const Scope& scope, const UdpEntrySyntax& syntax,
...
                 }
             }
         }
+        return 0;
+    };
 
-        if (!stateChar)
-            return;
-    }
+    auto stateChar = getstateChar(syntax.current);
```
MikePopoloski, line 1018: This is wrong now; not all UDPs have a state character, so I believe this breaks checking for combinational UDPs. I guess that means we're missing a test that fails for dup rows in a combinational UDP.

source/ast/symbols/MemberSymbols.cpp
udif: Fixed. Somehow this got through.

source/ast/symbols/MemberSymbols.cpp
udif: You are right. I accidently dropped the check on `syntax.current`.
I added a new test for conflicting outputs on a combinatorial UDP, and indeed the original PR breaks the test.

PR #933, opened 2024-04-03
Fix errors on non blocking assignments with RHS delays inside always_ff

### LINKED ISSUE #932: slang refuses always_ff with non blocking assignment and delay
**Describe the bug**
The following code is not allowed by slang:
```
`define D #1
module x;
reg a;
wire clk;
always_ff @(posedge clk)
  a <= `D 1'b0;
endmodule
```

**To Reproduce**
Run the code above.

**Additional context**
The LRM says:

_The always_ff procedure imposes the restriction that it contains one and only one event control and **no
blocking timing controls**. Variables on the left-hand side of assignments within an always_ff procedure,
including variables from the contents of a called function, shall not be written to by any other process._

I think the term _blocking timing controls_ is ill-defined (I couldn't find any definition in the spec of blocking vs non-blocking in the context of delays), but I don't think it includes the case above, which is typical for synthesizable code where you want such delays to make the output more clear.
The 2 simulators I tested happily accept this code.

udif:
This is my attempt to fix #932 .
It works, the new unit tests pass, but I'm really lacking confidence in my patch (at least it is very short and can be easily reviewed).

### DIFF
--- a/source/ast/Statements.cpp
+++ b/source/ast/Statements.cpp
@@ -2281,12 +2281,14 @@ Statement& ExpressionStatement::fromSyntax(Compilation& compilation,
             ok = true;
             break;
         }
-        case ExpressionKind::Assignment:
-            if (auto timing = expr.as<AssignmentExpression>().timingControl)
-                stmtCtx.observeTiming(*timing);
+        case ExpressionKind::Assignment: {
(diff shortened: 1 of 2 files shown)


MikePopoloski:
This makes sense to me, and seems to match behavior of other tools I tried.

PR #934, opened 2024-04-05
New tool, slang-hier

udif:
Here is a tool I needed so I wrote it.
It lists all the elaborated instances, along with their parameter values.
It is useful when trying to analyze a large design with many parameterized instances and generate blocks.


MikePopoloski:
FYI this needs to be fixed for the noexcept builds, but the idea seems fine to me.

udif:
As far as I can tell, failure seems to be unrelated to noexcept, but rather a transient failure due to not being able to download one of the tarballs for the build:
> Warning: Failed to download action 'https://api.github.com/repos/ilammy/msvc-dev-cmd/tarball/0b201ec74fa43914dc39ae48a89fd1d8cb592756'. Error: Response status code does not indicate success: 500 (Internal Server Error).  

Other jobs seems to be fine downloading the exact tarball:
> Download action repository 'ilammy/msvc-dev-cmd@v1' (SHA:0b201ec74fa43914dc39ae48a89fd1d8cb592756)

MikePopoloski:
Ah yeah, I didn't look closely at the failures. Some package repository glitch I guess.

PR #973, opened 2024-04-26
slang-hier bug fix (see #972)

### LINKED ISSUE #972: Slang incorrectly claims that a parameter is recursively defined
**Describe the bug**
In some places when calling the AST to get a parameter value, it triggers an erroneous error message.

**To Reproduce**
```
`define PARAM_DECL parameter X=1,\
                           parameter I=3,\
                           parameter W=I+1\


module t
#(// Parameters
  `PARAM_DECL,
  parameter J=2,
  parameter Y=I+1,
  parameter Z=J+1
)
  (
output [W-1:0]w,
output [Y-1:0]y,
output [X-1:0]z
);
assign w = {W{1'b0}};
endmodule
```
Run the following with:   
`slang-hier test.v`    
Which produces:
```
Module="t" Instance="t" 
Top level design units:
    t

```
Then run again with:  
`slang-hier --params test.v`  
Which produces:
```
Module="t" Instance="t" Parameters: X=1, I=3, W=4, J=2, Y=4, Z=3
Top level design units:
    t

test.v:14:9: error: 'W' recursively depends on its own definition
output [W-1:0]w,
        ^
test13.v:8:3: note: declared here
  `PARAM_DECL,
  ^
test.v:3:38: note: expanded from macro 'PARAM_DECL'
                           parameter W=I+1\
                                     ^

Build failed: 1 error, 0 warnings
```

Notice that the error appears only when 2 parameters are both defined in a macro and one depends on the other.
All other test cases, where only one is defined in the macro, do not trigger this.

The error is triggered by the following lines in `slang-hier.v` on lines 93-94:
```
                            if (p->symbol.kind == SymbolKind::Parameter)
                                v = p->symbol.as<ParameterSymbol>().getValue().toString();
```

The `assign` is in the test because this was an effort to get a minimum sized example to another bug where a parameter on the same code was incorrectly inferred as 0, but instead it triggered this bug I already ran into earlier but didn't have a clue as how to recreate.

udif:
This fixes #972 caused by a slang-hier bug.

### DIFF
--- a/tools/hier/hier.cpp
+++ b/tools/hier/hier.cpp
@@ -53,6 +53,7 @@ int main(int argc, char** argv) {
     bool ok = driver.parseAllSources();
 
     auto compilation = driver.createCompilation();
+    ok &= driver.reportCompilation(*compilation, /* quiet */ false);
 
     if (instRegex.has_value())
         regex = instRegex.value();
@@ -111,7 +112,6 @@ int main(int argc, char** argv) {
             }
         }));
     }
-    ok &= driver.reportCompilation(*compilation, /* quiet */ false);
 
     return ok ? 0 : 3;
 }


MikePopoloski:
This is a perfectly fine change but I fixed the API in b9d54238b0a62838c99a5a442312b95ad8e1c1f8 so it should no longer be needed.

PR #998, opened 2024-05-17
[slang][lang-compliance] Introduce nondegeneracy check

likeamahoney:
Based on 16.12.22 SystemVerilog LRM section and partially (F.4.3/F.5.2/F.5.5).

Also Ben Cohen's [paper about nondegeneracy](https://systemverilog.us/vf/Degeneracy111723Ben.pdf) understanding from [here](https://systemverilog.us/vf/Cohen_Links_to_papers_books.pdf).

Complex sequences are not often used in industrial/open source designs, but I think these checks will help to verify restrictions from 16.12.22 in most cases.

I written about ~300 lines of `sv` tests but i need help with more complex test cases but if anyone could also come up with more complex tests to test heuristics, that would be nice.

### DIFF
--- a/include/slang/ast/expressions/AssertionExpr.h
+++ b/include/slang/ast/expressions/AssertionExpr.h
@@ -7,10 +7,6 @@
 //------------------------------------------------------------------------------
 #pragma once
 
-#include <optional>
-#include <set>
-#include <utility>
-
 #include "slang/ast/ASTSerializer.h"
 #include "slang/ast/Expression.h"
 #include "slang/ast/TimingControl.h"
@@ -131,9 +127,9 @@ public:
     static const AssertionExpr& bind(const syntax::SequenceExprSyntax& syntax,
                                      const ASTContext& context, bool allowDisable = false);
 
-    static const AssertionExpr& bind(
-        const syntax::PropertyExprSyntax& syntax, const ASTContext& context,
-        bool allowDisable = false, const NondegeneracyFlags nondegFlags = NondegeneracyFlags::None);
+    static const AssertionExpr& bind(const syntax::PropertyExprSyntax& syntax,
+                                     const ASTContext& context, bool allowDisable = false,
+                                     NondegeneracyFlags nondegFlags = NondegeneracyFlags::None);
 
     static const AssertionExpr& bind(const syntax::PropertySpecSyntax& syntax,
                                      const ASTContext& context);
@@ -295,7 +291,7 @@ public:
     /// An optional repetition of the sequence.
     std::optional<SequenceRepetition> repetition;
 
-    /// Store `true` if sequence exression is can be evaluated as constant `false` (`0`).
+    /// Store `true` if sequence exression can be evaluated as constant `false` (`0`).
     bool isNullExpr;
 
     SimpleAssertionExpr(const Expression& expr, std::optional<SequenceRepetition> repetition,
--- a/include/slang/ast/expressions/AssertionExpr.h
+++ b/include/slang/ast/expressions/AssertionExpr.h
@@ -131,9 +131,9 @@ public:
     static const AssertionExpr& bind(const syntax::SequenceExprSyntax& syntax,
                                      const ASTContext& context, bool allowDisable = false);
 
-    static const AssertionExpr& bind(const syntax::PropertyExprSyntax& syntax,
-                                     const ASTContext& context, bool allowDisable = false,
-                                     const NondegeneracyFlags nondegFlags = NondegeneracyFlags::None);
+    static const AssertionExpr& bind(
+        const syntax::PropertyExprSyntax& syntax, const ASTContext& context,
+        bool allowDisable = false, const NondegeneracyFlags nondegFlags = NondegeneracyFlags::None);
 
     static const AssertionExpr& bind(const syntax::PropertySpecSyntax& syntax,
                                      const ASTContext& context);
@@ -206,7 +206,9 @@ public:
     explicit InvalidAssertionExpr(const AssertionExpr* child) :
         AssertionExpr(AssertionExprKind::Invalid), child(child) {}
 
-    bitmask<NondegeneracyStatus> checkNondegeneracyImpl() const { return NondegeneracyStatus::None; }
+    bitmask<NondegeneracyStatus> checkNondegeneracyImpl() const {
+        return NondegeneracyStatus::None;
+    }
 
     std::optional<SequenceRange> computeSequenceLengthImpl() const;
 
@@ -298,8 +300,8 @@ public:
 
     SimpleAssertionExpr(const Expression& expr, std::optional<SequenceRepetition> repetition,
                         bool isNullExpr = false) :
-        AssertionExpr(AssertionExprKind::Simple),
-        expr(expr), repetition(repetition), isNullExpr(isNullExpr) {}
+        AssertionExpr(AssertionExprKind::Simple), expr(expr), repetition(repetition),
+        isNullExpr(isNullExpr) {}
 
     void requireSequence(const ASTContext& context, DiagCode code) const;
 
@@ -413,7 +415,9 @@ public:
                        std::optional<SequenceRange> range) :
         AssertionExpr(AssertionExprKind::Unary), op(op), expr(expr), range(range) {}
 
-    bitmask<NondegeneracyStatus> checkNondegeneracyImpl() const { return expr.checkNondegeneracy(); }
+    bitmask<NondegeneracyStatus> checkNondegeneracyImpl() const {
+        return expr.checkNondegeneracy();
+    }
 
     std::optional<SequenceRange> computeSequenceLengthImpl() const {
         return expr.computeSequenceLength();
@@ -567,7 +571,9 @@ public:
     StrongWeakAssertionExpr(const AssertionExpr& expr, Strength strength) :
         AssertionExpr(AssertionExprKind::StrongWeak), expr(expr), strength(strength) {}
 
-    bitmask<NondegeneracyStatus> checkNondegeneracyImpl() const { return expr.checkNondegeneracy(); }
+    bitmask<NondegeneracyStatus> checkNondegeneracyImpl() const {
+        return expr.checkNondegeneracy();
+    }
 
     std::optional<SequenceRange> computeSequenceLengthImpl() const {
         return expr.computeSequenceLength();
@@ -603,8 +609,8 @@ public:
 
     AbortAssertionExpr(const Expression& condition, const AssertionExpr& expr, Action action,
                        bool isSync) :
-        AssertionExpr(AssertionExprKind::Abort),
-        condition(condition), expr(expr), action(action), isSync(isSync) {}
+        AssertionExpr(AssertionExprKind::Abort), condition(condition), expr(expr), action(action),
+        isSync(isSync) {}
 
     bitmask<NondegeneracyStatus> checkNondegeneracyImpl() const {
         return expr.checkNondegeneracy();
@@ -645,7 +651,9 @@ public:
         AssertionExpr(AssertionExprKind::Conditional), condition(condition), ifExpr(ifExpr),
         elseExpr(elseExpr) {}
 
-    bitmask<NondegeneracyStatus> checkNondegeneracyImpl() const { return NondegeneracyStatus::None; }
+    bitmask<NondegeneracyStatus> checkNondegeneracyImpl() const {
+        return NondegeneracyStatus::None;
+    }
 
     /// Returns maximum possible sequence length beetween `if` and `else` sequence expressions.
     std::optional<SequenceRange> computeSequenceLengthImpl() const;
@@ -689,10 +697,12 @@ public:
 
     CaseAssertionExpr(const Expression& expr, std::span<const ItemGroup> items,
                       const AssertionExpr* defaultCase) :
-        AssertionExpr(AssertionExprKind::Case),
-        expr(expr), items(items), defaultCase(defaultCase) {}
+        AssertionExpr(AssertionExprKind::Case), expr(expr), items(items), defaultCase(defaultCase) {
+    }
 
-    bitmask<NondegeneracyStatus> checkNondegeneracyImpl() const { return NondegeneracyStatus::None; }
+    bitmask<NondegeneracyStatus> checkNondegeneracyImpl() const {
+        return NondegeneracyStatus::None;
+    }
 
     /// Returns maximum possible sequence length beetween all case labels sequence expressions
     /// include `default` label
@@ -731,7 +741,9 @@ public:
     DisableIffAssertionExpr(const Expression& condition, const AssertionExpr& expr) :
         AssertionExpr(AssertionExprKind::DisableIff), condition(condition), expr(expr) {}
 
-    bitmask<NondegeneracyStatus> checkNondegeneracyImpl() const { return NondegeneracyStatus::None; }
+    bitmask<NondegeneracyStatus> checkNondegeneracyImpl() const {
+        return NondegeneracyStatus::None;
+    }
 
     /// Returns `nullopt` if `disable iff` condition is always `true` (`1`)
     std::optional<SequenceRange> computeSequenceLengthImpl() const;
--- a/include/slang/ast/expressions/AssertionExpr.h
+++ b/include/slang/ast/expressions/AssertionExpr.h
@@ -7,6 +7,10 @@
 //------------------------------------------------------------------------------
 #pragma once
 
+#include <optional>
+#include <set>
+#include <utility>
+
 #include "slang/ast/ASTSerializer.h"
 #include "slang/ast/Expression.h"
 #include "slang/ast/TimingControl.h"
@@ -67,6 +71,7 @@ SLANG_ENUM(BinaryAssertionOperator, OP)
 
 class ASTContext;
 class CallExpression;
+struct SequenceRange;
 
 /// The base class for assertion expressions (sequences and properties).
 class SLANG_EXPORT AssertionExpr {
@@ -91,12 +96,22 @@ public:
     /// and false otherwise.
     bool admitsEmpty() const;
 
+    bool isAdmitsOnlyEmpty() const { return admitsOnlyEmpty; }
+
+    /// Returns true if this is a sequence expression that admits no match,
+    /// and false otherwise.
+    bool admitsNoMatch() const;
+
+    /// Compute possible clock ticks length of sequence under assertion expression
+    std::optional<SequenceRange> computeSequenceLength() const;
+
     static const AssertionExpr& bind(const syntax::SequenceExprSyntax& syntax,
                                      const ASTContext& context, bool allowDisable = false);
 
     static const AssertionExpr& bind(const syntax::PropertyExprSyntax& syntax,
                                      const ASTContext& context, bool allowDisable = false,
-                                     bool allowSeqAdmitEmpty = false);
+                                     bool isFollowedByProp = false, bool isOverlapImpl = false,
+                                     bool isNonOverlapImpl = false);
 
     static const AssertionExpr& bind(const syntax::PropertySpecSyntax& syntax,
                                      const ASTContext& context);
@@ -155,6 +170,10 @@ protected:
     explicit AssertionExpr(AssertionExprKind kind) : kind(kind) {}
 
     static AssertionExpr& badExpr(Compilation& compilation, const AssertionExpr* expr);
+
+    /// Store `true` if sequence admits only empty matches.
+    /// Purpose for `mutable` is to update it in `admitsEmptyImpl` `const` methods.
+    mutable bool admitsOnlyEmpty = false;
 };
 
 /// @brief Represents an invalid expression
@@ -170,6 +189,10 @@ public:
         AssertionExpr(AssertionExprKind::Invalid), child(child) {}
 
     bool admitsEmptyImpl() const { return false; }
+
+    bool admitsNoMatchImpl() const { return false; }
+
+    std::optional<SequenceRange> computeSequenceLengthImpl() const;
 
     static bool isKind(AssertionExprKind kind) { return kind == AssertionExprKind::Invalid; }
 
@@ -190,6 +213,22 @@ struct SequenceRange {
                                     const ASTContext& context, bool allowUnbounded);
 
     void serializeTo(ASTSerializer& serializer) const;
+
+    bool operator!=(const SequenceRange& right) const;
+
+    bool operator==(const SequenceRange& right) const;
+
+    bool operator>(const SequenceRange& right) const;
+
+    bool operator<(const SequenceRange& right) const;
+
+    bool operator>=(const SequenceRange& right) const;
+
+    bool operator<=(const SequenceRange& right) const;
+
+    bool isIntersect(const SequenceRange& other) const;
+
+    bool isWithin(const SequenceRange& other) const;
 };
 
 /// Encodes a repetition of some sub-sequence.
@@ -221,7 +260,10 @@ struct SequenceRepetition {
         No,
 
         /// The sequence may or may not admit an empty match.
-        Depends
+        Depends,
+
+        /// The sequence admits only empty matches and nothing else.
+        Only,
     };
 
     /// Classifies the repetition as admitting an empty match or not.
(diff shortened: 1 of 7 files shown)


MikePopoloski:
Thank you for tackling this; the sections of the LRM that address this are daunting which is why I had not implemented it yet. I'll have to review this quite closely.

One thing that can help inform additional tests is looking at the coverage report, which shows some big chunks that haven't been tested at all yet.

include/slang/ast/expressions/AssertionExpr.h
```diff
@@ -91,12 +96,22 @@ class SLANG_EXPORT AssertionExpr {
     /// and false otherwise.
     bool admitsEmpty() const;
 
+    bool isAdmitsOnlyEmpty() const { return admitsOnlyEmpty; }
```
MikePopoloski, line 99: Can you rework these three functions into a single one that returns the admits empty / no-match properties via a single flags enum? You can move the `AdmitsEmpty` enum up here and expand it to include a bit for the no-match part as well.  Then you won't need the clunky extra `admitsOnlyEmpty` member below that is only validly set when you call admitsEmpty first, which seems like a footgun.

include/slang/ast/expressions/AssertionExpr.h
```diff
@@ -190,6 +213,22 @@ struct SequenceRange {
                                     const ASTContext& context, bool allowUnbounded);
 
     void serializeTo(ASTSerializer& serializer) const;
+
+    bool operator!=(const SequenceRange& right) const;
+
+    bool operator==(const SequenceRange& right) const;
+
+    bool operator>(const SequenceRange& right) const;
```
MikePopoloski, line 221: You can instead implement all of the comparison operators by implementing the "spaceship" operator: `operator<=>` to simplify this a bunch.

source/ast/expressions/AssertionExpr.cpp
```diff
@@ -417,6 +490,71 @@ void SequenceRange::serializeTo(ASTSerializer& serializer) const {
         serializer.write("max", "$"sv);
 }
 
+bool SequenceRange::operator!=(const SequenceRange& right) const {
```
MikePopoloski, line 493: I think the logic here could just be defaulted? You're just comparing field by field right? So `operator==(const SourceRange&) = default;` should have the correct behavior?

source/ast/expressions/AssertionExpr.cpp
```diff
@@ -417,6 +490,71 @@ void SequenceRange::serializeTo(ASTSerializer& serializer) const {
...
+
+bool SequenceRange::operator>=(const SequenceRange& right) const {
+    return (*this == right) || (*this > right);
+}
+
+bool SequenceRange::operator<=(const SequenceRange& right) const {
+    return (*this == right) || (*this < right);
+}
+
+/// `SequenceRange`s intersects if max delay of one range
```
MikePopoloski, line 532: Documentation comments should go in the header, otherwise I don't think they'll be picked up by doxygen into the doc build.

source/ast/expressions/AssertionExpr.cpp
```diff
@@ -1052,8 +1356,13 @@ AssertionExpr& StrongWeakAssertionExpr::fromSyntax(const StrongWeakPropertyExprS
     auto& expr = bind(*syntax.expr, context);
     expr.requireSequence(context);
 
-    if (expr.admitsEmpty())
-        context.addDiag(diag::SeqPropAdmitEmpty, syntax.expr->sourceRange());
+    // Any sequence that is used as a property shall be nondegenerate and shall not admit
+    // any empty match.
+    if (bool isSeqAdmitsEmpty = expr.admitsEmpty(); isSeqAdmitsEmpty || expr.admitsNoMatch()) {
```
MikePopoloski, line 1361: Rather than checking this again here can you rely on the bind() call above to do it by passing the right flags?

source/ast/expressions/AssertionExpr.cpp
```diff
@@ -1125,6 +1464,28 @@ void ConditionalAssertionExpr::serializeTo(ASTSerializer& serializer) const {
         serializer.write("else", *elseExpr);
 }
 
+/// Returns maximum possible sequence length beetween all case labels sequence expressions
+/// include `default` label
+std::optional<SequenceRange> CaseAssertionExpr::computeSequenceLengthImpl() const {
+    std::set<SequenceRange> seqRangeItems;
```
MikePopoloski, line 1470: You're just using the set for sorting here right? It's generally a good idea to avoid allocating containers unless strictly necessary, and in this case I think you can just store the current maximum and update it on each iteration if the new value is larger? You don't actually need to store all the items.

source/ast/expressions/AssertionExpr.cpp
```diff
@@ -1178,6 +1539,14 @@ void CaseAssertionExpr::serializeTo(ASTSerializer& serializer) const {
         serializer.write("defaultCase", *defaultCase);
 }
 
+/// Returns `nullopt` if `disable iff` condition is always `true` (`1`)
+std::optional<SequenceRange> DisableIffAssertionExpr::computeSequenceLengthImpl() const {
+    if (condition.constant && condition.constant->integer().countOnes())
```
MikePopoloski, line 1544: The `ConstantValue::isTrue()` function is probably a nicer way to express this than calling countOnes()

include/slang/ast/expressions/AssertionExpr.h
```diff
@@ -91,12 +96,22 @@ class SLANG_EXPORT AssertionExpr {
...
+    /// Compute possible clock ticks length of sequence under assertion expression
+    std::optional<SequenceRange> computeSequenceLength() const;
+
     static const AssertionExpr& bind(const syntax::SequenceExprSyntax& syntax,
                                      const ASTContext& context, bool allowDisable = false);
 
     static const AssertionExpr& bind(const syntax::PropertyExprSyntax& syntax,
                                      const ASTContext& context, bool allowDisable = false,
-                                     bool allowSeqAdmitEmpty = false);
+                                     bool isFollowedByProp = false, bool isOverlapImpl = false,
```
MikePopoloski, line 113: This is becoming too many boolean arguments in a row, which makes it easy to mess up the ordering of them when calling the function. I think this should be made a separate flags enum.

include/slang/ast/expressions/AssertionExpr.h
likeamahoney: Thanks for suggestion, fixed)!

source/ast/expressions/AssertionExpr.cpp
likeamahoney: it was removed in terms of fixing a previous comment

source/ast/expressions/AssertionExpr.cpp
likeamahoney: fixed

include/slang/ast/expressions/AssertionExpr.h
likeamahoney: fixed

source/ast/expressions/AssertionExpr.cpp
likeamahoney: fixed

source/ast/expressions/AssertionExpr.cpp
likeamahoney: currently I don't have any idea how to correctly fix that because there is call to `bind` overload which accepts `SequenceExprSyntax` instead of `PropertyExprSyntax`

there are no way to fix that without changes in `bind` call behavior or `StrongWeakPropertyExprSyntax  struct`

include/slang/ast/expressions/AssertionExpr.h
likeamahoney: currently it's failed on MacOS build - I'm working with it

likeamahoney: I chose to implement only the `operator<` and changed the comparison logic

likeamahoney: Spaceship (`<=>`) operator requires realization for any field but for `std::optional` that operator marked as deleted. It requires for changing processing of `SequenceRange`s with unbonded maximum I don't have any good idea for resolving with another way.

likeamahoney:
> > Thank you for tackling this; the sections of the LRM that address this are daunting which is why I had not implemented it yet. I'll have to review this quite closely.
> > One thing that can help inform additional tests is looking at the coverage report, which shows some big chunks that haven't been tested at all yet.
> 
> At the moment I only added a few tests to increase the possible coverage but at all it still in progress

And also add support for checking nondegenerancy for lhs of "followed-by property" properties (e.g. `#-#` and `#=#`)

include/slang/ast/expressions/AssertionExpr.h
```diff
@@ -87,16 +119,21 @@ class SLANG_EXPORT AssertionExpr {
...
 
     static const AssertionExpr& bind(const syntax::SequenceExprSyntax& syntax,
                                      const ASTContext& context, bool allowDisable = false);
 
-    static const AssertionExpr& bind(const syntax::PropertyExprSyntax& syntax,
-                                     const ASTContext& context, bool allowDisable = false,
-                                     bool allowSeqAdmitEmpty = false);
+    static const AssertionExpr& bind(
+        const syntax::PropertyExprSyntax& syntax, const ASTContext& context,
+        bool allowDisable = false, const NondegeneracyFlags nondegFlags = NondegeneracyFlags::None);
```
MikePopoloski, line 136: The `const` here does nothing -- since this is passed by value a copy will be made, which is fine, it's a small type.

include/slang/ast/expressions/AssertionExpr.h
```diff
@@ -239,11 +295,19 @@ class SLANG_EXPORT SimpleAssertionExpr : public AssertionExpr {
     /// An optional repetition of the sequence.
     std::optional<SequenceRepetition> repetition;
 
-    SimpleAssertionExpr(const Expression& expr, std::optional<SequenceRepetition> repetition) :
-        AssertionExpr(AssertionExprKind::Simple), expr(expr), repetition(repetition) {}
+    /// Store `true` if sequence exression is can be evaluated as constant `false` (`0`).
```
MikePopoloski, line 298: "expression", -is

source/ast/expressions/AssertionExpr.cpp
```diff
@@ -7,6 +7,9 @@
 //------------------------------------------------------------------------------
 #include "slang/ast/expressions/AssertionExpr.h"
 
+#include <iostream>
```
MikePopoloski, line 10: iostream include should be unnecessary

source/ast/expressions/AssertionExpr.cpp
```diff
@@ -417,6 +471,47 @@ void SequenceRange::serializeTo(ASTSerializer& serializer) const {
         serializer.write("max", "$"sv);
 }
 
+bool SequenceRange::operator<(const SequenceRange& right) const {
+    // if both sequences are unbounded than check min's
```
MikePopoloski, line 475: s/than/then

source/ast/expressions/AssertionExpr.cpp
```diff
@@ -417,6 +471,47 @@ void SequenceRange::serializeTo(ASTSerializer& serializer) const {
...
+    if (!max.has_value() && !right.max.has_value())
+        return min < right.min;
+
+    // if both sequences are bounded than compare it's diff's
```
MikePopoloski, line 479: Same here

source/ast/expressions/AssertionExpr.cpp
```diff
@@ -441,6 +536,11 @@ SequenceRepetition::SequenceRepetition(const SequenceRepetitionSyntax& syntax,
 }
 
 SequenceRepetition::AdmitsEmpty SequenceRepetition::admitsEmpty() const {
+    // If min and max delays are both zero, then the sequence accepts only empty matches.
+    if (range.min == 0 &&
+        (!range.max.has_value() || (range.max.has_value() && range.max.value() == 0)))
```
MikePopoloski, line 541: This will say that the repetition admits only empty matches when the max is unbounded (i.e. range.max is unset), which I think is not right. It's only when both values are actually zero that it admits only empty matches.

source/ast/expressions/AssertionExpr.cpp
```diff
@@ -523,22 +625,44 @@ void SimpleAssertionExpr::requireSequence(const ASTContext& context, DiagCode co
     }
 }
 
-bool SimpleAssertionExpr::admitsEmptyImpl() const {
+bitmask<NondegeneracyStatus> SimpleAssertionExpr::checkNondegeneracyImpl() const {
+    if (expr.kind == ExpressionKind::AssertionInstance) {
```
MikePopoloski, line 629: I'm not sure this is correct; this ignores the possible repetition when the sequence is a reference to another sequence. The repetition could be for [*0] meaning it would admit only empty matches, regardless of the instance's nondegeneracy status.

source/ast/expressions/AssertionExpr.cpp
```diff
@@ -523,22 +625,44 @@ void SimpleAssertionExpr::requireSequence(const ASTContext& context, DiagCode co
...
         auto result = repetition->admitsEmpty();
-        if (result == SequenceRepetition::AdmitsEmpty::Yes)
-            return true;
-        if (result == SequenceRepetition::AdmitsEmpty::No)
-            return false;
+        res = (result == SequenceRepetition::AdmitsEmpty::Yes)
+                  ? res | NondegeneracyStatus::AdmitsEmpty
+                  : res;
+        if (result == SequenceRepetition::AdmitsEmpty::Only)
+            res = res | NondegeneracyStatus::AcceptsOnlyEmpty | NondegeneracyStatus::AdmitsEmpty;
```
MikePopoloski, line 648: Can just use `|=` here.

source/ast/expressions/AssertionExpr.cpp
```diff
@@ -523,22 +625,44 @@ void SimpleAssertionExpr::requireSequence(const ASTContext& context, DiagCode co
...
+    // with no possible match
+    res = (isNullExpr) ? res | NondegeneracyStatus::AdmitsNoMatch : res;
+
     if (repetition) {
         auto result = repetition->admitsEmpty();
-        if (result == SequenceRepetition::AdmitsEmpty::Yes)
-            return true;
-        if (result == SequenceRepetition::AdmitsEmpty::No)
-            return false;
+        res = (result == SequenceRepetition::AdmitsEmpty::Yes)
```
MikePopoloski, line 644: This seems unnecessarily clunky; you can just check `if (result == SequenceRepetition::AdmitsEmpty::Yes)` and add the flag if so.

source/ast/expressions/AssertionExpr.cpp
```diff
@@ -523,22 +625,44 @@ void SimpleAssertionExpr::requireSequence(const ASTContext& context, DiagCode co
...
 
+    return res;
+}
+
+std::optional<SequenceRange> SimpleAssertionExpr::computeSequenceLengthImpl() const {
     if (expr.kind == ExpressionKind::AssertionInstance) {
         auto& aie = expr.as<AssertionInstanceExpression>();
         if (aie.type->isSequenceType())
-            return aie.body.admitsEmpty();
+            return aie.body.computeSequenceLength();
```
MikePopoloski, line 658: Same comment here about skipping the repetition on this expression, which seems like it should affect the computed length.

source/ast/expressions/AssertionExpr.cpp
```diff
@@ -594,26 +718,93 @@ AssertionExpr& SequenceConcatExpr::fromSyntax(const DelayedSequenceExprSyntax& s
...
 
-        if (it->delay.min > 1)
-            return false;
+        const auto rightNondegenSt = right->sequence->checkNondegeneracy();
+        // If any element of concat sequence is no match then all concat sequence is no match
+        if (rightNondegenSt.has(NondegeneracyStatus::AdmitsNoMatch))
+            admitsNoMatch = true;
+
+        // If right part is present with zero min and max delays
+        if (!right->delay.min && (!right->delay.max.has_value() || !right->delay.max.value())) {
```
MikePopoloski, line 746: Same comment here about the unbounded max being treated like zero here.

source/ast/expressions/AssertionExpr.cpp
```diff
@@ -594,26 +718,93 @@ AssertionExpr& SequenceConcatExpr::fromSyntax(const DelayedSequenceExprSyntax& s
...
+            if (left == elements.begin() &&
+                leftNondegenSt.has(NondegeneracyStatus::AcceptsOnlyEmpty))
+                admitsNoMatch = true;
+
+            // (seq ##0 empty) does not result in a match.
+            if (rightNondegenSt.has(NondegeneracyStatus::AcceptsOnlyEmpty))
+                admitsNoMatch = true;
+        }
+
+        if (!right->sequence->checkNondegeneracy().has(NondegeneracyStatus::AdmitsEmpty))
```
MikePopoloski, line 759: Should reuse `rightNondegenSt` here

source/ast/expressions/AssertionExpr.cpp
```diff
@@ -594,26 +718,93 @@ AssertionExpr& SequenceConcatExpr::fromSyntax(const DelayedSequenceExprSyntax& s
...
+            admitsEmpty = false;
+
+        if (right->delay.min == 0u && right->delay.max == 0u)
+            admitsEmpty = false;
```
MikePopoloski, line 763: This is the same check as the condition above, so you should be able to move it into the same branch.

source/ast/expressions/AssertionExpr.cpp
```diff
@@ -594,26 +718,93 @@ AssertionExpr& SequenceConcatExpr::fromSyntax(const DelayedSequenceExprSyntax& s
...
-    while (++it != elements.end()) {
-        if (!it->sequence->admitsEmpty())
-            return false;
+    if (leftNondegenSt.has(NondegeneracyStatus::AdmitsNoMatch))
+        admitsNoMatch = true;
 
-        if (it->delay.min == 0u && it->delay.max == 0u)
-            return false;
+    // See SystemVerilog LRM sections 16.9.2.1 and F.3.4.2.2 for details
+    while (admitsEmpty || !admitsNoMatch) {
```
MikePopoloski, line 735: Does this condition need to be `&&` instead of `||` ? I'm wondering if you can see a sequence that makes you set admitsEmpty to false and then you end the loop, whereas if you had kept going you would have found that a later sequence makes it actually accept no match at all?

source/ast/expressions/AssertionExpr.cpp
```diff
@@ -1186,6 +1541,14 @@ AssertionExpr& DisableIffAssertionExpr::fromSyntax(const DisableIffSyntax& synta
 
     checkSampledValueExpr(cond, context, false, diag::DisableIffLocalVar, diag::DisableIffMatched);
 
+    // If condition of `disable iff` can be evaluated only as `true` (`1`) than sequence property
+    // is no match
+    if (const auto evaluatedValue = context.tryEval(cond);
+        evaluatedValue.isInteger() && evaluatedValue.integer().countOnes()) {
```
MikePopoloski, line 1547: `isTrue`

source/ast/expressions/AssertionExpr.cpp
```diff
@@ -1125,6 +1452,27 @@ void ConditionalAssertionExpr::serializeTo(ASTSerializer& serializer) const {
         serializer.write("else", *elseExpr);
 }
 
+std::optional<SequenceRange> CaseAssertionExpr::computeSequenceLengthImpl() const {
+    std::optional<SequenceRange> max;
+    for (const auto& item : items) {
+        if (const auto itemLen = item.body->computeSequenceLength(); itemLen.has_value()) {
+            if (const auto itemLenVal = itemLen.value(); max.has_value() || max < itemLenVal)
```
MikePopoloski, line 1459: Is this meant to be `!max.has_value()` ?

source/ast/expressions/AssertionExpr.cpp
```diff
@@ -939,19 +1152,93 @@ void BinaryAssertionExpr::requireSequence(const ASTContext& context, DiagCode co
...
+            res = (leftAdmitsEmpty || rightAdmitsEmpty) ? res | NondegeneracyStatus::AdmitsEmpty
+                                                        : res;
+
+            // In case of `or` full sequence is no match if and only if both parts are no match
+            res = (leftAdmitsNoMatch && rightAdmitsNoMatch)
+                      ? res | NondegeneracyStatus::AdmitsNoMatch
+                      : res;
+            break;
+        }
+        case BinaryAssertionOperator::And: {
```
MikePopoloski, line 1175: You no longer set the admitsEmpty flag for the And case.

source/ast/expressions/AssertionExpr.cpp
```diff
@@ -939,19 +1152,93 @@ void BinaryAssertionExpr::requireSequence(const ASTContext& context, DiagCode co
...
+            // In case of `and` full sequence is no match if and only if any part is no match
+            res = (leftAdmitsNoMatch || rightAdmitsNoMatch)
+                      ? res | NondegeneracyStatus::AdmitsNoMatch
+                      : res;
+            break;
+        }
+        case BinaryAssertionOperator::Intersect: {
```
MikePopoloski, line 1182: Same here.

source/ast/expressions/AssertionExpr.cpp
```diff
@@ -939,19 +1152,93 @@ void BinaryAssertionExpr::requireSequence(const ASTContext& context, DiagCode co
...
+            res = (leftAdmitsNoMatch || rightAdmitsNoMatch || !leftLen.has_value() ||
+                   !rightLen.has_value() || !leftLen.value().isWithin(rightLen.value()))
+                      ? res | NondegeneracyStatus::AdmitsNoMatch
+                      : res;
+            break;
+        }
+        case BinaryAssertionOperator::Throughout: {
+            res = (rightAdmitsEmpty) ? res | NondegeneracyStatus::AdmitsEmpty : res;
+
+            // In case of `throught` full sequence is no match if right part is no match
```
MikePopoloski, line 1212: s/throught/throughout

include/slang/ast/expressions/AssertionExpr.h
```diff
@@ -7,6 +7,10 @@
 //------------------------------------------------------------------------------
 #pragma once
 
+#include <optional>
+#include <set>
```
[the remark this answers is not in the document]
likeamahoney: removed all unnecessary headers which I was added early

include/slang/ast/expressions/AssertionExpr.h
likeamahoney: fixed

include/slang/ast/expressions/AssertionExpr.h
likeamahoney: fixed

source/ast/expressions/AssertionExpr.cpp
likeamahoney: done

source/ast/expressions/AssertionExpr.cpp
likeamahoney: fixed

source/ast/expressions/AssertionExpr.cpp
likeamahoney: fixed

source/ast/expressions/AssertionExpr.cpp
likeamahoney: fixed

source/ast/expressions/AssertionExpr.cpp
likeamahoney: fixed

source/ast/expressions/AssertionExpr.cpp
likeamahoney: fixed

source/ast/expressions/AssertionExpr.cpp
likeamahoney: fixed

source/ast/expressions/AssertionExpr.cpp
likeamahoney: fixed

source/ast/expressions/AssertionExpr.cpp
likeamahoney: fixed

source/ast/expressions/AssertionExpr.cpp
likeamahoney: fixed

source/ast/expressions/AssertionExpr.cpp
likeamahoney: fixed for all locations when I used `countOnes`

source/ast/expressions/AssertionExpr.cpp
likeamahoney: yes it's typo - fixed

source/ast/expressions/AssertionExpr.cpp
likeamahoney: fixed

source/ast/expressions/AssertionExpr.cpp
likeamahoney: Currently I haven't fix that because concat sequences can be no match without empty subsequences for example:

```verilog
sequence s;
    1'b0 ##1 seq1 ##2 seq2  // empty because of first `false` don't give an any chance to match it - `1'b0`
endsequence
```
If we change loop condition from `||` to `&&` we may have a very limited info about certain type of sequences. For example:

```verilog
sequence s;
    1'b0 ##1 seq1 ##2 seq2[*0]  // this seq will be marked only as "no match" but not as "admits an empty matches"
endsequence
```

MikePopoloski:
Ok, I think this is good enough to land -- we can tweak or improve it later if needed.

PR #1005, opened 2024-05-22
Fix netlist creation when code contains concurrent assertions

### LINKED ISSUE #1002: Assertion 'T::isKind(kind)' failed
**Describe the bug**
A full SoC run passes with `slang` but when it is ran with `slang-netlist` (no additional flags), it fails with:

```
Assertion 'T::isKind(kind)' failed
  in file /home/udif/git2/slang/tools/netlist/../../include/slang/ast/Statements.h, line 238
  function: const T& slang::ast::Statement::as() const [with T = slang::ast::TimedStatement]
```
.

**To Reproduce**
At the moment that would be hard, as it is the result of proprietary SoC full tree run.

**Additional context**
I wish I could give out more information, but if the assertion could provide a `SourceLocation` I might be able to reduce it to an MRE.

udif:
This is the beginning of a PR to fix #1002 .
It works for a simple testcase that fails otherwise, but I still see different failures (DirectedGraph issue, might be unrelated) when reading larger examples such as those provided in #1002, so I don't think it is ready yet, but it's a start and it can be reviewed.


tools/netlist/include/NetlistVisitor.h
```diff
@@ -605,6 +605,14 @@ class NetlistVisitor : public ast::ASTVisitor<NetlistVisitor, true, false> {
         ast::EdgeKind edgeKind = ast::EdgeKind::None;
         if (symbol.procedureKind == ast::ProceduralBlockKind::AlwaysFF ||
             symbol.procedureKind == ast::ProceduralBlockKind::Always) {
+            // auto& body = symbol.getBody();
```
MikePopoloski, line 608: Remove extraneous commented out line.

tools/netlist/include/NetlistVisitor.h
```diff
@@ -605,6 +605,14 @@ class NetlistVisitor : public ast::ASTVisitor<NetlistVisitor, true, false> {
...
+            if (symbol.getBody().kind == ast::StatementKind::Block) {
```
MikePopoloski, line 609: Do you want to just ignore all statements that aren't a TimedStatement? Otherwise you'll still crash below.

tools/netlist/tests/NameTests.cpp
```diff
@@ -55,3 +55,40 @@ endmodule
     auto netlist = createNetlist(compilation);
     CHECK(netlist.numNodes() > 0);
 }
+
+TEST_CASE("Ignore concurrent assertions") {
+    // Test that unused modules are not visited by the netlist builder.
```
MikePopoloski, line 60: Comment is copy pasta from the previous test case.

tools/netlist/include/NetlistVisitor.h
udif: As I'm not sure what elements can be nested in the AST, I would rather get a fault during debugging rather than silently skipping netlist elements by mistake.
At the moment, I'm explicitly checking for those AST elements I know I need to skip, in this case a concurrent assertion under a sequential block.
Can concurrent assertions be found elsewhere?

tools/netlist/tests/NameTests.cpp
udif: fixed locally, will be pushed when I have more meaningful changes later.

MikePopoloski:
What's the status of this PR?

udif:
I apologize, after being blocked by #1007 I forgot about this, thanks for reminding me.
I will have to rebase it on the new Netlist code. It will take me a few more days.

udif:
I think it is OK now.

PR #1011, opened 2024-05-26
Added printing of source line info

KennethDRoe:
I think that having source file/line info in the json outputs will be useful to the community as a whole.  I'm submitting a pull request to contribute this change.


MikePopoloski:
Thanks for the PR. In case it wasn't clear, I can't land this until the tests are fixed.

Also I think this output should be hidden behind a command line flag. The JSON output is already very large for real world designs, and this will make it much larger. I suspect many people won't want to see this in the output, so it should be controlled via the flag.

source/ast/ASTSerializer.cpp
```diff
@@ -198,6 +215,20 @@ void ASTSerializer::visit(const T& elem, bool inMembersArray) {
     else if constexpr (std::is_base_of_v<Statement, T>) {
         writer.startObject();
         write("kind", toString(elem.kind));
+        if (elem.syntax != nullptr) {
+            write("source_file_start",
```
MikePopoloski, line 219: There is a lot of duplication here. Please extract this logic out into a helper method that can operate on a SyntaxNode so it's not repeated several times.

KennethDRoe:
I’ll add a flag and fix the tests. It make a few days.             - KenSent from my iPhoneOn May 29, 2024, at 8:43 PM, Michael Popoloski ***@***.***> wrote:﻿
Thanks for the PR. In case it wasn't clear, I can't land this until the tests are fixed.
Also I think this output should be hidden behind a command line flag. The JSON output is already very large for real world designs, and this will make it much larger. I suspect many people won't want to see this in the output, so it should be controlled via the flag.

MikePopoloski:
It seems that the tests are still failing in the CI build. Also I got access to a Macbook M3 and the tests all pass locally there, so I'm not sure why you're unable to get them to pass on your machine unless the problem is specific to the M1s.

KennethDRoe:
The failures on my machine may be due to the software stack (and not the M1
processor).

Can you send over the list of failures you are seeing with my pull
request?  Also, make sure you have the latest version of files from my
fork.  I made a couple of last minute fixes.

                  - Ken

MikePopoloski:
The CI results are on this thread of the PR. For example: https://github.com/MikePopoloski/slang/actions/runs/9364528795/job/25888520547?pr=1011

KennethDRoe:
I believe I fixed the one failing test.  Try again.

            - Ken

PR #1019, opened 2024-06-03
[slang] Fix bug with SignalEventControl iff clause emitting.

likeamahoney:
There is copypasta bug with `SignalEventControl` `iffCondition` emitting during it's AST constructing. Right hand of `iff` was constructed from left hand part.

And I also added support for `iffCondition` serializing.


MikePopoloski:
Nice, thanks for fixing.

Issue #1022, opened 2024-06-07
False degenerate error when using within-statement


jrudess:
I believe the below example is a case of a false degenerate report, or if it were degenerate and could never emit a match, I would expect simulation failures for this assertion with VCS.
```
slangtest167.sv:9:22: error: any sequence that is used as a property shall be nondegenerate and shall not admit any empty match
        $fell(a) |-> !$stable(b) within((~a)[*1:$] ##1 a);
                     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
slangtest167.sv:7:14: note: while expanding property 'p1'
    property p1(clk, rstb, a, b);
             ^
slangtest167.sv:9:22: note: sequence property can never be matched
        $fell(a) |-> !$stable(b) within((~a)[*1:$] ##1 a);
                     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
```Systemverilog
module top;
    logic clk;
    logic rstb;
    logic a;
    logic b;

    property p1(clk, rstb, a, b);
        @(posedge clk) disable iff(~rstb)
        $fell(a) |-> !$stable(b) within((~a)[*1:$] ##1 a);
    endproperty

    LABEL1: assert property(p1(clk, rstb, a, b));
endmodule
```

likeamahoney:
Hi! Thanks for the bug report!

I opened the PR which fixes that - https://github.com/MikePopoloski/slang/pull/1023

MikePopoloski:
Fixed in 2122d639817cca5234d377baa2e1b2aa8abf4592 instead

PR #1023, opened 2024-06-07
[fix][slang][nondegeneracy] Fix implication consequent nondegeneracy check

likeamahoney:
Fixes https://github.com/MikePopoloski/slang/issues/1022

The consequent of the implication should be ignored, since there are no restrictions on it in the standard.

### DIFF
--- a/include/slang/ast/expressions/AssertionExpr.h
+++ b/include/slang/ast/expressions/AssertionExpr.h
@@ -151,6 +151,9 @@ public:
         /// or followed-by operator shall admit at least one match. Such a sequence can
         /// admit only empty matches.
         NonOverlapOp,
+
+        /// Don't perform any nondegeneracy check
+        None,
     };
 
     static const AssertionExpr& bind(const syntax::SequenceExprSyntax& syntax,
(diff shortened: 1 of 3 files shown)


likeamahoney:
Have no idea why it's segfaults on `ThreadPoolTests`. After looking on other PR's I suppose this is a common problem for current CI state.

likeamahoney:
> Have no idea why it's segfaults on `ThreadPoolTests`. After looking on other PR's I suppose this is a common problem for current CI state.

win64-release build issue

MikePopoloski:
Looks like the Windows builds are broken only on GitHub due to this: https://github.com/actions/runner-images/issues/10004

MikePopoloski:
I don't think this is the right solution. The rule "Any sequence that is used as a property shall be nondegenerate and shall not admit any empty match." should still apply here. The reason it's broken is that the checking for the "within" condition is just incorrect. I'll see about putting up an alternative fix later today.

MikePopoloski:
Fixed in 2122d639817cca5234d377baa2e1b2aa8abf4592 instead

PR #1031, opened 2024-06-15
Fix hierarchical paths inside generate arrays

povik:
Take some input with an unnamed generate array, e.g.

    module top;
      genvar i;
      for (i = 0; i < 1; i = i + 1) begin
        logic a;
      end
    endmodule

make a fix so that the hierpath of the wire is `top.genblk1[0].a` not
 `top[0].a`.


MikePopoloski:
Thanks for the PR. Can you add a test that covers this logic so that it doesn't break in the future?

povik:
@MikePopoloski take a look

PR #1040, opened 2024-06-28
Add multiple slang-tidy checks and minor fixes in existing ones

JoelSole-Semidyn:
This PR adds some new slang-tidy checks, along with some minor fixes to existing ones, mainly aiming to throw the warnings in a more precise location. The new slang-tidy checks added to this PR are the following:

### style checks
* _AlwaysCombNamed_
This check ensures that all always_comb blocks have a name

* _GenerateNamed_
Similar to the previous check, this ensures that all generate blocks are named, including for loops

* _NoDotVarInPortConnection_
A check analogous to _NoDotStarInPortConnection_, enforces that no '.var' is used in port connection, and '.var(var)' should be used instead

* _NoLegacyGenerate_
This check enforces that no generate block is used in the code, since it is an old system verilog syntax

### synthesis checks
* _AlwaysFFAssignmentOutsideConditional_
Check that ensures that if there is a conditional block with reset, all the assignments must be either inside the if or else statements

* _UnusedSensitiveSignal_
This check verifies that every signal inside the sensitivity list of an 'always' or 'always_ff' block is used inside the block statement. The only exception is the clock signal

### DIFF
--- a/tools/tidy/CMakeLists.txt
+++ b/tools/tidy/CMakeLists.txt
@@ -22,6 +22,7 @@ add_library(
   src/style/NoDotStarInPortConnection.cpp
   src/style/NoImplicitPortNameInPortConnection.cpp
   src/style/AlwaysCombNamed.cpp
+  src/style/GenerateNamed.cpp
   )
 
 target_include_directories(slang_tidy_obj_lib PUBLIC include ../../include)
--- a/tools/tidy/CMakeLists.txt
+++ b/tools/tidy/CMakeLists.txt
@@ -25,6 +25,7 @@ add_library(
   src/style/GenerateNamed.cpp
   src/style/NoDotVarInPortConnection.cpp
   src/style/NoLegacyGenerate.cpp
+  src/synthesis/AlwaysFFAssignmentOutsideConditional.cpp
   )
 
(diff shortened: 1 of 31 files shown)


jcurtiss8086:
I have not reviewed all of the current Tidy checks, so my apologies if this is covered.   `NoForInsideGenerate` seems like it would want an equivalent `NoIfInsideGenerate` as well.

Rather than having separate options to flag the use of for/if blocks inside `generate` statemnts, how about a different option `NoLegacyGenerate` which just flags any use of the keyword 'generate' at all?  This is an option that Verible has for its lint rules.

JoelSole-Semidyn:
You are totally right, so I just uploaded a new version with the check _NoLegacyGenerate_ instead of _NoForInsideGenerate_

> I have not reviewed all of the current Tidy checks, so my apologies if this is covered. `NoForInsideGenerate` seems like it would want an equivalent `NoIfInsideGenerate` as well.
> 
> Rather than having separate options to flag the use of for/if blocks inside `generate` statemnts, how about a different option `NoLegacyGenerate` which just flags any use of the keyword 'generate' at all? This is an option that Verible has for its lint rules.

Issue #947, opened 2024-04-07
Add a global check for overlapping net aliases


MikePopoloski:
```
To avoid errors in specification, it is not allowed to specify an alias from an individual signal to itself or to
specify a given alias more than once. The following version of the preceding code would be illegal because
the top 4 bits and bottom 4 bits are the same in both statements:
  alias bus16 = {high12[11:8], low12};
  alias bus16 = {high12, low12[3:0]};

This alternative is also illegal because the bits of bus16 are being aliased to itself:
  alias bus16 = {high12, bus16[3:0]} = {bus16[15:12], low12};
```

PR #1045, opened 2024-07-04
[slang] Introduce net aliases bits duplication checks

likeamahoney:
Introduce 2 new net alias bit checks from SystemVerilog LRM section 10.11:

- check that net aliased to itself
- check that aliased bits specified more than once at module instance scope


likeamahoney:
I think with it LRM section 10.11 is done

include/slang/ast/symbols/InstanceSymbols.h
```diff
@@ -207,6 +208,10 @@ class SLANG_EXPORT InstanceBodySymbol : public Symbol, public Scope {
     const DefinitionSymbol& definition;
     mutable std::span<const Symbol* const> portList;
     std::span<const ParameterSymbolBase* const> parameters;
+    // Store any net alias bits to check it's possible intersections.
+    // Since alias statements can appear anywhere module instance statements can appears
+    // good option is to store aliased bits info in the module instance scope.
+    mutable AliasMap* aliasedBits;
```
MikePopoloski, line 214: It seems unnecessary to have an alias map allocated for every instance. The rules about net alias overlap are global, so it seems fine to have just one alias map stored in the Compilation object that all aliases can access. Then you don't need a custom allocator for it either.

source/ast/symbols/MemberSymbols.cpp
```diff
@@ -2196,7 +2196,6 @@ NetAliasSymbol& NetAliasSymbol::fromSyntax(const ASTContext& parentContext,
         if (!netType.isError()) {
             SmallVector<const IdentifierNameSyntax*> implicitNetNames;
             Expression::findPotentiallyImplicitNets(*expr, context, implicitNetNames);
-
```
MikePopoloski, line 2199: Random newline deletion seems unrelated to the rest of the diff

source/ast/symbols/MemberSymbols.cpp
```diff
@@ -2214,34 +2213,89 @@ struct NetAliasVisitor {
...
+                        aliasedBits[sym] = {
+                            AliasedBits{std::nullopt, std::nullopt, expr.sourceRange}};
+                    }
+                    break;
+                }
+                case ExpressionKind::ElementSelect: {
+                    if (const auto* sym = processExpr(expr)) {
+                        const auto& eSelExpr = expr.template as<ElementSelectExpression>();
+                        // If value can't be evaluated that means it is unbounded
+                        const auto evalVal = toInt(context.tryEval(eSelExpr.selector()));
```
MikePopoloski, line 2264: I don't think this is correct in all cases. An example is when there is multiple selections of a multi-dimensional array. It's actually pretty tricky to figure out the statically assigned bits of an arbitrary expression -- that's essentially what the ValueDriver code in ValueSymbol.cpp is doing to be able to issue multi-driven errors. It occurs to me that it may make more sense to use that code to implement the net alias check. You could add a new flag to AssignFlags that says it's a net alias and do the check for overlap there.

include/slang/ast/symbols/InstanceSymbols.h
likeamahoney: reimplemented all in order to last commentary

source/ast/symbols/MemberSymbols.cpp
likeamahoney: fixed

source/ast/ASTContext.cpp
```diff
@@ -136,7 +136,8 @@ void ASTContext::setAttributes(const Expression& expr,
 
 void ASTContext::addDriver(const ValueSymbol& symbol, const Expression& longestStaticPrefix,
                            bitmask<AssignFlags> assignFlags) const {
-    if (flags.has(ASTFlags::NotADriver) || scope->isUninstantiated())
+    if ((flags.has(ASTFlags::NotADriver) && !assignFlags.has(AssignFlags::NetAlias)) ||
```
MikePopoloski, line 139: I think you can get rid of this change if you simply don't use the NotADriver flag when binding the net alias expressions

source/ast/symbols/MemberSymbols.cpp
```diff
@@ -2119,16 +2124,18 @@ std::span<const Expression* const> NetAliasSymbol::getNetReferences() const {
     auto syntax = getSyntax();
     SLANG_ASSERT(scope && syntax);
 
-    // TODO: there should be a global check somewhere that any given bit
-    // of a net isn't aliased to the same target signal bit multiple times.
     bitwidth_t bitWidth = 0;
     bool issuedError = false;
     SmallVector<const Expression*> buffer;
     ASTContext context(*scope, LookupLocation::after(*this),
                        ASTFlags::NonProcedural | ASTFlags::NotADriver);
```
MikePopoloski, line 2131: Down here, you can remove the NotADriver

source/ast/symbols/MemberSymbols.cpp
```diff
@@ -2119,16 +2124,18 @@ std::span<const Expression* const> NetAliasSymbol::getNetReferences() const {
...
     NetAliasVisitor visitor(context);
+    // In order to save the leftmost alias
+    SmallMap<const ValueSymbol*, const Expression*, 2> leftAlias;
+
+    const auto nets = syntax->as<NetAliasSyntax>().nets;
 
-    for (auto exprSyntax : syntax->as<NetAliasSyntax>().nets) {
+    for (auto exprSyntax : nets) {
         auto& netRef = Expression::bind(*exprSyntax, context);
         if (!netRef.requireLValue(context))
```
MikePopoloski, line 2140: And then here pass your new AssignFlag and then you won't need the additional addDriver call below I think.

MikePopoloski (commented):
Ok yeah, this is way better. I left some additional comments to help simplify this a bit more.

source/ast/symbols/MemberSymbols.cpp
likeamahoney: done

source/ast/symbols/ValueSymbol.cpp
```diff
@@ -7,6 +7,8 @@
 //------------------------------------------------------------------------------
 #include "slang/ast/symbols/ValueSymbol.h"
 
+#include <iostream>
```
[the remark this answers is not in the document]
likeamahoney: removed

source/ast/ASTContext.cpp
likeamahoney: fixed

source/ast/symbols/MemberSymbols.cpp
likeamahoney: The flag is needed to skip the first alias (the leftmost one) in order to prevent FP. I left it, but remove it after the first iteration of the loop.

source/ast/symbols/MemberSymbols.cpp
```diff
@@ -2144,6 +2159,21 @@ std::span<const Expression* const> NetAliasSymbol::getNetReferences() const {
 
         netRef.visit(visitor);
         buffer.push_back(&netRef);
+
+        if (exprSyntax != *nets.begin()) {
```
MikePopoloski, line 2163: This logic seems confusing to me; you really just need to check whether the same symbol appears in more than one of the net alias expressions right? The `aliasedSyms` map in the NetAliasVisitor is already tracking that; instead of insert_or_assign up above you can just check whether the insertion fails because the symbol already exists, in which case you can issue the error right there. Then you don't need any of this special casing for it here and the separate map tracking the same thing.

source/ast/symbols/MemberSymbols.cpp
MikePopoloski: I'm not following why the first alias expression should be exempt from the duplicate checking. There isn't anything special about the first alias vs any others; you can reverse the order without changing the behavior.

I do think we need more info included in the addDriver call though, since it's only an error if the overlap occurs with the same target net? Like you can alias the same range of bits from one net to multiple different nets, it's only error if it's the same nets each time. This example from the LRM I think won't pass with the current code?
```
module overlap(inout wire [15:0] bus16, inout wire [11:0] low12, high12);
alias bus16[11:0] = low12;
alias bus16[15:4] = high12;
endmodule

module overlap(inout wire [15:0] bus16, inout wire [11:0] low12, high12);
alias bus16 = {high12, low12[3:0]};
alias high12[7:0] = low12[11:4];
endmodule
```
Or maybe it will, but not if you reverse the ordering in the alias statements.

source/ast/symbols/MemberSymbols.cpp
likeamahoney: solution was reimplemented by focusing on merge net alias expressions 

`aliasedSyms` map logic was reduced from `NetAliasVisitor`

likeamahoney: My solution no longer ignores the leftoutermost net alias expression

likeamahoney: To relate net alias expressions to existing alias sybmols, I used a comparison of their syntaxes

it's not a best solution but in this case it is possible, since in aliases all symbols are global nets from the point of view of the module, in the future we need to come up with a better comparator

likeamahoney:
closes https://github.com/MikePopoloski/slang/issues/947

source/ast/symbols/MemberSymbols.cpp
likeamahoney: > I'm not following why the first alias expression should be exempt from the duplicate checking. There isn't anything special about the first alias vs any others; you can reverse the order without changing the behavior.
> 
> I do think we need more info included in the addDriver call though, since it's only an error if the overlap occurs with the same target net? Like you can alias the same range of bits from one net to multiple different nets, it's only error if it's the same nets each time. This example from the LRM I think won't pass with the current code?
> 
> module overlap(inout wire [15:0] bus16, inout wire [11:0] low12, high12);
> alias bus16[11:0] = low12;
> alias bus16[15:4] = high12;
> endmodule
> 
> module overlap(inout wire [15:0] bus16, inout wire [11:0] low12, high12);
> alias bus16 = {high12, low12[3:0]};
> alias high12[7:0] = low12[11:4];
> endmodule
> Or maybe it will, but not if you reverse the ordering in the alias statements.

Provided example works correctly regardless of the alias permutations.

source/ast/symbols/MemberSymbols.cpp
```diff
@@ -2119,18 +2119,40 @@ std::span<const Expression* const> NetAliasSymbol::getNetReferences() const {
...
+    auto& comp = context.getCompilation();
+    auto& aliasedSyms = comp.getNetAliases();
     NetAliasVisitor visitor(context);
 
-    for (auto exprSyntax : syntax->as<NetAliasSyntax>().nets) {
+    const auto nets = syntax->as<NetAliasSyntax>().nets;
+
+    // Preprocessing net alias expressions to relate them to existing net aliases.
+    const NetAliasSymbol* current = this;
+    for (auto exprSyntax : nets) {
```
MikePopoloski, line 2134: The logic here still seems wrong. Consider this simple example:
```
module m;
    wire [1:0] a, b, c;
    alias a = b[1:0];
    alias c = b[1:0];
endmodule
```
which results in the following error (when it should not):
```
source:5:15: error: it is not allowed to specify an alias from an individual signal to itself or to specify a given alias more than once
    alias c = b[1:0];
              ^~~~~~
source:4:15: note: alias declared here
    alias a = b[1:0];
              ^~~~~~
```

I don't think this mergeOrAddNetAlias idea works or is even necessary really. We don't need a compilation-wide map. Each net can already track which bits have been aliased using the driver mechanism. I think what you need to do is to do an addDriver() call across the cartesian product of the aliases in a given net alias statement. Instead of the "containing symbol" being the net alias, instead you make it the net it's being *aliased to*.

As an example:
```
alias a = b = c;
```
you want the following sequence of addDriver calls to be made, where the argument here is what gets set as the "containing symbol" inside ValueDriver:
a.addDriver(b);
a.addDriver(c);
b.addDriver(a);
b.addDriver(c);
c.addDriver(a);
c.addDriver(b);

And then I think the existing logic will just do the right thing without the need for a global map or comparing syntax nodes to each other.

include/slang/ast/SemanticFacts.h
```diff
@@ -132,9 +132,12 @@ enum class SLANG_EXPORT AssignFlags : uint8_t {
 
     /// The assignment is for an output port that was sliced due to an array of instances
     /// being connected to an array argument.
-    SlicedPort = 1 << 7
+    SlicedPort = 1 << 7,
+
+    // The assignment declares net alias
```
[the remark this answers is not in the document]
likeamahoney: fixed

include/slang/ast/ASTContext.h
```diff
@@ -298,6 +299,9 @@ class SLANG_EXPORT ASTContext {
     /// Various flags that control how AST creation is performed.
     bitmask<ASTFlags> flags;
 
+    /// Placeholder for storing currently processed net alias
+    const NetAliasSymbol* netAlias = nullptr;
```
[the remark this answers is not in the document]
likeamahoney: removed

source/ast/symbols/MemberSymbols.cpp
likeamahoney: sorry for delay

solution was reimplemented according to your suggestion

likeamahoney: but i need to add more complex tests

likeamahoney: and fix CI

likeamahoney: CI was fixed and new tests were added

likeamahoney: CI was fixed and new tests were added

likeamahoney: CI was fixed and new tests were added

PR #1047, opened 2024-07-05
[slang] Fix PLA task concatenation ascending order

likeamahoney:
Added support for elaborating correctness of examples like the one from LRM section 20.16:

```verilog
module async_array(a1,a2,a3,a4,a5,a6,a7,b1,b2,b3);
  input a1, a2, a3, a4, a5, a6, a7 ;
  output b1, b2, b3;
  logic [1:7] mem[1:3]; // memory declaration for array personality
  logic b1, b2, b3;
  initial begin
      // set up the personality from the file array.dat
     $readmemb("array.dat", mem);
     // set up an asynchronous logic array with the input
     // and output terms expressed as concatenations
     $async$and$array(mem,{a1,a2,a3,a4,a5,a6,a7},{b1,b2,b3});
  end
endmodule
```

By default, slang sets all concatenation expression bitwidth ranges in descending order. I added an ASTFlag to change this behavior when concatenation is an argument of the PLA task.

### DIFF
--- a/source/ast/builtins/SystemTasks.cpp
+++ b/source/ast/builtins/SystemTasks.cpp
@@ -25,8 +25,8 @@ using namespace syntax;
 
 class SystemTaskBase : public SystemSubroutine {
 public:
-    explicit SystemTaskBase(const std::string& name, bool isPLATask = false) :
-        SystemSubroutine(name, SubroutineKind::Task, isPLATask) {}
+    explicit SystemTaskBase(const std::string& name) :
+        SystemSubroutine(name, SubroutineKind::Task) {}
 
     ConstantValue eval(EvalContext& context, const Args&, SourceRange range,
                        const CallExpression::SystemCallInfo&) const final {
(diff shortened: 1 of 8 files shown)


MikePopoloski:
This seems like an overly intrusive way of fixing this -- I'd prefer to avoid adding a specialized ASTFlag just for this one case. It seems like it should be fixable in the PLATask handler itself by changing checkArguments to see that it's a concat expression and just not issuing an error in that case.

source/ast/builtins/SystemTasks.cpp
```diff
@@ -725,9 +725,9 @@ class SdfAnnotateTask : public SystemTaskBase {
     }
 };
 
-class PlaTask : public SystemTaskBase {
+class SLANG_EXPORT PlaTask : public SystemTaskBase {
```
MikePopoloski, line 728: Adding SLANG_EXPORT here doesn't do anything; this class is not externally visible to downstream users of the library.

likeamahoney: removed

likeamahoney:
> This seems like an overly intrusive way of fixing this -- I'd prefer to avoid adding a specialized ASTFlag just for this one case. It seems like it should be fixable in the PLATask handler itself by changing checkArguments to see that it's a concat expression and just not issuing an error in that case.

fixed

PR #1048, opened 2024-07-05
[slang] Fix PATHPULSE limit values

likeamahoney:
In order to `SystemVerilog` LRM grammar it is legal to specialize only 1 value for both `PATHPULSE$` reject and error limits.

According to 30.7.1 section:

```
pulse_control_specparam ::=
          PATHPULSE$ = ( reject_limit_value [ , error_limit_value ] )
       | PATHPULSE$specify_input_terminal_descriptor$specify_output_terminal_descriptor
= ( reject_limit_value [ , error_limit_value ] )
```

At the text below LRM says:

"The path (data=>q) is not explicitly defined in any of the PATHPULSE$ declarations; therefore, it acquires reject and error limit of 3, as defined by the last PATHPULSE$ declaration." for such example:

```verilog
specify
...
    (data => q) = 10;
...
    specparam
    ...
        PATHPULSE$ = 3;
endspecify
```

### DIFF
--- a/source/parsing/Parser_members.cpp
+++ b/source/parsing/Parser_members.cpp
@@ -3223,13 +3223,10 @@ SpecparamDeclaratorSyntax& Parser::parseSpecparamDeclarator(SyntaxKind parentKin
         closeParen = expect(TokenKind::CloseParenthesis);
 
     if (!name.isMissing()) {
-        if (isPathPulse) {
-            if (parentKind != SyntaxKind::SpecifyBlock)
-                addDiag(diag::PulseControlSpecifyParent, name.range());
-            else if (!expr2)
-                expr2 = &expr1;
-        }
-        else if (expr2) {
+        if (isPathPulse && parentKind != SyntaxKind::SpecifyBlock)
+            addDiag(diag::PulseControlSpecifyParent, name.range());
+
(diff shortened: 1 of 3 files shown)


source/parsing/Parser_members.cpp
```diff
@@ -3227,7 +3227,7 @@ SpecparamDeclaratorSyntax& Parser::parseSpecparamDeclarator(SyntaxKind parentKin
             if (parentKind != SyntaxKind::SpecifyBlock)
                 addDiag(diag::PulseControlSpecifyParent, name.range());
             else if (!expr2)
-                addDiag(diag::PulseControlTwoValues, expr1.sourceRange()) << name.range();
+                expr2 = &expr1;
```
MikePopoloski, line 3230: This should not duplicate the expression here -- leaving it null seems fine. The parse tree must maintain the property that visiting the tokens in it via depth-first traversal recovers the original source file as typed by the user.

MikePopoloski (commented):
Huh, not sure why I added this but it looks like you're correct.

source/parsing/Parser_members.cpp
likeamahoney: fxied in last commit with simple code refactoring

PR #1049, opened 2024-07-05
[slang] Make string simple type

likeamahoney:
To support such sample from SystemVerilog LRM section 10.9.2:

```verilog
module top;
   typedef struct {
      logic [7:0] a;
      bit b;
      bit signed [31:0] c;
      string s;
   } sa;
   
   sa s2;
   initial s2 = '{int:1, default:0, string:""}; // set all to 0 except the
                                                // array of bits to 1 and
                                                // string to ""
endmodule
```


MikePopoloski:
Huh, interesting that the grammar doesn't include `string` as a simple type even though it seems like it makes sense. The only other use of simple_type is inside casting_type which also explicitly adds string to the list, so this change makes sense to me.

PR #1056, opened 2024-07-12
Fix a bug in the tidy-check UnusedSensitiveSignal

JoelSole-Semidyn:
This PR adds a simple fix to ensure that the body kind is the correct one when accessing the body attributes


likeamahoney:
I think a test case proving the fix is ​​needed

MikePopoloski:
Indeed, in general fixes should come with a unit test included. This change is so minor and straightforward though that I'll merge it anyway.

Issue #1054, opened 2024-07-11
unpredicted behavior when code contains Japanese comments (apparently encoded in Shift+JIS)


udif:
**Describe the bug**
Apparently slang is confused by Japanese comments encoded in Shift+JIS

**To Reproduce**
Unfortunately I don't have yet a minimal example based on non-proprietary code.

**Additional context**
I ran slang against some proprietary code from Japan and it crashed with errors such as `port 'XYZ' has no I/O member declaration`. Please notice that all identifiers are in English, it's just the comments that are in Japanese.
In addition, the error occurs only when they are encoded as Shift+JIS.
If I reencode the same file in UTF-8, there are no errors.
`iconv -f SHIFT-JIS -t UTF-8 jpn_shiftJIS_err.v >jpn_utf8_noerr.v`
When I reencode the UTF8 file back into Shift+JIS, the error returns.
`iconv -t SHIFT-JIS -f UTF-8 jpn_utf8_noerr.v >jpn_back_to_shiftJIS.v`
I suspect that slang fails to detect end of line when the newline falls in such places that due to Shift+JIS encodings would be interpreted as the middle of a UTF-8 codepoint.
Needless to say, the code passes Xcelium (zero warnings) and according to the team who wrote the code also passes SpyGlass.

MikePopoloski:
This is not surprising to me; slang only supports UTF-8 inputs. Giving it text in any other encoding is likely to break something. Extending slang to support other encodings would be a fair amount of work.

udif:
I took a look at `lexer.cpp`, more specifically, `scanLineCommen()`, `scanBlockComment()`, `scanUTF8Char()` and `utf8Decode()`.
Given that:
1. All UTF8 filtering is done inside slang (no external libraries)
2. UTF-8 is not really mentioned in IEEE1800 (At least searching the PDF for `UTF` revealed zero hits).
3. All multibyte UTF-8 codepoints has 1 as their most significant bit.
4. (Specifically for my test case) Legal Shift+JIS only has a 0x0a byte as newline in the 1st byte. The 2nd byte in a multibyte char cannot be 0x0a.

I think that what this implies is that if 0x0a is detected in the middle of a candidate multibyte UTF-8 codepoint, it should be safely identified as 0x0a without any risk of misinterpreting a legal UTF-8 codepoint.
I will try running it in a debugger to see what is the exact cause of the failure, and I'm willing to try writing a PR to fix it, perhaps with a new command line flag if you prefer.

[...]

PR #1059, opened 2024-07-14
Do not skip control characters embedded in malformed UTF-8 characters in comments

udif:
This PR is a partial fix for #1054
Since SystemVerilog files are simple ASCII files, some code bases use international languages in their comments.
Slang expects comments to be in the UTF8 encoding.
If a different coding is used, it is possible that the comment is misinterpreted as an illegal UTF8 string, and effectively skips control characters such as newline.
For this to work, it may be necessary to add the `-Wno-invalid-source-encoding` , since slang will abort any file when the maximum number of lexer errors is encountered (default is 16).

### DIFF
--- a/include/slang/text/CharInfo.h
+++ b/include/slang/text/CharInfo.h
@@ -197,21 +197,12 @@ constexpr const char* utf8Decode(const char* b, uint32_t* c, int* e, int& comput
     // For normal path, this should not be checked
     if (*e) {
         // if error, trim next pointer so that control char is read as next char
-        switch (len) {
-            case 4:
-                if (uc(b[3]) < 0x20)
-                    next--; // last byte in a 4-byte UTF8 is illegal
-                // fall through because earlier bytes might also be illegal
-                [[fallthrough]];
-            case 3:
-                if (uc(b[2]) < 0x20)
-                    next--;
-                [[fallthrough]];
-            case 2:
-                if (uc(b[1]) < 0x20)
-                    next--;
-                // No need for len==1 check because this would not be a unicode
-        }
+        if ((len > 1) && (uc(b[1]) < 0x20))
+            next = b + 1;
+        else if ((len > 2) && (uc(b[2]) < 0x20))
+            next = b + 2;
+        else if ((len > 3) && (uc(b[3]) < 0x20))
+            next = b + 3;
     }
 
     return next;
--- a/include/slang/text/CharInfo.h
+++ b/include/slang/text/CharInfo.h
@@ -202,9 +202,11 @@ constexpr const char* utf8Decode(const char* b, uint32_t* c, int* e, int& comput
                 if (uc(b[3]) < 0x20)
                     next--; // last byte in a 4-byte UTF8 is illegal
                 // fall through because earlier bytes might also be illegal
+                [[fallthrough]];
             case 3:
                 if (uc(b[2]) < 0x20)
                     next--;
+                [[fallthrough]];
             case 2:
                 if (uc(b[1]) < 0x20)
                     next--;
--- a/include/slang/text/CharInfo.h
+++ b/include/slang/text/CharInfo.h
@@ -194,7 +194,6 @@ constexpr const char* utf8Decode(const char* b, uint32_t* c, int* e, int& comput
     *e |= (uc(b[3])) >> 6;
     *e ^= 0x2a; // top two bits of each tail byte correct?
     *e >>= shifte[len];
-    // For normal path, this should not be checked
 
     return next;
 }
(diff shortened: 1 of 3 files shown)


include/slang/text/CharInfo.h
```diff
@@ -194,6 +194,16 @@ constexpr const char* utf8Decode(const char* b, uint32_t* c, int* e, int& comput
     *e |= (uc(b[3])) >> 6;
     *e ^= 0x2a; // top two bits of each tail byte correct?
     *e >>= shifte[len];
+    // For normal path, this should not be checked
```
MikePopoloski, line 197: The comment above describes this function as branchless; I think we should keep it that way. You can put this handling in Lexer::scanUTF8Char instead.

MikePopoloski (commented):
As you mention, I think you probably want to also add a new command line option that doesn't count invalid UTF8 sequences as errors. Otherwise the lexer is going to early-out regardless of whether you hide the warning or not. Or maybe the InvalidUTF8Seq warning should simply not count towards the lexer error count ever, since if the user suppresses they'll be pretty surprised to see them add up and cause a real error anyway.

include/slang/text/CharInfo.h
udif: Added.

udif:
> As you mention, I think you probably want to also add a new command line option that doesn't count invalid UTF8 sequences as errors.

At the moment I settled for using `--max-lexer-errors=1000000`, as I wasn't sure what other cases also trigger lexer errors.
Going now through the Lexer error cases I got this list:
1. Unicode BOM.
2. embedded nulls in source buffer
3. Other ASCII character (<128) not used by the lexer (non-printable)
4. UTF8 in code
5. Embedded null in string literal.
6. Embedded null in line comment
7. Embedded null in block comment
8. And ofcourse, **Illegal UTF8 in comments** (our case).

A cleaner way to deal with it is to check if `invalid-source-encoding` warning is turned off with `-Wno-invalid-source-encoding` and in that case not increment the lexer error count.
The problem is that `Lexer::scanUTF8Char` is also used by `Lexer::lexToken` and `Lexer::lexStringLiteral`, so instead we must do this check explicitly in `Lexer::scanLineComment` and `Lexer::scanBlockComment` and undo the error increment.

I looked into it and from the quick search I did, the `Lexer` class is too low level to get access to the `DiagnosticsEngine::getSeverity()` method. In addition, this solution means that `-Wno-invalid-source-encoding` has an additional side effect other than disabling the warning.

One way to solve this is to add an explicit lexer option to disable these errors from being counted.
What do you think?

MikePopoloski:
slang deliberately as a design decision doesn't change internal behaviors based on whether a particular warning is disabled or not; filtering of warning output happens at the very end of the pipeline.

I think the right thing to do here is to just not count these cases as errors; any time we would issue a warning instead of a hard error we shouldn't then count it as an error in the lexer, since the user can suppress those warnings.

include/slang/text/CharInfo.h
```diff
@@ -194,6 +194,7 @@ constexpr const char* utf8Decode(const char* b, uint32_t* c, int* e, int& comput
...
```
MikePopoloski, line 197: Looks like you left this partial comment here.

udif: Yes, this was left by mistake. I'll remove it.

udif:
> I think the right thing to do here is to just not count these cases as errors

OK, next revision backs out the `errorCount` increment on `scanUTF8Char` but only in case it was called from line or block comments. I also added a test to make sure this works, and the test is conveniently pushed before the change so it can be shown to fail on the existing code by checking out https://github.com/MikePopoloski/slang/pull/1059/commits/0d21f7fe047901ba87d0079fc1d69c28e100b0a6 .

source/parsing/Lexer.cpp
```diff
@@ -1239,7 +1239,10 @@ void Lexer::scanLineComment() {
             advance();
         }
         else {
-            sawUTF8Error |= !scanUTF8Char(sawUTF8Error);
+            auto error = !scanUTF8Char(sawUTF8Error);
+            sawUTF8Error |= error;
+            if (error)
```
MikePopoloski, line 1244: I don't think you need to do this; I would just remove the errorCount increment from scanUTF8Char, since the "error" it issues is actually a suppressible warning, and this is otherwise pretty clunky.

udif: As I wrote above, I did this to minimize the impact of this change. `scanUTF8Char` is also called by other routines and removing `errorCount++` will also affect error counting of other UTF8 errors, mainly on strings.
Assuming this is indeed your intention, I'll prepare another commit for this.

MikePopoloski: Yeah, that was my intent. I think it's fine, the lexer error count thing is perhaps not the most impactful feature anyway.

udif: It's just rolling back the original code.

PR #1062, opened 2024-07-15
In unused warnings attach a range if possible

povik:


MikePopoloski:
I'm not against the idea of this change but I think you want to be more targeted about which warnings you change over to pass the symbol instead of the location. Some symbols have syntax nodes that are very large, like for example a module definition syntax node will span the entire module body, which will look weird if you try to highlight the entire thing instead of just pointing at its name.

povik:
Thanks for taking a look, I know this came missing a cover letter.

> I'm not against the idea of this change but I think you want to be more targeted about which warnings you change over to pass the symbol instead of the location.

OK, good to know you are fine with `addDiag(..., Symbol&)`, and yes, this turned out too indiscriminate about the ranges it attaches also in the downstream linting application which prompted the change. Let me re-open once that is refined.

PR #1064, opened 2024-07-17
[slang] Compute getEffectiveWidth() more accurately for integer constant expressions

Stalker7101:
In this PR, I add more accurate computation of `getEffectiveWidth()` method for constant expression through evaluation, which can be useful for decreasing number of false positive warnings and for future slang API usage.
`--eval-effective-width` option was added for convenient setting up of more accurate effective width computation.
Here are some SystemVerilog constant expression and it's effective width examples with and without  `--eval-effective-width` option.

With `--eval-effective-width` option:
```verilog
// left shift
10'b100  << 3'b100; // effective width 7
10'sb100 << 4'sb100; // effective width 8
// right shift
10'b100  >> 3'b10; // effective width 1
10'sb100 >> 3'sb10; // effective width 2
-3'b100  >> 3'b10; // -4 to 1, effective width 1
-3'sb100 >> 3'sb10; // effective width 2
// right arithmetic shift
-3'sb100 >>> 3'sb10; // effective width 1  
-3'sb100 >>> 3'sb11; // effective width 1 
// add
32'd1  + 32'd2; // effective width 2
32'sd1 + 32'sd2; // effective width 3
// substruct
32'd15  - 32'd10; //  effective width 3
32'sd15 - 32'sd10; // effective width 4
// divide
32'd15  / 32'd5; // effective width 2
32'sd15 / 32'sd5; // effective width 2
// power
32'd2  ** 32'd10; // effective width 11
32'sd2 ** 32'sd10; // effective width 12
```
Without `--eval-effective-width` option:
```verilog
// left shift
10'b100  << 3'b100; // effective width 3
10'sb100 << 4'sb100; // effective width 3
// right shift
10'b100  >> 3'b10; // effective width 3
10'sb100 >> 3'sb10; // effective width 3
-3'b100  >> 3'b10; // -4 to 1, effective width 3
-3'sb100 >> 3'sb10; // effective width 3
// right arithmetic shift
-3'sb100 >>> 3'sb10; // effective width 3  
-3'sb100 >>> 3'sb11; // effective width 3 
// add
32'd1  + 32'd2; // effective width 2
32'sd1 + 32'sd2; // effective width 2
// substruct
32'd15  - 32'd10; //  effective width 4
32'sd15 - 32'sd10; // effective width 4
// divide
32'd15  / 32'd5; // effective width 4
32'sd15 / 32'sd5; // effective width 4
// power
32'd2  ** 32'd10; // effective width 2
32'sd2 ** 32'sd10; // effective width 2
```

Also, here is the example with false positive warning `IntBoolConv` that will happen if `--eval-effective-width` is disabled and `-Weverything` option is enabled:
```verilog
module m();
    initial begin
        if (3'b100 - 2'b11) begin
        end
    end
endmodule
```
Warning:
```
source:4:13: warning: implicit conversion from 'bit[2:0]' to boolean value [-Wint-bool-conv]
        if (3'b100 - 2'b11) begin
            ^~~~~~~~~~~~~~
```


MikePopoloski:
Thanks for the PR. I have some reservations about this change; do you have a motivating example for why you're making it? The IntBoolConv you posted doesn't seem compelling, given that the point of the warning is to point out cases where you're accidentally missing a comparison operator and your example is indeed missing a comparison operator; it's unlikely you would do arithmetic with constant values in such a way and check the result in an if condition.

Otherwise, mechanically the change seems ok, though I will say that this should not be controlled via an option; either this is the right behavior for width checking or it's not. An option like this is very hard to explain to users and they won't know whether they should be using it or not (and probably they won't even know that it exists and will never bother to try it).

Stalker7101:
Sorry for delay.
Pushed commit with deleted option.
PR Motivation: I have student project where I'm going to use `getEffectiveWidth()` method for checking `case` expression labels and selector on width matching. Here is the example of `case` expression with mismatch written on `SystemVerilog`: 
```verilog
module test(input clk);
  const reg [2:0] c = 2;
  const reg [4:0] d = 3;
  wire [4:0] pState;
  wire in1, in2, in3;
  reg out;

  always @ (posedge clk) begin
    case (pState) // selector width = 5
      (2 + 15) : out = in1; // effectiveWidth() = 5 (value = 17)
      (100 << 4) : out = in2; // effectiveWidth() = 11, violation
      (100 >> 2) : out = in2; // effectiveWidth() = 5
      default : out = in3;
    endcase
  end
endmodule
```

Stalker7101:
Changed tests.
Some of them have been changed according to signed / unsigned positive values width computation logic in `IntegerLiteral::getEffectiveWidthImpl()`:
```verilog
// left shift
10'b100  << 3'b100; // effective width 7
10'sb100 << 4'sb100; // effective width 7, changed: don't count unused sign bit
// right shift
10'b100  >> 3'b10; // effective width 1
10'sb100 >> 3'sb10; // effective width 1, changed: don't count unused sign bit
-4'b0100  >> 3'b10; // effective width 2, changed
-4'sb0100 >> 3'sb10; // effective width 2, changed
// right arithmetic shift
-3'sb100 >>> 3'sb10; // effective width 1  
-3'sb100 >>> 3'sb11; // effective width 1
-3'sb100 >>> 3'sb10; // effective width 1, new test
-3'sb100 >>> 3'sb11; // effective width 1, new test
// add
32'd1  + 32'd2; // effective width 2
32'sd1 + 32'sd2; // effective width 2, changed: don't count unused sign bit
// subtract
32'd15  - 32'd10; //  effective width 3
32'sd15 - 32'sd10; // effective width 3, changed: don't count unused sign bit
// divide
32'd15  / 32'd5; // effective width 2
32'sd15 / 32'sd5; // effective width 2
// power
32'd2  ** 32'd10; // effective width 11
32'sd2 ** 32'sd10; // effective width 11, changed: don't count unused sign bit
```

MikePopoloski:
Sorry it took me a while to get back to this. Thank you for the work you put into the PR, but I will say that without a motivating example I'm not convinced that we want this change to effect the existing uses of getEffectiveWidth in terms of providing warnings. If the main goal is to use this as an API then it should be relatively easy to write the same logic as a separate ASTVisitor and just use that locally in your project instead of changing this behavior within slang.

MikePopoloski:
Going to close this; feel free to reopen if you want to add more context.

PR #1066, opened 2024-07-18
 Prevent superflous separators after node removal 

sgizler:
Problem:
When you call `SyntaxRewriter::remove()` on node stored in SeparatedSyntaxList, node will disappear leaving superfluous separator resulting in syntax error.

This PR skips separators left by node removal.

### DIFF
--- a/tests/unittests/parsing/VisitorTests.cpp
+++ b/tests/unittests/parsing/VisitorTests.cpp
@@ -250,6 +250,39 @@ endmodule
 )");
 }
 
+TEST_CASE("Remove node from comma-separated list") {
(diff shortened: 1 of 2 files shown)


tests/unittests/parsing/VisitorTests.cpp
```diff
@@ -250,6 +250,40 @@ endmodule
 )");
 }
 
+#include <iostream>
```
MikePopoloski, line 253: What's the need for the iostream include here? It's a heavy header, I'd rather not include it if it's not absolutely necessary.

sgizler: it is leftover from debugging. Removed it

PR #1068, opened 2024-07-18
Added a script to filter messages and summarize slang logfiles

udif:
Here is a small script I wrote when I needed to filter a log with >5000 errors and warnings.
The script has two modes:

In the first mode (--summary), it produced a summary of all errors and warnings, where all errors and warnings of each type are counted and listed in a table. error messages that are unique due to a specific identifier mentioned in the error are also merged by replacing any identifier found in single quotes by the word `'something'`, so you can see error messages such as:
```
error: redefinition of <something>: 7
```
Indicating 7 errors where 7 different identifiers have been redefined.

The 2nd mode is used to list specific subsets of errors and warnings. you can include and exclude specific files or directories (using a regex), or you can include or exclude specific error/warning types.
You cannot mix include and exclude options because include options starts from an empty list, while exclude options starts from the full list.
When listing messsages, the script is smart enough (I hope) to understand which lines in the log belong to which error or warning.
e.g.A full error may be comprised of:
* Optional `'in file'` line
* Optional `'  in instance'` lines
* Optional one or more  1-3 `note:` lines
  * if `note:` is at the beginning of the line,  this is a 1-line note.
  * If it is in the form `<filename>: note:` This would be a 3-line note:
* A 3-line `error:` / `warning:` 
* Optional one or more 1-3 `note:` lines (see above).

### DIFF
--- a/tools/scripts/slang_log.py
+++ b/tools/scripts/slang_log.py
@@ -1,14 +1,16 @@
 #!/usr/bin/env python3
 
-import re
 import argparse
+import re
 from collections import defaultdict
+
 errors = 0
 warnings = 0
 
+
 def parse_log_file(file_path):
     global errors, warnings
-    with open(file_path, 'r') as file:
+    with open(file_path, "r") as file:
         lines = file.readlines()
 
     log_entries = []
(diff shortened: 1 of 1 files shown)


MikePopoloski:
Thanks for the contribution, but I'm wondering if it would make more sense to add an option to slang to output diagnostics in a structured (e.g. json) format so that the parsing / heuristics aren't necessary. I'd be wary of some inadvertent change on my end breaking the script (especially since there are no tests for it).

udif:
A JSON output file would be useful, but would probably come along the text output, and generated in parallel, so you wouldn't need to run slang twice. If you think JSON is the way to go, I will adapt and use it.
I can also see some future enhancements for CI, such as running `git blame` on problematic lines, and collecting errors/warnings by users so they could be emailed only their relevant log.
Tests could be added, but it would probably make more sense to use a separate python based unit testing package.

MikePopoloski:
I think the text diagnostic output could be included in the JSON as well if you want it; like the message and the context / snippet line. I do think that if we want the diagnostics to be read programmatically then we should just use a machine readable format.

MikePopoloski:
I opened #1073 to track the discussed feature.

PR #1069, opened 2024-07-22
Fix excessive memory usage caused by transforming tree multiple times

sgizler:
Problem: Transformed tree returned by SyntaxRewriter holds `shared_ptr` to original one, preventing it from being freed as long as its "child" is in scope. This becomes a problem when you transform tree iteratively a lot. 

This PR:
- Fixes incomplete deep-clone routines of token related stuff.
- Replaces shallow clones of tokens in SyntaxRewriter with deep ones so transformed tree is not longer dependent on old one.
- Adds option for not storing this problematic `shared_ptr`. Since there might be cases which depend on that behavior
  (e.g user's handler copies something shallowly from old tree), we don't make it default. Test cases that are fine with it, were made to use `holdParentPtr=false`


include/slang/syntax/SyntaxVisitor.h
```diff
@@ -109,6 +109,10 @@ class SyntaxRewriter : public SyntaxVisitor<TDerived> {
     ///
     /// @return if no changes are requested, returns the original syntax tree.
     /// Otherwise, the changes are applied and the newly rewritten syntax tree is returned.
+    ///
+    /// By default transformed tree holds shared_ptr to original one, preventing it from
+    /// being freed as long as its child is in scope. If you don't depend on this behaviour,
+    /// you can pass false as @a shareLifetimes, in order to save some memory.
```
MikePopoloski, line 115: "shareLifetimes" isn't an argument here; I assume you changed that at some point to the member variable?

MikePopoloski (commented):
I don't love having the two different modes for storing the parent pointer vs not doing it. Is there a good reason at this point to have the old behavior? Maybe we can just remove it and let people save the trees themselves if they need to.

include/slang/syntax/SyntaxVisitor.h
sgizler: Yeah, indeed. I guess I will remove this comment entirely since it duplicates docs of `setHoldParentPtr()` method

sgizler:
> I don't love having the two different modes for storing the parent pointer vs not doing it. Is there a good reason at this point to have the old behavior?  Maybe we can just remove it and let people save the trees themselves if they need to.

There are two tests that depend on it, and would access freed memory otherwise:
- ["Shallow Clone"](https://github.com/antmicro/slang/blob/9e777123719e1cdc96d5837de62ab364fafa59f5/tests/unittests/parsing/VisitorTests.cpp#L574-L599)
- ["Syntax rewriting with metadata updates"](https://github.com/antmicro/slang/blob/fix_leak/tests/unittests/parsing/VisitorTests.cpp#L648-L731)
so I thought that there might be some code in wild that would be broken by this.

If you are fine with this, I can:
- Remove `holdParentPtr` related logic, and make transform to not save parent pointer
- Make setter like `SyntaxTree::setParentPtr()` that could be used to emulate old behavior.

Alternatively we could make transform save parent pointer, but add method for removing it.

MikePopoloski:
Yeah, I think it's ok to break this. I would just remove the parent pointer thing from SyntaxTree entirely, which simplifies things a bit. In the examples you provided the problem can be fixed by just using a new variable instead of overwriting the old one.

sgizler:
>  I would just remove the parent pointer thing from SyntaxTree entirely, which simplifies things a bit.
done

PR #1083, opened 2024-08-07
Introduce IEEE Std 1497-2001 Standard Delay Format (SDF) parsing

likeamahoney:
Good day everyone! In order to be able to support restrictions from 32 LRM section and annotate the `SystemVerilog` code using SDF (the last and only standard is `IEEE Std 1497-2001`) constructs I implemented the `LL(2)` parser using the `slang` infrastructure.

I encourage testing on individual `SDF` files for feedback, as the current implementation has been tested on various open-source `SDF` files.

The new `--sdf` option in the slang infrastructure allows for feeding slang one or more `SDF` files and also directory with `SDF` files. The implementation supports the full `SDF `grammar (including the syntax of conditional and timing check `SDF` expressions which is significantly differs from `SystemVerilog` expressions), handling all possible syntax errors efficiently. Example of usage:

```shell
./slang --sdf=test1.sdf,test2.sdf tests.sv  // slang always expects verilog/systemverilog files as input
```

What has not been done at the moment:

- now only translation into a syntax tree has been implemented (it seems to me that there is no need to translate into a separate AST but only to supplement the existing `SystemVerilog` nodes of the `slang` AST - but it questionable)  
- It is not clear how large the number of grammar tests should be
- processing of `SDF DELAY`s constructs - is it necessary to directly put values ​​from SDF to `SystemVerilog specify` AST? or it should be done in simulation (possibly using some pseudo `SDF` annotator - which defined at LRM 32.3 section), and nothing should be changed in AST? LRM specifies no possible restrictions with `SDF DELAY`s
- `SDF TIMINGCHECK`s constructs - `slang` timing check system tasks currently are not annotated - it is future work
- `SDF LABEL`s constructs - same questions as `SDF DELAY` section - is it need to propagate it's values directly in AST?
- possible changes in `$sdf_annotate` processing - is required or not? 

I suppose `SDF TIMINGENV` constructs should be ignored by `slang` as it said at LRM 32.3 section:

> An SDF file can contain many constructs that are not related to specify path delays, specparam values,
timing check constraint values, or interconnect delays. An example is any construct in the TIMINGENV
section of the SDF file. All constructs unrelated to SystemVerilog timing shall be ignored without any
warnings issued.

An opened question pertains to annotating unique delays between source/load pairs using the `INTERCONNECT` construct. How this warning should be processed, either by slang or the `SDF` annotator, remains unclear:

> In the case of multisource nets, unique delays can be annotated between each source/load pair using the
INTERCONNECT construct. When annotating this construct, the SDF annotator shall find the source port and
the load port; and if both exist, it shall annotate an interconnect delay between the two. If the source port is
not found or if the source port and the load port are not actually on the same net, then a warning is issued,
but the delay to the load port is annotated anyway


MikePopoloski:
Thanks for the PR. I'm curious about the approach of fully reusing the Lexer, Preprocessor, Parser, and SyntaxTree classes from Verilog to also parse SDF files. It results in a lot of additional branching along the way since syntax differs significantly between the two languages. It also leads to some weird side effects that I don't think you'd want, like for example it seems you could probably use SystemVerilog defines and macro expansion in SDF files? Or at the lexical level, the Lexer will accept a bunch of token kinds that aren't actually allowed in SDF files.

I think it would make more sense to create separate copies of these classes, that target just parsing SDF. If there is useful overlap between the two you can factor them into a common base; the ParseBase class kind of does that already for the parser, for example. SDF files are much simpler than SystemVerilog so the resulting classes would be much more compact.

MikePopoloski:
Going to close this out; we haven't needed an SDF parser thus far and the PR proposed here is too invasive. Probably it should live in a downstream project until there's good reason to bring it in here.

likeamahoney:
> Going to close this out; we haven't needed an SDF parser thus far and the PR proposed here is too invasive. Probably it should live in a downstream project until there's good reason to bring it in here.

Should I reopen it when I will reimplement this in terms of comment above? Will it be less invasive if it will be implemented like that:

> I think it would make more sense to create separate copies of these classes, that target just parsing SDF. If there is useful overlap between the two you can factor them into a common base; the ParseBase class kind of does that already for the parser, for example. SDF files are much simpler than SystemVerilog so the resulting classes would be much more compact.

MikePopoloski:
I think for now it should be implemented in a downstream repo and then we can reintegrate it here if we need it later.

Issue #522, opened 2022-04-17
Finish support for virtual interfaces


MikePopoloski:

MikePopoloski:
Remaining tasks:
```
It shall be illegal to assign an interface instance to a virtual interface if there is a defparam to a parameter of
that interface instance or interface hierarchy and that defparam statement is declared outside the interface.
```

```
Although an interface may contain hierarchical references to objects outside its body or ports that reference
other interfaces, it shall be illegal to use an interface containing those references in the declaration of a
virtual interface.
```

- Allow virtual interface assignments inside unpacked array concatenations
- Figure out what happens if a configuration changes the target of the virtual interface assignment

Issue #949, opened 2024-04-07
Add caching of identical instance bodies


MikePopoloski:
Support caching instance bodies and skipping over identical copies when elaborating, to save time. Expose this functionality so that downstream consumers of the AST can easily do the same.

Note that the tricky detail here is figuring out how to handle hierarchical references into and out of the cached instances.

yanggeorge:
It is better that elabarated AST only show one instance of identical instances.

likeamahoney:
My hash suggestion - https://github.com/MikePopoloski/slang/pull/1088#discussion_r1761208918

MikePopoloski:
This can now be turned on with a compilation flag, with a caveat that it can cause some multi-driven errors to not be issued correctly. Things to handle before it's fully robust:
- Assignments through interface ports need to be reapplied to specific instances being connected
- Any upward hierarchical references that extend out of the cached instance need to be re-resolved, and the resolved types must match, otherwise we can't use the cache entry
- Downward hierarchical references into such instances need to be accounted for
- Bind directives needs to be accounted for
- Defparams need to be accounted for
- Configuration rules need to be accounted for

MikePopoloski:
This is complete.

PR #1088, opened 2024-08-12
[slang] Add virtual interface declaration checks

likeamahoney:
In this PR i add some virtual interface checks such as:
 - hierarchical references usage
 - interface ports usage

That checks related to part of https://github.com/MikePopoloski/slang/issues/522

> Although an interface may contain hierarchical references to objects outside its body or ports that reference
other interfaces, it shall be illegal to use an interface containing those references in the declaration of a
virtual interface.

### DIFF
--- a/scripts/diagnostics.txt
+++ b/scripts/diagnostics.txt
@@ -489,6 +489,7 @@ error TimingCheckEventEdgeRequired "'{}' event argument requires an edge trigger
 error NoChangeEdgeRequired "trigger for first argument to '$nochange' must be posedge or negedge only"
 error PathPulseInvalidPathName "PATHPULSE$ path '{}' must be of the form 'source$dest'"
 error VirtualInterfaceIfaceMember "virtual interfaces cannot be used as the type for interface items"
+error VirtualInterfaceHierRef "Declaration of virtual interface should not contain hierarchical references to objects outside its body"
 error DefparamBadHierarchy "defparam in a hierarchy in or under a generate block, array of instances, or bind instantiation cannot change a parameter value outside that hierarchy"
 error BindTargetPrimitive "cannot bind a primitive"
 error BindUnderBind "cannot bind an instance underneath the scope of another bind instance"
--- a/scripts/diagnostics.txt
+++ b/scripts/diagnostics.txt
@@ -490,6 +490,7 @@ error NoChangeEdgeRequired "trigger for first argument to '$nochange' must be po
 error PathPulseInvalidPathName "PATHPULSE$ path '{}' must be of the form 'source$dest'"
 error VirtualInterfaceIfaceMember "virtual interfaces cannot be used as the type for interface items"
 error VirtualInterfaceHierRef "Declaration of virtual interface should not contain hierarchical references to objects outside its body"
+error VirtualInterfaceIfacePort "Declaration of virtual interface should not contain ports that reference other interfaces"
 error DefparamBadHierarchy "defparam in a hierarchy in or under a generate block, array of instances, or bind instantiation cannot change a parameter value outside that hierarchy"
 error BindTargetPrimitive "cannot bind a primitive"
 error BindUnderBind "cannot bind an instance underneath the scope of another bind instance"
--- a/source/ast/types/AllTypes.cpp
+++ b/source/ast/types/AllTypes.cpp
@@ -1129,19 +1129,29 @@ const Type& VirtualInterfaceType::fromSyntax(const ASTContext& context,
 
     auto loc = syntax.name.location();
     auto& defSym = def->as<DefinitionSymbol>();
-    auto& iface = InstanceSymbol::createVirtual(context, loc, def->as<DefinitionSymbol>(),
+    auto& iface = InstanceSymbol::createVirtual(context, loc, defSym,
                                                 syntax.parameters);
 
     // To control duplicates
     SmallSet<const DefinitionSymbol*, 4> visited;
     // Find cached instance hierarchical identifiers
     std::function<std::optional<const InstanceSymbol*>(const Scope& scope)> findInst =
-        [&defSym, &findInst, &visited](const Scope& scope) -> std::optional<const InstanceSymbol*> {
+        [&defSym, &findInst, &visited, &context,
+         &iface](const Scope& scope) -> std::optional<const InstanceSymbol*> {
         for (auto& name : scope.getNameMap()) {
             if (auto inst = name.second->as_if<InstanceSymbol>()) {
                 auto instDef = &inst->getDefinition();
-                if (instDef == &defSym && inst->body.hierIdentifiers.has_value())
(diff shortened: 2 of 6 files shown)


source/ast/types/AllTypes.cpp
```diff
@@ -1131,6 +1181,9 @@ const Type& VirtualInterfaceType::fromSyntax(const ASTContext& context,
     auto& iface = InstanceSymbol::createVirtual(context, loc, def->as<DefinitionSymbol>(),
                                                 syntax.parameters);
 
+    InterfaceDefVisitor iDVisitor(iface, context);
```
MikePopoloski, line 1184: We really don't want to rerun this checker on every virtual interface type declaration -- ideally we only run it once per unique interface instantiation. My thinking was to track hierarchical lookups more closely, registering them on instances as they pass through them. Primarily this would be to support caching of instance bodies to avoid unnecessary elaboration work, but as a side effect would make this kind of checking much easier.

source/ast/types/AllTypes.cpp
```diff
@@ -31,6 +32,55 @@ using namespace slang;
...
 
+class InterfaceDefVisitor : public ASTVisitor<InterfaceDefVisitor, true, true> {
+public:
+    InterfaceDefVisitor(const InstanceSymbol& iface, const ASTContext& context) :
+        iface(iface), context(context) {};
+    template<typename T>
+    void handle(const T& symbol) {
+        if constexpr (std::is_base_of_v<Symbol, T>) {
+            if (symbol.kind == SymbolKind::Instance) {
+                const auto* inst = symbol.template as_if<InstanceSymbol>();
```
MikePopoloski, line 43: This seems wrong; instantiating an interface inside another interface doesn't make it a port, so we shouldn't error on it.

scripts/diagnostics.txt
```diff
@@ -543,6 +545,7 @@ note InfoTask "$info encountered{}"
 note NoteComparisonReduces "comparison reduces to ({} {} {})"
 note NoteCommonAncestor "common ancestor here violates that constraint"
 warning explicit-static StaticInitializerMustBeExplicit "initializing a static variable in a procedural context requires an explicit 'static' keyword"
+note VirtualInterfaceDecl "virtual interface declared here"
```
MikePopoloski, line 548: In general we don't need to keep adding new notes that are very slight variations of existing ones; the more general "declared here" note is fine for these purposes.

likeamahoney: removed

source/ast/types/AllTypes.cpp
likeamahoney: instance hierarchical identifiers caching was added

source/ast/types/AllTypes.cpp
likeamahoney: still in progress

likeamahoney: in the first version of the solution, I misunderstood LRM in this term, sorry

likeamahoney: add that check in the last commit

source/ast/types/AllTypes.cpp
```diff
@@ -1128,8 +1128,64 @@ const Type& VirtualInterfaceType::fromSyntax(const ASTContext& context,
...
+    // To control duplicates
+    SmallSet<const DefinitionSymbol*, 4> visited;
+    // Find cached instance hierarchical identifiers
+    std::function<std::optional<const InstanceSymbol*>(const Scope& scope)> findInst =
+        [&defSym, &findInst, &visited, &context,
+         &iface](const Scope& scope) -> std::optional<const InstanceSymbol*> {
+        for (auto& name : scope.getNameMap()) {
+            if (auto inst = name.second->as_if<InstanceSymbol>()) {
+                auto instDef = &inst->getDefinition();
+                if (instDef == &defSym && inst->body.hierIdentifiers.has_value()) {
```
MikePopoloski, line 1143: This is definitely not correct; just because the definition is the same doesn't mean the hierarchical identifiers will be the same, since behavior can vary on the parameters of the instance controlling generate blocks in the child.

This is a generally hard problem; my thinking in the past was that I would keep track of found hierarchical identifiers during Lookup and caching based on a more complicated cache key that takes into account parameters, interface ports, any active configurations, etc. Tracking for that is here: https://github.com/MikePopoloski/slang/issues/949

likeamahoney: Hi, I have an idea for a hash - is it can be a `std::hash` of `std::string` consisting of an instance kind (for example `interface` or `module`), instance type (for example - module type identifier) and a set of parameter values ​​with their identifiers? 

For example for such module:

```verilog
module mem_model #(
 parameter ADDR_WIDTH=8;
 parameter DATA_WIDTH=32;)
 (clk, addr, data);

 input  clk;
 input  [ADDR_WIDTH-1:0] addr;
 output [DATA_WIDTH-1:0] data;
 .....
 .....
endmodule
```

and it's instances:

```verilog
//Creates mem_1 instance with default addr and data widths.
mem_model mem_1 (.clk(clk),
                 .addr(addr_1),
                 .data(data_1));

//Creates mem_2 instance with addr width = 4 and data width = 8.
mem_model #(4,8) mem_2 (.clk(clk),
                        .addr(addr_2),
                        .data(data_2));

//Creates mem_3 instance with addr width = 32 and data width = 64.
mem_model #(32,64) mem_3 (.clk(clk),
                          .addr(addr_3),
                          .data(data_3));

//Creates mem_4 instance with default addr width and data width = 64.
mem_model #(DATA_WIDTH=64) mem_4 (.clk(clk),
                                 .addr(addr_3),
                                 .data(data_3));
```

hash string can be:

```с++
"module mem_model(ADDR_WIDTH=8, DATA_WIDTH=32)" // for mem_1

"module mem_model(ADDR_WIDTH=4, DATA_WIDTH=8)" // for mem_2

"module mem_model(ADDR_WIDTH=32, DATA_WIDTH=64)" // for mem_3

"module mem_model(ADDR_WIDTH=8, DATA_WIDTH=64)" // for mem_4
```

What do you think about this suggestion?

MikePopoloski: There's no reason for it to be a string; you can just hash the relevant fields. This diff from a long time ago removed an InstanceCache that I had in slang at one point: https://github.com/MikePopoloski/slang/commit/0892da30760454a58bdfbb9a2d1d40d728643e36#diff-18c21a73a1abf2b9e1aa72758848d9bceee11a575bfab7d762dfc51217ec26c3

That's a good start but it also needs to work correctly with hierarchical references into and out of cached instances, which is very tricky and something I need to think more carefully about (which is one of the reasons why I removed this back in the day).

MikePopoloski:
Superseded by 76e5ec487d1711c16c50daa4130d1ff6c88dd4cd

PR #1089, opened 2024-08-13
Avoid internal symbol in tidy port suffix rule

povik:
Do not look at the internal symbol, scrutinize the name of the port directly. This way we don't crash on explicit ports, as those have the internal symbol pointer set to null.


povik:
Force pushed to fix the included test

PR #1090, opened 2024-08-13
[slang] Add a check for local variables same as formal args declaration

likeamahoney:
Implement a check for restriction from LRM 16.10 section:

> When a local variable is a formal argument of a sequence declaration, it is illegal to declare the variable, as
> shown in the following example:
> ```verilog
> sequence sub_seq3(lv);
> int lv; // illegal because lv is a formal argument
> (a ##1 !a, lv = data_in) ##1 !b[*0:$] ##1 b && (data_out == lv);
> endsequence
> ```

### DIFF
--- a/source/ast/expressions/MiscExpressions.cpp
+++ b/source/ast/expressions/MiscExpressions.cpp
@@ -1120,9 +1120,6 @@ Expression& AssertionInstanceExpression::makeDefault(const Symbol& symbol) {
     auto result = comp.emplace<AssertionInstanceExpression>(*type, symbol, body,
                                                             /* isRecursiveProperty */ false, range);
     result->localVars = localVars.copy(comp);
-    // Check LRM 16.10 section restriction:
-    // > It's illegal to declare a local variable if it is a formal argument of a sequence
-    // declaration.
     return *result;
 }
 
--- a/source/ast/expressions/MiscExpressions.cpp
+++ b/source/ast/expressions/MiscExpressions.cpp
@@ -739,6 +739,12 @@ static const AssertionExpr& bindAssertionBody(const Symbol& symbol, const Syntax
                                               AssertionInstanceDetails& instance,
                                               SmallVectorBase<const Symbol*>& localVars) {
     auto createLocals = [&](auto& syntaxType) {
+        const auto* seq = symbol.as_if<SequenceSymbol>();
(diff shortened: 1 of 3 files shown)


source/ast/expressions/MiscExpressions.cpp
```diff
@@ -1101,6 +1101,22 @@ Expression& AssertionInstanceExpression::makeDefault(const Symbol& symbol) {
     auto result = comp.emplace<AssertionInstanceExpression>(*type, symbol, body,
                                                             /* isRecursiveProperty */ false, range);
     result->localVars = localVars.copy(comp);
+    // Check LRM 16.10 section restriction:
+    // > It's illegal to declare a local variable if it is a formal argument of a sequence
+    // declaration.
+    if (!localVars.empty() && !formalPorts.empty() && symbol.kind == SymbolKind::Sequence) {
```
MikePopoloski, line 1107: I don't think this should be restricted to just sequences. It applies to properties as well. Also I think you can place this in the createLocals() function up above in this file. There's already a map of local var names and duplicate checking there, so you can avoid the O(n^2) algorithm here.

source/ast/expressions/MiscExpressions.cpp
```diff
@@ -1101,6 +1101,22 @@ Expression& AssertionInstanceExpression::makeDefault(const Symbol& symbol) {
...
+        for (const auto* port : formalPorts) {
+            // Map port symbol and local variable symbol by name
+            auto isPort = [&](const Symbol* localVarSym) {
+                return port->name == localVarSym->name;
+            };
+            if (auto it = std::find_if(localVars.begin(), localVars.end(), isPort);
+                it != localVars.end()) {
+                auto& diag = context.addDiag(diag::SequenceLocalVarFormalArg, (*it)->location);
```
MikePopoloski, line 1115: Rather than having a separate diagnostic for this, you can reuse the Scope::reportNameConflict method to get standardized reporting of the duplicate. Note its use in the createLocals function mentioned earlier.

likeamahoney: fixed

source/ast/expressions/MiscExpressions.cpp
likeamahoney: fixed

PR #1093, opened 2024-08-15
[slang] Fix deffered assertion function call check

likeamahoney:
There is an example from LRM which not marked as `illegal`:

```verilog
function int error_type (int opcode);
    func_assert: assert (opcode < 64) else $display("Opcode error.");
    if (opcode < 32)
        return (0);
    else
        return (1);
endfunction

always_comb begin : b1
    a1: assert #0 (my_cond) else
        $error("Error on operation of type %d\n", error_type(opcode));
    a2: assert #0 (my_cond) else
        error_type(opcode); // returns non-void
...
end
```

`error_type` is not system call and returns non-void value.

I suppose that is valid case when non-void function can be used inside deferred assertions.

In confirmation of my words, a comment was left in the `slang` (`source/ast/Statements.cpp`) that only the type of system calls is checked:

```cpp
    // The subroutine being called has some restrictions:
    // - No output or inout arguments
    // - If a system call, must be a task or void returning
    // - Any ref arguments cannot reference automatic or dynamic variables
    auto& call = stmt.as<ExpressionStatement>().expr.as<CallExpression>();
    AssertionExpr::checkAssertionCall(call, context, diag::DeferredAssertOutArg,
                                      diag::DeferredAssertAutoRefArg, diag::DeferredAssertNonVoid,
                                      stmt.sourceRange);
```

### DIFF
--- a/include/slang/ast/ASTContext.h
+++ b/include/slang/ast/ASTContext.h
@@ -194,11 +194,8 @@ enum class SLANG_EXPORT ASTFlags : uint64_t {
 
     /// AST binding is for a bind instantiation (port connection or param value).
     BindInstantiation = 1ull << 45,
-
-    /// The expression is inside sequence.
-    InsideSequence = 1ull << 46
 };
-SLANG_BITMASK(ASTFlags, InsideSequence)
+SLANG_BITMASK(ASTFlags, BindInstantiation)
 
 // clang-format off
 #define DK(x) \
--- a/include/slang/ast/ASTContext.h
+++ b/include/slang/ast/ASTContext.h
@@ -193,9 +193,12 @@ enum class SLANG_EXPORT ASTFlags : uint64_t {
     DisallowUDNT = 1ull << 44,
 
     /// AST binding is for a bind instantiation (port connection or param value).
-    BindInstantiation = 1ull << 45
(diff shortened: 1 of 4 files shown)


include/slang/ast/ASTContext.h
```diff
@@ -193,9 +193,12 @@ enum class SLANG_EXPORT ASTFlags : uint64_t {
     DisallowUDNT = 1ull << 44,
 
     /// AST binding is for a bind instantiation (port connection or param value).
-    BindInstantiation = 1ull << 45
+    BindInstantiation = 1ull << 45,
+
+    /// The expression is inside sequence.
+    InsideSequence = 1ull << 46
```
MikePopoloski, line 199: You shouldn't need an ASTFlag for this; the checkAssertionCall function is called directly. I would just add another parameter that indicates whether the function return type should be checked or not.

likeamahoney: fixed

likeamahoney:
last CI sanitize job was canceled due to reasons beyond my control

likeamahoney:
> last CI sanitize job was canceled due to reasons beyond my control

likeamahoney:
Sorry, accidentally closed PR, reopened now

PR #1094, opened 2024-08-15
[slang] Fix recursive property bug

likeamahoney:
An LRM example which fails on `slang`:

```verilog
property p3(p, bit b, abort);
    (p and (1'b1 |=> p4(p, b, abort)));
endproperty

property p4(p, bit b, abort);
    accept_on(b) reject_on(abort) p3(p, b, abort);
endproperty
````

It is legal because `p4` invoked as rhs of non-overlap implication.

### DIFF
--- a/source/ast/expressions/MiscExpressions.cpp
+++ b/source/ast/expressions/MiscExpressions.cpp
@@ -987,7 +987,8 @@ Expression& AssertionInstanceExpression::fromLookup(const Symbol& symbol,
     ASTContext bodyContext(*symbolScope, LookupLocation::max);
     bodyContext.assertionInstance = &instance;
     // Propagate previously founded time advance specs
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: source/ast/expressions/MiscExpressions.cpp [hunk 1]
```diff
@@ -986,6 +986,8 @@ Expression& AssertionInstanceExpression::fromLookup(const Symbol& symbol,
 
     ASTContext bodyContext(*symbolScope, LookupLocation::max);
     bodyContext.assertionInstance = &instance;
+    // Propagate previously founded time advance specs
+    bodyContext.flags |= ASTFlags::PropertyTimeAdvance;
```
MikePopoloski, line 990: This should probably only set the flag if the parent context had the flag set right?

likeamahoney: fixed

PR #1095, opened 2024-08-15
[slang] Finish implementation of matched/triggered sequence method restrictions.

likeamahoney:
Closes https://github.com/MikePopoloski/slang/issues/524

After researching full LRM in terms of `matched`/`triggered` sequence methods correct usage I considered that such checks were not implemented in `slang`:

1 (matched):
> ... sequence triggered method should not be used in the right-hand side of a continuous assignment or of an assignment in an always_comb procedure.

example:
```verilog
checker my_check(input logic clk, output logic a, output logic b);
sequence s; 1; endsequence

assign b = s.triggered; // illegal 

always_comb a = s.triggered;  // illegal
endchecker

module m(input logic clk, output logic a, output logic b);
my_check mc(clk, a, b);
endmodule
```

2 (triggered):

> There are additional restrictions when passing local variables into an instance of a named sequence to which triggered is applied: In the declaration of the named sequence, the formal argument to which the local variable is bound shall not be referenced before it is assigned.

example:
```verilog
sequence sub_seq2(lv);
(lv, lv = 1);  // violation
endsequence
sequence seq2a;
int v1; ##1 sub_seq2(v1).triggered ##1 (1 == v1);
// v1 is now bound to lv
endsequence

module m();
assert property(seq2a);
endmodule
```

3 (matched):

> The method matched is used to detect the end point of one sequence (the source sequence) referenced in a multiclocked sequence (the destination sequence). It can only be used in sequence expressions.

example:

```verilog
program check;

sequence e1;
1;
endsequence

sequence e2;
1;
endsequence

initial begin
wait (e1.matched || e2.matched); // violation
if (e1.matched) // violation
$display("e1 passed");
if (e2.matched) // violation
$display("e2 passed");
end
endprogram
```

Also the check that only sample named value expressions are used when passing local variables into an instance of a named sequence has been clarified (now only calls to `triggered` methods are checked), since LRM says that this only applies to them:

> There are additional restrictions when passing local variables into an instance of a named sequence to which
> triggered is applied:
> —Local variables can be passed in only as entire actual arguments, not as proper subexpressions of
> actual arguments.

The last one is `triggered` method circular dependencies restriction check is not implemented for now.

`triggered` and `matched` methods clock check should be implemented in terms of https://github.com/MikePopoloski/slang/pull/1027


MikePopoloski:
Ok, I finally got some time to look at this more and I think most of this diff is not what we want.

1. "sequence triggered method should not be used in the right-hand side of a continuous assignment or of an assignment in an always_comb procedure" -- this use of "should" along with it being mentioned off hand in a random section from checkers means that it's not actually a rule enforced by the compiler, just the LRM saying you probably shouldn't do this or the behavior will be weird. No commercial tool enforces this either, so I think this is not a real rule and slang should not error on it.

2. The implementation for use-before-assigned works in simple cases but the rules for local variable flow are actually complicated and probably need a dedicated analysis pass.

3. Yes, not using matched outside sequences is a simple check that we need to add.

4. The changes to make local variable argument rules only apply to triggered are not correct. The LRM says "Local variables can be passed into an instance of a named sequence to which matched is applied. The same
restrictions apply as in the case of triggered."

5. The circular dependencies check is already implemented so we don't have to worry about that.

There's also this rule which I believe is missing for triggered:
"It shall be considered an error to invoke this method outside a sequence context on
sequences that treat their formal arguments as local variables. A sequence treats its formal argument as a
local variable if the formal argument is used as an lvalue in operator_assignment or inc_or_dec_expression
in sequence_match_item."

I'll look into implementing these as part of the new analysis pass.

PR #1097, opened 2024-08-20
[slang] Fix UDP overlap bugs

likeamahoney:
That PR fixes some types of slang udp overlap analysis bugs.

Bad `n` and `p` edge analysis:

```verilog
// Slang do not errors on it but there is overlap:
primitive d_edge_ff (q, clock, data);
  output q; reg q;
  input clock, data;
  table
    // clock data q q+
    p ? : ? : 1 ;
    (x1) ? : ? : 0;  //  it overlaps a  "p ? : ? : 1" row with different output
    n ? : ? : 1 ;
    (1x) ? : ? : 0; // it overlaps a  "n ? : ? : 1" row with different output 
  endtable
endprimitive
```
And also when one row syntax replaces another in a `BitTrie`:

```verilog
// Slang do not errors on it but it obviously overlap:
primitive d_edge_ff (q, clock, data, data1);
  output q; reg q;
  input clock, data, data1;
  table
    // clock data q q+
    (?0) ? 1 : ? : 1 ;
    (0?) ? 1 : ? : 1 ;
    (01) ? 1 : ? : 0 ; // overlaps "(0?) ? 1 : ? : 1" row with different output
  endtable
endprimitive

// Moreover, slang errors about such an overlap (if 3 and 1 rows  in previous example will be swapped),
// but the order of the rows in the table should not affect the error due to LRM.
primitive d_edge_ff (q, clock, data, data1);
  output q; reg q;
  input clock, data, data1;
  table
    // clock data q q+
    (01) ? 1 : ? : 0 ;
    (?0) ? 1 : ? : 1 ;
    (0?) ? 1 : ? : 1 ;
  endtable
endprimitive
```

### DIFF
--- a/source/ast/symbols/MemberSymbols.cpp
+++ b/source/ast/symbols/MemberSymbols.cpp
@@ -9,6 +9,7 @@
 
 #include "../FmtHelpers.h"
 #include "fmt/core.h"
+#include <set>
 
 #include "slang/ast/ASTSerializer.h"
 #include "slang/ast/ASTVisitor.h"
@@ -766,13 +767,14 @@ void PrimitivePortSymbol::serializeTo(ASTSerializer& serializer) const {
 // Each 'bit' is one input value in the row.
 class BitTrie {
 public:
-    // Tries to insert the row. Returns nullptr if the row is
-    // successfully inserted, and a pointer to the previously
-    // inserted row with the same bit pattern if a collision
+    // Tries to insert the row. Returns nullopt if the row is
+    // successfully inserted, and a set of the previously
+    // inserted rows with the same bit pattern if a collision
     // is found.
     template<typename TAllocator>
-    const UdpEntrySyntax* insert(const UdpEntrySyntax& syntax, std::span<const char> inputs,
-                                 char stateChar, TAllocator& allocator) {
+    std::optional<std::set<const UdpEntrySyntax*>> insert(const UdpEntrySyntax& syntax,
+                                                          std::span<const char> inputs,
+                                                          char stateChar, TAllocator& allocator) {
         SmallVector<BitTrie*> nodes;
         BitTrie* primary = this;
         nodes.push_back(this);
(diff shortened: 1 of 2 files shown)


source/ast/symbols/MemberSymbols.cpp
```diff
@@ -899,6 +913,21 @@ class BitTrie {
...
                 break;
+            // Below are implicit node identifiers that cannot be found in the UDP grammar. They are
+            // helpers for `p` and `n`. Handling `0` or `x` (for `p` and `r` matching cases)
+            case '6':
+                handle(6, true);
+                handle(0);
+                handle(2);
+                handle(3);
+                break;
+            // Handling `1` or `x` (for `p` and `r` matching cases)
```
MikePopoloski, line 924: Comment should be for 'n' right?

source/ast/symbols/MemberSymbols.cpp
```diff
@@ -766,13 +767,14 @@ void PrimitivePortSymbol::serializeTo(ASTSerializer& serializer) const {
...
-    // successfully inserted, and a pointer to the previously
-    // inserted row with the same bit pattern if a collision
+    // Tries to insert the row. Returns nullopt if the row is
+    // successfully inserted, and a set of the previously
+    // inserted rows with the same bit pattern if a collision
     // is found.
     template<typename TAllocator>
-    const UdpEntrySyntax* insert(const UdpEntrySyntax& syntax, std::span<const char> inputs,
-                                 char stateChar, TAllocator& allocator) {
+    std::optional<std::set<const UdpEntrySyntax*>> insert(const UdpEntrySyntax& syntax,
```
MikePopoloski, line 775: Rather than returning a std::set, which requires allocations (and you're not even specifically using the properties of a set as far as I can tell) can you make this function take a SmallVector that you add the matches to? Then it can return a boolean for whether the insertion was valid or not and the SmallVector can live on the stack in the caller.

source/ast/symbols/MemberSymbols.cpp
likeamahoney: yes, I meant 'n', sorry, fixed

source/ast/symbols/MemberSymbols.cpp
likeamahoney: thanks, fixed

PR #1100, opened 2024-08-26
Add preprocessor option to interpret translate_on/off pragmas

povik:
For synthesis tools based on slang, add a preprocessor option to interpret pragmas like

```
    // pragma translate_off
    assert(...); // meant to be ignored in synthesis
    // pragma translate_on
```

Given the non-standard nature of the pragmas, make the exact regex a preprocessor option.

Related: povik/yosys-slang#28

### DIFF
--- a/include/slang/parsing/Preprocessor.h
+++ b/include/slang/parsing/Preprocessor.h
@@ -8,6 +8,7 @@
 #pragma once
 
 #include <memory>
+#include <regex>
 
 #include "slang/parsing/Lexer.h"
 #include "slang/parsing/NumberParser.h"
@@ -57,6 +58,16 @@ struct SLANG_EXPORT PreprocessorOptions {
 
     /// A set of preprocessor directives to be ignored.
     flat_hash_set<std::string_view> ignoreDirectives;
+
+    /// A flag to enable the interpretation of non-standard line comment
+    /// pragmas within the preprocessor.
+    bool enableCompatPragmas = false;
+
+    /// A pair of regexes to match on the on/off pragmas in line comments.
+    /// The off pragma instructs the preprocessor to ignore all tokens until
+    /// the occurence of the on pragma.
+    std::regex compatOffPragma;
+    std::regex compatOnPragma;
 };
 
 /// Preprocessor - Interface between lexer and parser
@@ -153,6 +164,7 @@ private:
 
     // Internal methods to grab and handle the next token
     Token nextProcessed();
+    Token nextMasked();
     Token nextRaw();
     void popSource();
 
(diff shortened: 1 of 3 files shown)


include/slang/parsing/Preprocessor.h
```diff
@@ -57,6 +58,16 @@ struct SLANG_EXPORT PreprocessorOptions {
 
     /// A set of preprocessor directives to be ignored.
     flat_hash_set<std::string_view> ignoreDirectives;
+
+    /// A flag to enable the interpretation of non-standard line comment
+    /// pragmas within the preprocessor.
+    bool enableCompatPragmas = false;
```
likeamahoney, line 64: Is it assumed that these options can only be used from the API?

I think it would be useful to use them from the slang cli interface as well

povik: Right now I am setting these with a downstream patch to `Driver.cpp`. For yosys-slang I think we want it on by default, possibly with a cli option to change the regexes or disable it.

If we can agree on standard slang options for this (with yosys-slang having it on by default, vanilla slang having it off by default), that would be great

MikePopoloski:
The Lexer already has an optional path for checking for pragma comments -- see Lexer::scanLineComment and scanProtectComment. Can this functionality be put there as well instead of adding a new branch in the preprocessor?

Also how necessary is the regex behavior? Can a substring match be used instead? I ask because std::regex is notoriously slow -- if full regex functionality is required we probably want to at least provide the option of pulling in a performant 3rd party regex lib.

povik:
> The Lexer already has an optional path for checking for pragma comments -- see Lexer::scanLineComment and scanProtectComment. Can this functionality be put there as well instead of adding a new branch in the preprocessor?

It can though I think we want to collect the full comment line before we try to match it against the user-provided list of pragmas -- both for the off pragma and the on pragma. On the other hand the protect stuff is matched verbatim. So the new functionality can be moved to the Lexer but I think it would involve a new branch there too.

> Also how necessary is the regex behavior? Can a substring match be used instead? I ask because std::regex is notoriously slow -- if full regex functionality is required we probably want to at least provide the option of pulling in a performant 3rd party regex lib.

I don't think a simple substring match would cut it but if we write something simple which would bridge over any whitespacing differences in matching against the known pragma list then that would work. In any case rolling out custom matching sounds simpler than setting up the infrastructure for optional 3rd party regex libs.

MikePopoloski:
Yes, I think so. I'd like to keep all comment parsing isolated to that one branch if possible.

udif:
If we're going for legacy, you also have `// synopsys translate_off` and `// synopsys translate_on` that had been traditionally supported by DC.

povik:
The detection is configurable and accounts for those too. There's no set of pragmas which is hard-coded in the patch.

MikePopoloski:
Going to close this out for now. Feel free to reopen if you modify the approach, or open an issue if you want me to look into it when I get some spare cycles.

PR #1107, opened 2024-09-04
Implements getFullyOriginalRange

suzizecat:
Equivalent as getFullyOriginalLoc but directly for ranges

Simply generate a range for which `getFullyOriginalLoc` has been applied to both start() and end().
Return the original range if the resulting files mismatch.

Please let me know if there is a need to add tests for this.

### DIFF
--- a/source/text/SourceManager.cpp
+++ b/source/text/SourceManager.cpp
@@ -199,6 +199,15 @@ SourceLocation SourceManager::getFullyOriginalLoc(SourceLocation location) const
     return location;
 }
 
+SourceRange SourceManager::getFullyOriginalRange(SourceRange range) const {
+    SourceLocation start(getFullyOriginalLoc(range.start()));
+    SourceLocation end(getFullyOriginalLoc(range.end()));
+
+    if (getFileName(start) != getFileName(end))
+        return range;
+    return SourceRange(start, end);
+}
+
 SourceLocation SourceManager::getFullyExpandedLoc(SourceLocation location) const {
     std::shared_lock lock(mutex);
     return getFullyExpandedLocImpl(location, lock);
--- a/source/text/SourceManager.cpp
+++ b/source/text/SourceManager.cpp
@@ -202,9 +202,6 @@ SourceLocation SourceManager::getFullyOriginalLoc(SourceLocation location) const
 SourceRange SourceManager::getFullyOriginalRange(SourceRange range) const {
     SourceLocation start(getFullyOriginalLoc(range.start()));
     SourceLocation end(getFullyOriginalLoc(range.end()));
-
-    if (getFileName(start) != getFileName(end))
-        return range;
     return SourceRange(start, end);
 }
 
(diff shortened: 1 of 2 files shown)


source/text/SourceManager.cpp
```diff
@@ -199,6 +199,15 @@ SourceLocation SourceManager::getFullyOriginalLoc(SourceLocation location) const
     return location;
 }
 
+SourceRange SourceManager::getFullyOriginalRange(SourceRange range) const {
+    SourceLocation start(getFullyOriginalLoc(range.start()));
+    SourceLocation end(getFullyOriginalLoc(range.end()));
+
+    if (getFileName(start) != getFileName(end))
```
MikePopoloski, line 206: This is surprising behavior; why would you want to return the original range in this case? None of the other methods here do this.

suzizecat: The rationale is to mimic `getFullyOriginalLoc` which actually does that.

I could have raised an exception but I saw that is was not particularly encouraged.
Moreover, it may lead to some tedious error management where you are almost forced to use a `try ... catch` in order to use this function properly.

If you have a preferred way to handle this, please let me know and I'll try to implement it.

MikePopoloski: I don't think it's an error; if you have an include file then you can end up with a source range that starts in one file and ends in another and it's perfectly fine. Returning the original range doesn't even help in this case because they will still point to different files, so you may as well do what the caller asked and expand the macro locations.

MikePopoloski: Yes, you can have SourceRanges with different buffers in start and end in "normal" syntax trees, just by using include files, so you already need to check these conditions regardless of whether you call getFullyOriginalRange or not. `contains` does not work on these ranges -- that's part of the reason slang does not use that method internally anyway.

suzizecat: I still don't quite see which situation would lead to a CST returning an invalid range, but I won't argue any further. At this point, it is more curiosity than anything. 

Would it be interesting to add a `SourceRange::haveConsistentBuffer()` (I'm not gifted for choosing good function names) in order to easily performs this check ?  

I will update to remove this check and always expand the SourceRange. Thanks for your time !

MikePopoloski: Here's an easy, if contrived, example:
```
// inc.svh
module m;
```
```
// main.sv
`include "inc.svh"
endmodule
```
What is the SourceRange of the ModuleDeclarationSyntax in this case?

I'm not against adding a helper to SourceRange but honestly it's pretty easy to check `if (start.buffer == end.buffer)` so it's not a big deal.

suzizecat: In this situation, wouldn't you have a single buffer containing 
```systemverilog
module m;
endmodule
```

due to the macro expansion done at compilation level ? (With different files when using `getFullyOriginalLoc` though)

I'll see if adding this comparison (which can really quickly become `if(my_node.range().start().buffer() == my_node.range().end.buffer())` ) is (too) annoying and add the the helper if relevant in another discussion/issue.

suzizecat: Changes done.

PR #1113, opened 2024-09-06
addition of a abstract syntax tree to source code printer

jeffKeysight:
his pull request introduces the initial implementation of an AST to Source Code Printer as part of a recent internship project. The goal of this addition is to enable the conversion of AST representations back into human-readable source code. Currently, the functionality is limited to handling simple programs, more complex programs may not be accurately represented yet. The current version lacks support for 'property', 'checker', 'cover', and 'generate' statements, although it does provide basic support for other elements .

changes to the orginal AST are the addition of a 'name' member to the CovergroupType and corresponding updates to the JSON AST printer. A 'isDuplicate' flag has been introduced to VariableFlags to track duplicate information. Furthermore, a 'MinimalInfoEnabled' option has been added to the serializer, producing a JSON output with reduced information, facilitating the comparison of different ASTs for equivalence.

A test suite based on the 'all.sv' file has been made. Programs are validated by running the AST printer and comparing the resulting AST of the generated source code against the original code's AST. This comparison is made through the json representation.

(The workflow currently fails but i am currently developing a fix for this)


MikePopoloski:
Thanks for submitting a PR. Please get the tests and pre-commit checks to pass and revert any changes that aren't required to be part of this diff, such as the changes to the build workflow. Also from a higher level point of view I'm not against adding a printer for the AST but I'm curious what your use case is and whether the printer really needs to live in the slang library or could just exist in a downstream application.

MikePopoloski:
Closing this out until there's some updates.

Fromhell777:
Hey @MikePopoloski , I'll try to take over this pull request. I supervised the internship but I was busy with other stuff so I haven't gotten around to finalising this feature. Sorry that it's taking so long.

I think that Slang can greatly benefit from having a printer for the AST. This would be the perfect puzzle piece to unlock a whole new set of tooling capabilities. That's also the reason that we focused on this feature during the internship. The way I see it is that you broadly speaking have 3 kind of applications for Slang:

* Code Checking: This is meanly using Slang to read in SystemVerilog and doing some checks. Some common tool examples could be an LSP (Language Server Protocol), Linting (Langauage diagnostics), ...
  * Syntax Tree: This seems to be already greatly supported. You can read SystemVerilog and even generate diagnostics. Even add custom diagnostics
  * AST: For some diagnostics you need some more intelligent view of the SystemVerilog code. For this you need the AST. The tree you get after resolution or compilation I think you call it. This also seems to be already supported
* Code Generation: These applications just generate SystemVerilog code as an output. Here it's less important how the output looks as long as it's consistent and generates valid output. Although it would still be nice to have full control over the look of the output just to retain full flexibility. Some common tool examples could be a tool that generates structural hierarchy levels that just stitches RTL blocks together. This is common to do with extra tools. Another example could just be the creation of very repetitive SystemVerilog structures
  * Syntax Tree: This seems to be good supported. Although I may be overlooking some things here but at first glance you have enough flexibility do manipulate the output format
  * AST: This is lacking at the moment. Although you could say that you can create everything in just the Syntax Tree, it becomes much more cumbersome due to the more verbose nodes. Plus if you are planning to support the next "Code Adaptation" feature then this will be almost free
* Code Adaptation: These applications take in SystemVerilog code, do some manipulation and output some SystemVerilog code. Here it would be nice to retain the original style of the input as much as possible. For example, you would like to retain the comments. Here it would also be nice to have full control over the look of the output just to retain full flexibility. Some common tool examples could be an output formatter for SystemVerilog. It takes in SystemVerilog and formats it based on a set of rules for the given project. This is common for software project and to some extend RTL projects. Some other example could be to read in the SystemVerilog and change all the Asynchronously reset registers to Synchronously reset ones. This is a bit more involved but you can imaging trying to adapt part of the SystemVerilog code in order to streamline the project or maybe to circumvent some other tool limations 
  * Syntax Tree: This seems to be more or less ok. Although I may be overlooking some things here but at first glance you have enough flexibility do manipulate the output format. Plus you have the feature to remember a pointer to original code. This helps you in retaining the original code structure
  * AST: This is lacking at the moment. I think it's good trade-off to make a printer to just print everything in a standard way. If people want some other format  they can just replace those specific printing functions. We tried to make some default printing options for the AST nodes but noticed that some extra info was needed to print certain structures. You'll probably have a much better view on how to incorporate this extra intelligence for the AST printer to work optimally

Once I fixed the failing pre-commit checks we can discuss further on how to tackle the remaining problems. 

I see great potential in the software you already wrote and look forward to start integrating it into our own tools

MikePopoloski:
What you've said makes sense and I'm not against adding this if the PR can be polished up and into a good state. Looking forward to your contributions there.

trile127:
Any reason why this was closed?

Fromhell777:
@trile127 It was closed until the PR is cleaned up. I'm working on cleaning it up and trying to solve the problems with the github actions

ChrisLuck:
Very interesting feature!
I am planning to translate some expressions written in an obscure domain specific language to SV expressions and I am wondering if I could do that by translating from AST to AST and then use this feature to dump SV.

@Fromhell777 is there already updated code (compared to this PR) that I could look into and try to run?

inigo-easics:
FYI: Fromhell777 is also my account. I have multiple. Sorry if I mix between both sometimes

inigo-easics:
@MikePopoloski I have the clean version of the AST printer ready. Should this PR be opened again so I can add commits to it (If that works which I'm not sure about)? Or do you want me to create a new PR?

MikePopoloski:
Probably best to open a new clean PR.

inigo-easics:
I opened a new PR #1352

PR #1134, opened 2024-09-23
Netlist: add option to control unrolling of procedural for loops

jameshanlon:
This patch also removes the logic that implements some ad-hoc data dependencies in procedural blocks.
This means that the netlist tool takes a clearer default position that no data dependencies are captured within them, which is perfectly reasonable for most use cases.


jameshanlon:
Not to be landed please until I've figured out what is causing an infinite loop!

jameshanlon:
Okay, good to go.

PR #1157, opened 2024-10-17
Add lexer detection of translate_on/off pragmas

povik:
Follow-up to #1100 with a changed approach. The detection now happens in lexer, not preprocessor. Tests are missing.


source/parsing/Lexer.cpp
```diff
@@ -1198,6 +1198,96 @@ void Lexer::scanWhitespace() {
     addTrivia(TriviaKind::Whitespace);
 }
 
+bool detectTranslateOnOffPragma(std::string_view view, bool offMode) {
+    static std::vector<std::string_view> offCandidates = {
```
MikePopoloski, line 1202: These should be moved out of the function and marked const; statics inside of functions require the compiler to insert code to make sure statics get atomically initialized the first time the function is called, which is an unnecessary expense.

MikePopoloski: Also you should clearly document that the ordering here is really important, since your algorithm relies on them being in alphabetical order.

MikePopoloski (commented):
This approach seems ok. I think there must be a way to optimize it a bit more to avoid going into the full loop to check for cases since 99.9% of comments won't match, but it's probably ok for now.

You do need to add tests though; ideally you get full coverage of all branches, especially for something critical like the lexer.

source/parsing/Lexer.cpp
povik: I've moved the trigger word list, and documented the hidden requirements

povik:
> You do need to add tests though; ideally you get full coverage of all branches, especially for something critical like the lexer.

OK, I have started some testing, let me see what the report will say

povik:
Looks like the coverage is full

PR #1160, opened 2024-10-20
hermetic builds

### LINKED ISSUE #1155: hermetic builds
**Describe the bug**
Hermetic builds are a bit more strict with includes and paths.

**Additional context**
I have a potential [PR](https://github.com/MikePopoloski/slang/compare/master...cc10512:slang:cc-hermetic?expand=1) that consists of a couple of commits enabling such builds. Two main issues addressed in different commits:
1. adding missing includes for features used in header files.
2. add arguments to generative scripts that allow separating the location of the script from the location of the input. Or in other words, allow inputs to be specified explicitly on the command line. In addition a few more suggestions from static checking Python tools.

Please let me know it this is of interest.

cc10512:
Adds explicit arguments to the generator scripts to point to the input
such that they can be used in a hermetic build where the scripts do
not run from the source directory. Remaining inputs are assume to be
in the same directory.

Adds missing includes.

Fixes #1155

### DIFF
--- a/scripts/diagnostic_gen.py
+++ b/scripts/diagnostic_gen.py
@@ -8,15 +8,14 @@ import argparse
 import os
 import shlex
 import subprocess
+import sys
 
 
 def writefile(path, contents):
-    try:
-        with open(path, "r") as f:
+    existing = None
+    if os.path.exists(path):
+        with open(path, 'r') as f:
             existing = f.read()
-    except OSError:
-        existing = ""
-
     if existing != contents:
         with open(path, "w") as f:
             f.write(contents)
@@ -28,17 +27,23 @@ def main():
     parser.add_argument("--srcDir", help="Source directory to search for usages")
     parser.add_argument("--incDir", help="Include directory to search for usages")
     parser.add_argument("--docs", action="store_true")
+    parser.add_argument("--diagnostics", help="path to diagnostics file")
     parser.add_argument("--slangBin")
     args = parser.parse_args()
 
-    ourdir = os.path.dirname(os.path.realpath(__file__))
-    inf = open(os.path.join(ourdir, "diagnostics.txt"))
+    inf = None
+    try:
+        inf = open(args.diagnostics)
+    except Exception:
+        print("failed to open", args.diagnostics)
+        sys.exit(1)
 
     headerdir = os.path.join(args.outDir, "slang", "diagnostics")
     try:
-        os.makedirs(headerdir)
-    except OSError:
-        pass
+        os.makedirs(headerdir, exist_ok=True)
+    except OSError as err:
+        print("Directory '%s' can not be created" % headerdir)
+        sys.exit(1)
 
     diags = {}
     groups = []
@@ -88,7 +93,7 @@ def main():
     if args.docs:
         createdocs(
             args.outDir,
-            os.path.join(ourdir, "warning_docs.txt"),
+            os.path.join(os.path.dirname(args.diagnostics), "warning_docs.txt"),
             args.slangBin,
             diags,
             groups,
@@ -280,7 +285,7 @@ def createallheader(path, diags):
 
 def createdocs(outDir, inpath, slangBin, diags, groups):
     inf = open(inpath)
-    curropt = None
+    curropt = [ None, None, None, None, None ]
     inexample = False
     exampleMap = {}
 
@@ -310,7 +315,7 @@ def createdocs(outDir, inpath, slangBin, diags, groups):
                 curropt[1] += " "
             curropt[1] += line
 
-    if curropt:
+    if len(curropt[0]) > 0:
         exampleMap[curropt[0]] = curropt
 
     for k, v in exampleMap.items():
--- a/scripts/diagnostic_gen.py
+++ b/scripts/diagnostic_gen.py
@@ -42,7 +42,8 @@ def main():
     try:
         os.makedirs(headerdir, exist_ok=True)
     except OSError as err:
-        print("Directory '%s' can not be created" % headerdir)
+        print(err)
+        print("Directory '%s' can not be created: %s" % headerdir)
         sys.exit(1)
 
     diags = {}
--- a/scripts/diagnostic_gen.py
+++ b/scripts/diagnostic_gen.py
@@ -12,10 +12,12 @@ import sys
 
 
 def writefile(path, contents):
-    existing = None
-    if os.path.exists(path):
+    try:
         with open(path, "r") as f:
             existing = f.read()
+    except OSError:
+        existing = ""
+
     if existing != contents:
         with open(path, "w") as f:
             f.write(contents)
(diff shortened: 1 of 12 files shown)


tools/netlist/include/DirectedGraph.h
```diff
@@ -79,7 +79,7 @@ class Node {
         return *this;
     }
     Node<NodeType, EdgeType>& operator=(Node<NodeType, EdgeType>&& node) noexcept {
-        edges = std::move(node.Edges);
+        edges = std::move(node.edges);
```
MikePopoloski, line 82: What happened here?

scripts/diagnostic_gen.py
```diff
@@ -8,15 +8,14 @@
 import os
 import shlex
 import subprocess
+import sys
 
 
 def writefile(path, contents):
-    try:
```
MikePopoloski, line 14: Why this change to remove the try block?

scripts/diagnostic_gen.py
```diff
@@ -28,17 +27,24 @@ def main():
...
     parser.add_argument("--slangBin")
     args = parser.parse_args()
 
-    ourdir = os.path.dirname(os.path.realpath(__file__))
-    inf = open(os.path.join(ourdir, "diagnostics.txt"))
+    inf = None
+    try:
+        inf = open(args.diagnostics)
+    except Exception:
+        print("failed to open", args.diagnostics)
```
MikePopoloski, line 38: This doesn't seem a good use of a catch handler for the exception; you're going to exit anyway, just with a less informative error than you'd get if you let the exception bubble up naturally.

scripts/diagnostic_gen.py
```diff
@@ -28,17 +27,24 @@ def main():
...
+        sys.exit(1)
 
     headerdir = os.path.join(args.outDir, "slang", "diagnostics")
     try:
-        os.makedirs(headerdir)
-    except OSError:
-        pass
+        os.makedirs(headerdir, exist_ok=True)
+    except OSError as err:
```
MikePopoloski, line 44: Same here. The only reason the try block was here originally was in case the directory already exists, which it looks like `exist_ok` will handle now.

scripts/diagnostic_gen.py
```diff
@@ -280,7 +286,7 @@ def createallheader(path, diags):
 
 def createdocs(outDir, inpath, slangBin, diags, groups):
     inf = open(inpath)
-    curropt = None
+    curropt = [None, None, None, None, None]
```
MikePopoloski, line 289: Why this change?

scripts/diagnostic_gen.py
```diff
@@ -310,7 +316,7 @@ def createdocs(outDir, inpath, slangBin, diags, groups):
                 curropt[1] += " "
             curropt[1] += line
 
-    if curropt:
+    if len(curropt[0]) > 0:
```
MikePopoloski, line 319: And this one?

MikePopoloski (commented):
Approach seems fine, just some questions about a few details.

tools/netlist/include/DirectedGraph.h
cc10512: Syntax error:
```
third_party/sv_slang/tools/netlist/include/DirectedGraph.h:82:32: error: no member named 'Edges' in 'Node<NodeType, EdgeType>'; did you mean 'edges'?
   82 |         edges = std::move(node.Edges);
      |                                ^~~~~
      |                                edges
third_party/sv_slang/tools/netlist/include/DirectedGraph.h:156:18: note: 'edges' declared here
  156 |     EdgeListType edges;
      |                  ^
1 error generated.
Target //third_party/sv_slang:slang-netlist failed to build
```

And indeed, the only data member `Node` has is called `edges`.

scripts/diagnostic_gen.py
cc10512: As I was debugging the script for the hermetic build I looked at the code in this function.

I am of the old school that believes we should use error checking when available instead of trying to handle errors with an exception. I know it is a matter of taste, but in this case, where the exception doesn't even propagate, I find the code cleaner. Happy to defer to you though, if you prefer the try/catch, I can revert this change. I will still continue to argue that the code is clearer with the check :)

scripts/diagnostic_gen.py
cc10512: Agreed. However, the existing code had a `pass` in the catch block, and that means we'll get an exception later (when we try to write files) if the directory creation failed.

If you're happy having the exception bubble up, I can remove `sys.exit`. But I do think we should handle or throw it here, so that at least the user is pointed to the right location of the error.

scripts/diagnostic_gen.py
cc10512: Sorry, this was where the `pass` was, not above. So the comment above applies here. Exceptions can still happen, so I would keep the try/catch block, and let it buuble up, not pass, nor exit.

scripts/diagnostic_gen.py
cc10512: This was our more strict python type-checker. It really did not like the fact that we are assigning array elements to a None type. It did not figure out that the array was initialized in one of the branches before it was assigned to. 

I agree that your code made sense, and it is correct for the way the input files are structured, but the static checker does not have access to the inputs, and so it was throwing errors. This, arguably more verbose solution, quiets it down.

scripts/diagnostic_gen.py
cc10512: Consequence of the above. You could check for None before, but now it is assigned an array of None, so the original check would not work.

I'm not too fond of this change. I would like a solution upstream to avoid me maintaining a patch locally in our tree, so curious to hear whether you have a better way. But please also feel free to reject it if you "hate" it, and I'll deal with it.

tools/netlist/include/DirectedGraph.h
MikePopoloski: Hmm, I guess it's templated and nothing actually calls it.

scripts/diagnostic_gen.py
MikePopoloski: That's generally a good approach for a language like C++ but in Python it's considered idiomatic to use the exception path. It's also in this case not a correct change; doing `if (exists(file)) { open(file); }` has a race condition (in any language) where the file can be deleted in between the two calls. Doing it exceptionless requires an `open` method that returns an error code rather than throws.

scripts/diagnostic_gen.py
MikePopoloski: Yeah, just let the exception bubble up here if it happens.

scripts/diagnostic_gen.py
MikePopoloski: This is fine, I don't care too much one way or the other. I was just curious.

scripts/diagnostic_gen.py
MikePopoloski: Same here.

scripts/diagnostic_gen.py
cc10512: Reverted it.

scripts/diagnostic_gen.py
cc10512: Ack.

PR #1164, opened 2024-10-22
Support pre-loaded buffers in SourceLoader

povik:
Hi,

please see if this is acceptable. I have a use-case where I want to use Driver to get standard slang command line parsing, but I want to inject a piece of input which doesn't originate from a file.

### DIFF
--- a/include/slang/driver/SourceLoader.h
+++ b/include/slang/driver/SourceLoader.h
@@ -15,6 +15,7 @@
 #include <span>
 #include <vector>
 
+#include "slang/text/SourceLocation.h"
 #include "slang/syntax/SyntaxFwd.h"
 #include "slang/text/Glob.h"
 #include "slang/util/Hash.h"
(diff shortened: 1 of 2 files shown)


include/slang/driver/SourceLoader.h
```diff
@@ -159,6 +163,10 @@ class SLANG_EXPORT SourceLoader {
         // The filesystem path (as specified by the user).
         std::filesystem::path path;
 
+        // An optional pre-loaded buffer for when the source doesn't originate
+        // from the filesystem
+        const SourceBuffer preloadedBuffer;
```
MikePopoloski, line 168: This shouldn't be marked const; typically const fields in data structures can make them difficult to work with, like no longer being movable for example, which can make containers of them less efficient.

MikePopoloski (commented):
This approach seems fine to me, just one comment to fix.

povik:
I have addressed the comment.

PR #1184, opened 2024-12-03
Fix `BufferID.placeholder` static property, Doxygen docs

parker-research:
During the M.CSS documentation generation setup, it errored when it got to the following:

```
        .def_property_readonly_static("placeholder", &BufferID::getPlaceholder)
```

The issue is that the static function is actually treated as a `@classmethod` in Python, and the associated C++ function is called with the `cls = type(self)` argument. In the fix, an anonymous function is used to discard the `cls` in the call.

Also added a note about Doxygen.in which I wish was present from the start.

### DIFF
--- a/bindings/python/UtilBindings.cpp
+++ b/bindings/python/UtilBindings.cpp
@@ -58,9 +58,8 @@ void registerUtil(py::module_& m) {
         .def(py::init<>())
         .def_property_readonly("id", &BufferID::getId)
         .def_static("getPlaceholder", &BufferID::getPlaceholder)
-        .def_property_readonly_static("placeholder", [](py::object /* self or cls */) {
-            return BufferID::getPlaceholder();
-        })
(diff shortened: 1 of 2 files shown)


parker-research:
This bug is also mentioned in https://github.com/MikePopoloski/pyslang/issues/96 - it caused a roadblock when someone struggled to generate type stubs.

Esser50K:
So pparently been fixed? how can I generate things now?

I tried using the nightly build but the issue mentioned on https://github.com/MikePopoloski/pyslang/issues/96 is still present

parker-research:
Just confirmed that as of pyslang commit a864e10 (which bumped the submodule version), both of the following work:

```bash
pip install mypy
stubgen -p pyslang

pip install pybind11-stubgen
pybind11-stubgen pyslang
# Note: Lots of error log messages. Generation still succeeds though.
```

PR #1205, opened 2025-01-08
adding support for configurable clock name matching via regex

spomatasmd:
Adding support for clock name matching via regular expression.
The regex pattern is configurable and is used along with the static, single-name clock signal check already available in slang-tidy.

Defaulting to `"clk\S*|clock\S*"` i.e. `clk` or `clock` followed by zero or any non-whitespace characters.

When setting the option, double quotes must be used (will be parsed and skipped) since `[` and `]` are actually regex metacharacters and cannot be skipped as it's the case for lists of options.

### DIFF
--- a/tools/tidy/README.md
+++ b/tools/tidy/README.md
@@ -55,15 +55,16 @@ Disable all: -*
 The `CheckConfigs` is a dictionary like, `config: value`, comma separated list of options for the different checks.
 The available options are:
 
-|            Config             |   Type   | Default |
-|:-----------------------------:|:--------:|:-------:|
-|          **clkName**          |  string  |  clk_i  |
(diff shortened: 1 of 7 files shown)


MikePopoloski:
Thanks for the PR. Can you update the README to include the new functionality, and also add a test for UnusedSensitiveSignal that shows it actually working?

spomatasmd:
> Thanks for the PR. Can you update the README to include the new functionality, and also add a test for UnusedSensitiveSignal that shows it actually working?

Good point, added two extra tests (one failing, one succeeding) and described option in the README.

MikePopoloski:
No need, I can squash merge via the UI

PR #1217, opened 2025-01-20
[bug][slang] Fix case selector huge bitwidth segmentation fault

likeamahoney:
Hi @MikePopoloski!

There are a two examples of `casex` which works incorrect with slang:

```verilog
// This one is segfaults on release and debug builds

module test9(
  input reg [100:0] sel_i,
  input reg trg_i,
  output reg [1:0] reg_o
);
  always @(posedge trg_i) begin
    casex (sel_i)
      -12'b00zzzzz00001,
      12'b11: reg_o = 2'b01;
    endcase
  end
endmodule
```

and

```verilog
// This one produce the strange error on debug build:
// internal compiler error: Assertion 'noteLocation' failed
// in file /home/yan/projects/slang/source/diagnostics/Diagnostics.cpp, line 30
//  function: slang::Diagnostic& slang::Diagnostic::addNote(slang::DiagCode, slang::SourceLocation)

module test9(
  input reg [100:0] sel_i,
  input reg trg_i,
  output reg [1:0] reg_o
);
  always @(posedge trg_i) begin
    casex (sel_i)
      12'sb00xxxxx00001,
      -12'b00zzzzz00001,
      12'b11: reg_o = 2'b01;
    endcase
  end
endmodule
```

I think the whole point is that it is dangerous to use `unions` (for example - [here](https://github.com/MikePopoloski/slang/blob/master/source/ast/statements/ConditionalStatements.cpp#L287)), especially since it is impossible to check which [type](https://github.com/MikePopoloski/slang/blob/master/source/ast/statements/ConditionalStatements.cpp#L277) was actually written to the tree node. This can also happen because of different bit lengths of labels (in one case - for the first label `value.getBitWidth()` returns 12 (leading zeros in front), and in the other 101).

### DIFF
--- a/source/ast/statements/ConditionalStatements.cpp
+++ b/source/ast/statements/ConditionalStatements.cpp
@@ -239,21 +239,22 @@ public:
         if (auto result = find(value, 0, expr, wildcardX))
             return result;
 
-        CaseTrie* curr = this;
+        Node curr;
+        curr.trie = this;
         const auto width = value.getBitWidth();
         for (uint32_t i = 0; i < width; i++) {
             const auto bit = value[(int32_t)i];
             const bool isWildcard = wildcardX ? bit.isUnknown() : bit.value == logic_t::Z_VALUE;
             const auto valueIndex = isWildcard ? 3 : bit.isUnknown() ? 2 : bit.value;
 
-            auto& elem = curr->nodes[valueIndex];
+            auto& elem = curr.trie->nodes[valueIndex];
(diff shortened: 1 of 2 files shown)


MikePopoloski:
Thanks for the report, but this is not the right fix. The constant value *should* all be sized the same. There is an invariant being broken where the constant value of the expression is not sized to the width of the expression's type, due to some of the other warning analysis stuff I've added recently. #1222 is related.

PR #1254, opened 2025-02-24
Convert C++ docs config to conf.py

parker-research:
The [M.CSS docs suggest](https://mcss.mosra.cz/documentation/doxygen/#configuration) using this new `conf.py` configuration as it allows better/easier configurability than the old Doxygen way.

It also allows for easier syncing of the navbar spec between the pyslang and slang documentation configurations.


parker-research:
Waiting on this issue to be addressed first: https://github.com/mosra/m.css/issues/264

MikePopoloski:
Seems like this might be on hold for a while (and I'm not super convinced this is all that valuable anyway) so I'm going to close it out until something changes.

PR #1262, opened 2025-02-27
[slang] Do visit cached body

likeamahoney:
After adding instance bodies caching support (which is enabled by default) some old `instance` body usage patterns are stay untoched.

`ASTVisitor` is common interface for most of `slang` project tools. For example it widely used by `slang-tidy` checkers.

I think `ASTVisitor` interface should visit instance cached bodies if it presents. 

Otherwise, if `ASTVisitor` starts to visit the `body` of `instance` with previously cached body, it will elaborate it (for example in terms of `tidy` checker implementation this is odd implicit behaviour to do that, since elaboration should only be the `slang` compiler task).

### DIFF
--- a/include/slang/ast/ASTVisitor.h
+++ b/include/slang/ast/ASTVisitor.h
@@ -64,7 +64,8 @@ concept HasVisitExprs = requires(const T& t, TVisitor&& visitor) { t.visitExprs(
 /// visited -- you can include that behavior by invoking @a visitDefault
 /// in your handler.
 ///
-template<typename TDerived, bool VisitStatements, bool VisitExpressions, bool VisitBad = false>
+template<typename TDerived, bool VisitStatements, bool VisitExpressions, bool VisitBad = false,
+         bool VisitCanonical = false>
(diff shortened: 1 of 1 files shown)


likeamahoney:
An another solution is to move InstanceSymbol body field into private and provide a correct body resolving into getBody method for example. It prevent situations with InstanceSymbol incorrect API usage by anyone.

But I don't know which fix is better.

MikePopoloski:
It is not necessarily wrong to descend into the unique instance body; up until very recently that's what everything was doing anyway. I'm not convinced we always want to do this for all visitors. It shouldn't be wrong to visit every instance, just perhaps slower. I could see an option being added though.

likeamahoney:
> It is not necessarily wrong to descend into the unique instance body; up until very recently that's what everything was doing anyway. I'm not convinced we always want to do this for all visitors. It shouldn't be wrong to visit every instance, just perhaps slower. I could see an option being added though.

I added an template parameter `VisitCanonical` which allow to visit canonical bodies

PR #1275, opened 2025-03-12
BumpAllocator constructor compile issue fix

nickrallison:
C++ slang

**Describe the bug**
When building on a M3 Macbook Pro running MacOS Sequoia 15.3.1 with the macos-release cmake configuration (and any other configuration I tried), the build errored out with this error:

```
[ 67%] Building CXX object source/CMakeFiles/slang_slang.dir/ast/SystemSubroutine.cpp.o
[ 68%] Building CXX object source/CMakeFiles/slang_slang.dir/ast/TimingControl.cpp.o
In file included from /Users/nick/Projects/Rust/slang/source/ast/Compilation.cpp:8:
In file included from /Users/nick/Projects/Rust/slang/source/../include/slang/ast/Compilation.h:12:
In file included from /Users/nick/Projects/Rust/slang/source/../include/slang/ast/ASTDiagMap.h:13:
In file included from /Users/nick/Projects/Rust/slang/source/../include/slang/diagnostics/Diagnostics.h:17:
In file included from /Users/nick/Projects/Rust/slang/source/../include/slang/numeric/ConstantValue.h:16:
In file included from /Users/nick/Projects/Rust/slang/source/../include/slang/numeric/SVInt.h:16:
In file included from /Users/nick/Projects/Rust/slang/source/../include/slang/util/SmallVector.h:18:
/Users/nick/Projects/Rust/slang/source/../include/slang/util/BumpAllocator.h:41:54: error: no matching constructor for initialization of 'slang::ast::Compilation::NetAlias'
        return new (allocate(sizeof(T), alignof(T))) T(std::forward<Args>(args)...);
```

**To Reproduce**
Described above.

**Additional context**
I was able to fix the compile issue with a small edit to bump allocator as seem in this commit [here](https://github.com/nickrallison/slang/commit/f401323991e4c2887e269850c15ff2291707173d).

### DIFF
--- a/include/slang/util/BumpAllocator.h
+++ b/include/slang/util/BumpAllocator.h
@@ -38,7 +38,7 @@ public:
     template<typename T, typename... Args>
     T* emplace(Args&&... args) {
         static_assert(std::is_trivially_destructible_v<T>);
-        return new (allocate(sizeof(T), alignof(T))) T(std::forward<Args>(args)...);
+        return new (allocate(sizeof(T), alignof(T))) T{std::forward<Args>(args)...};
     }
 
     /// Allocate @a size bytes of memory with the given @a alignment.
@@ -134,7 +134,7 @@ public:
     /// Construct a new item using the allocator.
     template<typename... Args>
     T* emplace(Args&&... args) {
-        return new (allocate(sizeof(T), alignof(T))) T(std::forward<Args>(args)...);
+        return new (allocate(sizeof(T), alignof(T))) T{std::forward<Args>(args)...};
     }
 };
 


MikePopoloski:
This doesn't make much sense to me; I'm building locally on a Macbook with the same OS version and everything is fine, and the CI build also covers a macos-release build. Is there something special about your build environment?

nickrallison:
I'm honestly not very sure, I use nix to manage my dependencies but I'm using Apple Clang to build. I assume its something about the specific compiler version like some minor version change.

Here is the exact command I use to try to build:
```sh
cd ~/Downloads;
git clone https://github.com/MikePopoloski/slang.git;
cd slang;
cmake --preset macos-release;
cd build/macos-release;
make -j8;
```

And here is the log file showing the error, sorry I can't be more help, I have no idea what might be causing it.
[log.txt](https://github.com/user-attachments/files/19229831/log.txt)

MikePopoloski:
Looks like you're building with AppleClang version 15, which is not supported. The minimum is AppleClang 16 (i.e. Xcode 16). I'm actually surprised you're able to build with just this minor tweak; last I checked there were other broken things in the older Xcode versions.

This fix is innocuous but I'm pretty sure you're going to run into other troubles at some point, so I'd recommend just updating your Xcode version instead. In general slang requires whatever the latest release of Xcode is, since Apple is so far behind C++ support compared to other compilers.

PR #1278, opened 2025-03-17
Update file name in header comment in Pyslang bindings

parker-research:
Updated to match the style that appears to be present in recently-edited and recently-created files throughout the repo


MikePopoloski:
This is only needed for generating Doxygen API reference for the classes contained in the files. We don't want the Python binding classes to show up in the C++ API documentation so this change should not be landed.

parker-research:
Fair enough, thanks

PR #1302, opened 2025-04-05
Add semantic analysis for `celldefine directive

### LINKED ISSUE #1295: Identifying `celldefine annotation on a module
C++ slang

For synthesis, modules with the `celldefine annotation need special handling as they describe a cell, not user logic. Ideally there would be a way to detect this by looking at the slang AST.

I don't have a concrete idea. Something similar to the attribute handling, perhaps?

Looking up the syntax node for a module, and then looking up the directive in the trivia preceding the syntax node

**Additional context**
If you've a good idea of how this should be implemented, let me know and I'll take a stab at it.

whitequark:
Fixes #1295.

This is my first PR to slang, and I expect that I'll miss some of the finer points of the design. In particular I'm not sure if the transfer of the metadata from the AST to the compilation database is quite right.

### DIFF
--- a/source/parsing/Preprocessor.cpp
+++ b/source/parsing/Preprocessor.cpp
@@ -191,6 +191,7 @@ void Preprocessor::resetAllDirectives() {
     activeTimeScale = std::nullopt;
     defaultNetType = TokenKind::WireKeyword;
     unconnectedDrive = TokenKind::Unknown;
+    cellDefine = false;
     resetProtectState();
 }
 
@@ -356,6 +357,12 @@ Token Preprocessor::handleDirectives(Token token) {
                     case SyntaxKind::NoUnconnectedDriveDirective:
                         trivia.push_back(handleNoUnconnectedDriveDirective(token));
                         break;
+                    case SyntaxKind::CellDefineDirective:
+                        trivia.push_back(handleCellDefineDirective(token));
+                        break;
+                    case SyntaxKind::EndCellDefineDirective:
+                        trivia.push_back(handleEndCellDefineDirective(token));
+                        break;
                     case SyntaxKind::DefaultDecayTimeDirective:
                         trivia.push_back(handleDefaultDecayTimeDirective(token));
                         break;
@@ -369,8 +376,6 @@ Token Preprocessor::handleDirectives(Token token) {
                             trivia.push_back(skipped);
                         break;
                     }
-                    case SyntaxKind::CellDefineDirective:
-                    case SyntaxKind::EndCellDefineDirective:
                     case SyntaxKind::DelayModeDistributedDirective:
                     case SyntaxKind::DelayModePathDirective:
                     case SyntaxKind::DelayModeUnitDirective:
@@ -1088,6 +1093,16 @@ Trivia Preprocessor::handleUnconnectedDriveDirective(Token directive) {
 Trivia Preprocessor::handleNoUnconnectedDriveDirective(Token directive) {
     checkOutsideDesignElement(directive);
     unconnectedDrive = TokenKind::Unknown;
+    return createSimpleDirective(directive);
+}
+
+Trivia Preprocessor::handleCellDefineDirective(Token directive) {
+    cellDefine = true;
+    return createSimpleDirective(directive);
+}
+
+Trivia Preprocessor::handleEndCellDefineDirective(Token directive) {
+    cellDefine = false;
     return createSimpleDirective(directive);
 }
 
--- a/tests/unittests/parsing/PreprocessorTests.cpp
+++ b/tests/unittests/parsing/PreprocessorTests.cpp
@@ -1287,15 +1287,25 @@ TokenKind lexDefaultNetType(std::string_view text) {
 
 TEST_CASE("default_nettype directive") {
     CHECK(lexDefaultNetType("`default_nettype wire") == TokenKind::WireKeyword);
+    CHECK_DIAGNOSTICS_EMPTY;
     CHECK(lexDefaultNetType("`default_nettype uwire") == TokenKind::UWireKeyword);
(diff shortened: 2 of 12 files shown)


tests/unittests/parsing/PreprocessorTests.cpp
```diff
@@ -1326,6 +1326,30 @@ TEST_CASE("unconnected_drive directive") {
     CHECK(!diagnostics.empty());
 }
 
+TokenKind lexCellDefine(std::string_view text) {
```
MikePopoloski, line 1329: Looks like this function return type should be `bool`

tests/unittests/parsing/PreprocessorTests.cpp
```diff
@@ -1326,6 +1326,30 @@ TEST_CASE("unconnected_drive directive") {
...
+    Token token = preprocessor.next();
+    REQUIRE(token);
+    return preprocessor.getCellDefine();
+}
+
+TEST_CASE("celldefine directive") {
+    CHECK(lexCellDefine("`celldefine") == true);
+    CHECK(lexCellDefine("`celldefine\n`endcelldefine") == false);
+    CHECK(lexCellDefine("`endcelldefine") == false);
+    CHECK_DIAGNOSTICS_EMPTY; // TODO: is this right? lexCellDefine() does diagnostics.clear()
```
MikePopoloski, line 1344: Yeah, you'd need this after each lexCellDefine call.

whitequark: There are many other calls to `lex*` like this elsewhere in the file with the same issue; should they be fixed too?

MikePopoloski: Hmm yeah, looks like the unconnected drive and timescale tests have the same problem.

whitequark: Fixed those.

tests/unittests/parsing/PreprocessorTests.cpp
whitequark: Fixed

whitequark:
(Converted to draft as I realized that my original testing has failed to exercise the new codepaths and yosys-slang isn't picking up the celldefine directive for some reason.)

tests/unittests/parsing/PreprocessorTests.cpp
```diff
@@ -1318,14 +1328,44 @@ TokenKind lexUnconnectedDrive(std::string_view text) {
...
+    CHECK(lexCellDefine("`celldefine") == true);
+    CHECK_DIAGNOSTICS_EMPTY;
+
+    CHECK(lexCellDefine("`celldefine\n`endcelldefine") == false);
+    CHECK_DIAGNOSTICS_EMPTY;
+
+    CHECK(lexCellDefine("`endcelldefine") == false);
+    CHECK_DIAGNOSTICS_EMPTY;
+
+    CHECK(lexCellDefine("`celldefine asdf") == false);
```
MikePopoloski, line 1362: These are not an error because the celldefine handler isn't trying to parse any arguments after it. This just looks to the parser like a celldefine followed by an identifier called `asdf`

tests/unittests/parsing/VisitorTests.cpp
```diff
@@ -694,6 +694,7 @@ class C; endclass
     CHECK(SyntaxPrinter::printFile(*newTree) == R"(
 `default_nettype none
 `unconnected_drive pull0
+`celldefine
```
MikePopoloski, line 697: You added this to the *expected* output string. You should also add it up above to the actual input, and then this test should pass.

source/parsing/Preprocessor.cpp
```diff
@@ -1091,6 +1096,18 @@ Trivia Preprocessor::handleNoUnconnectedDriveDirective(Token directive) {
     return createSimpleDirective(directive);
 }
 
+Trivia Preprocessor::handleCellDefineDirective(Token directive) {
+    checkOutsideDesignElement(directive);
```
MikePopoloski, line 1100: The LRM says "These directives may appear anywhere in the source description, but it is recommended that the directives be specified outside any design elements." so we shouldn't actually enforce that they appear outside a design element.

MikePopoloski (commented):
I can't speak to the downstream usage but the approach here is correct, you just need some tweaks to fix the tests. You should figure out why your local test runs aren't failing first though. You should be able to just run the unittests binary after you build.

tests/unittests/parsing/PreprocessorTests.cpp
whitequark: Thanks. I was just looking up the detailed semantics of the Verilog preprocessor directives to try and figure out what's the expected behavior here.

source/parsing/Preprocessor.cpp
whitequark: Thanks, I missed that.

whitequark:
> You should figure out why your local test runs aren't failing first though. You should be able to just run the unittests binary after you build.

I was going to rely on the CI to run the tests, but failed to account for GitHub's behavior where it requires approval for CI workflows for every push. I did set that up now though.

MikePopoloski:
> > You should figure out why your local test runs aren't failing first though. You should be able to just run the unittests binary after you build.
> 
> I was going to rely on the CI to run the tests, but failed to account for GitHub's behavior where it requires approval for CI workflows for every push. I did set that up now though.

It should start running them automatically after you've had one accepted PR to the repo I think. Regardless though, the unit tests run in just a couple of seconds so it's not very onerous to run locally.

whitequark:
I've addressed all of the review comments, ensured the tests pass locally, and tested the new directive with yosys-slang. This PR should be ready to merge now.

PR #1307, opened 2025-04-07
pyslang: Create bindings for SyntaxRewriter

### LINKED ISSUE #1271: Pyslang: Add bindings for `SyntaxRewriter`
Pyslang Python bindings

It seems that the Python bindings for the `slang::syntax::SyntaxRewriter` don't exist yet. Porting these to Python would be a huge benefit.

Port the `slang::syntax::SyntaxRewriter` bindings.

I could learn C++ better ;)

**Additional context**
Seems that the visitor is already ported (as in `test_visitors.py`), but that the visitor system is exclusively read-only currently.

parker-research:
Massive project - implementing bindings and tests for SyntaxRewriter. Finally done. 

There seems to be one quirk - the check at the end of the `test_rewriter_nested()` test (see the `# FIXME` comment). Would really appreciate it if you could help me take a look at that and see why `assert result.root.isEquivalentTo(expected.root) is True` fails, but checking string equality (neglecting whitespaces) passes. 

Fixes #1271

### DIFF
--- a/bindings/python/SyntaxBindings.cpp
+++ b/bindings/python/SyntaxBindings.cpp
@@ -55,6 +55,63 @@ void pySyntaxVisit(const SyntaxNode& sn, py::object f) {
     sn.visit(visitor);
 }
 
+class PySyntaxRewriter : public SyntaxRewriter<PySyntaxRewriter> {
+    public:
+        PySyntaxRewriter(pybind11::function handler) : handler(std::move(handler)) {}
+    
+        void visit(const SyntaxNode& node) {
+            try {
+                handler(pybind11::cast(&node), pybind11::cast(this));
+            } catch (const pybind11::error_already_set& e) {
+                throw;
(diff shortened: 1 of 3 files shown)


MikePopoloski:
Thanks for the PR, the approach seems fine to me.

As to your question, if you walk the produced syntax tree you can see where they diverge: the new struct member you're inserting is a DataDeclaration but struct types expect their members to be StructUnionMember syntax nodes. They happen to stringify to the same thing here but it would not work properly if you wanted to build an AST from this.

A minimal fix is to change the new node you create to something like this:
```
new_decl = pyslang.SyntaxTree.fromText(
                            "typedef struct{logic logic_member_to_insert;}t;", "new.sv"
                        ).root.type.members[0]
```

parker-research:
Updated - should be good now! Thanks for your help on that one.

Issue #1269, opened 2025-03-03
Invalid `None`/`with` occurrences in Python bindings


parker-research:
**Is your bug report related to the C++ slang project or the pyslang Python bindings?**
Pyslang Python bindings

**Describe the bug**
The following enums in the Python bindings have the value `None = 0`, which is a syntax error:

```
239-class EdgeKind(enum.Enum):
240:    None = 0
--
311-class ForwardTypeRestriction(enum.Enum):
312:    None = 0
--
416-class RandMode(enum.Enum):
417:    None = 0
--
1541-class UnconnectedDrive(enum.Enum):
1542:    None = 0
--
1546-class UniquePriorityCheck(enum.Enum):
1547:    None = 0
```

Each of them must be changed to `None_ = 0`.

I think we've handled some of these cases before. I think this is the remaining list to fix.

Similarly, here are the examples of invalid `with` naming:

```python

class WithClauseSyntax(SyntaxNode):
    closeParen: Token
    expr: ExpressionSyntax
    openParen: Token
    with: Token
class WithFunctionClauseSyntax(SyntaxNode):
    name: NameSyntax
    with: Token
class WithFunctionSampleSyntax(SyntaxNode):
    function: Token
    portList: FunctionPortListSyntax
    sample: Token
    with: Token

class ArrayOrRandomizeMethodExpressionSyntax(ExpressionSyntax):
    args: ParenExpressionListSyntax
    constraints: ConstraintBlockSyntax
    method: ExpressionSyntax
    with: Token

class BinSelectWithFilterExprSyntax(BinsSelectExpressionSyntax):
    closeParen: Token
    expr: BinsSelectExpressionSyntax
    filter: ExpressionSyntax
    matchesClause: MatchesClauseSyntax
    openParen: Token
    with: Token
```

I may take a look later.

PR #1310, opened 2025-04-07
fix: Invalid autogenerated bindings using `None`/`with` fields

parker-research:
Partial fix for #1269


parker-research:
This is now a complete fix for the issue. Fixes #1269.

PR #1313, opened 2025-04-08
Account for UTF-8 chars to derive column numbers

racodond:


MikePopoloski:
Thanks for the PR. This change doesn't work as-is -- several parts of the diagnostic printing logic assume that the result of getColumn can be used to walk back to the start of line. Probably there needs to be a separate function that can be called to get a "display" column for purposes of printing the actual number. Also probably this needs to be controlled by a flag, since otherwise this will also affect the output for AST serialization and JSON diagnostic serialization, and consumers may be expecting that they get a byte offset and not a display character offset. GCC's `-fdiagnostics-column-unit` option seems relevant.

racodond:
> Thanks for the PR. This change doesn't work as-is -- several parts of the diagnostic printing logic assume that the result of getColumn can be used to walk back to the start of line. Probably there needs to be a separate function that can be called to get a "display" column for purposes of printing the actual number. Also probably this needs to be controlled by a flag, since otherwise this will also affect the output for AST serialization and JSON diagnostic serialization, and consumers may be expecting that they get a byte offset and not a display character offset. GCC's `-fdiagnostics-column-unit` option seems relevant.

Thanks @MikePopoloski for the feedback.
We'll have a look at it and keep you posted.
Note that it can take a few weeks since we're currently working on a higher priority task. But for sure, we'll eventually fix that one.

MikePopoloski:
Superseded by 2db7cd50da0889e7cea0b57e67ce4d8c0c69d2e6

PR #1325, opened 2025-04-16
Add root scope accessor on Compilation

povik:
This is the only outstanding downstream patch in yosys-slang that's critical. Please take a look if adding `getRootScope()` is fine. I use it with `compilation.tryGetDefinition()` to find if there's an existing definition without triggering finalization ([link](https://github.com/povik/yosys-slang/blob/25540d82e35d1478b453670b752001cb1f5b1ef6/src/blackboxes.cc#L59))

### DIFF
--- a/include/slang/ast/Compilation.h
+++ b/include/slang/ast/Compilation.h
@@ -759,6 +759,9 @@ public:
     /// Creates an empty ImplicitTypeSyntax object.
     const syntax::ImplicitTypeSyntax& createEmptyTypeSyntax(SourceLocation loc);
 
+    /// Get the root scope (c.f. getRoot() this method doesn't trigger finalization)
+    const Scope *getRootScope() const;
+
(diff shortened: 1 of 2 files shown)


MikePopoloski:
This looks the same as the existing `tryGetRoot` method?

povik:
> This looks the same as the existing `tryGetRoot` method?

Looks like I can just as well use that after 330f4d0cfff78a4fb90890432db309fff0f3603a but the part about it being null if not finalized isn't accurate

MikePopoloski:
Oh yeah, that's misleading. Probably it should be renamed and return a reference instead. I'll look at it.

povik:
Thanks, meanwhile I'm closing this

MikePopoloski:
f3c9f3a9d229a80c79d6646cb8b68cc73cfd46bf

PR #1326, opened 2025-04-16
Make "missing time scale" error demotable

povik:


povik:
I missed that's an option. Let me do that instead and I'll re-open if need be.

PR #1352, opened 2025-05-12
Initial step to having a printer of the AST nodes to text

inigo-easics:
(This is a continuation of a previous pull request)

Alright, I reviewed all the code and cleaned up where needed. All the CI jobs
are clean now.

I still have the following remarks. There is definitely still a bunch of work
left but it's interesting to lay out a couple of missing info/features of the
AST while working on the AST printer. As it is right now, there are some changes
needed to AST to fully reconstruct a valid SystemVerilog program:

* Let me start with the biggest topic. While gaining some more insight in
  parsers the last months, I realised that the current AST printer approach is
  not so optimal. Namely we try to transform all the AST nodes to a string. This
  causes a lot of tricky code of where do I need to put extra spaces and how do
  I correctly format the strings. I suggest to rework the AST printer to just
  output a token steam. These tokens could be something like the Lexer tokens.
  Afterwords you can let loose a formatter on this token stream to transform it
  to an appropriate string. This decouples the task of mapping the AST to text
  tokens and formatting it to the users liking
* We would want to create some options for style output of the AST printer.
  Ideally if we have a token stream (as discussed in the previous point) then we
  could make a formatter with format rule. Or maybe we can use an external
  formatter library
* What do we do with pre-processed code. I think it would be nice if we have an
  option to output AST code that is not pre-processed. Or have the option to
  some compiler directives in the output. But this might be bit tricky to do.
  For example, if you read in source code, transform it to AST and then print it
  back, you would like the compiler directive to not be expanded
* We added an isDuplicate flag to the VariableFlags of the VariableSymbols.h
  file. This was needed for the AST printer to know that some symbols were
  redundant to be printed. Maybe we can tackle this another way
* The CovergroupType class in CoverSymbols.h and CoverSymbols.cpp was updated to
  have an extra name member and to have this isDuplicate flag (from the previous
  point). This was needed for correct AST printing and JSON serialisation
* The AST does not have the right info to print the GenericClassDefSymbol node.
  At the moment we just print the underlying pointer to the syntax nodes
* The AST does not have the right info to print the GenerateBlockSymbol and the
  GenerateBlockArraySymbol node. At the moment we just print the underlying
  pointer to the syntax nodes
* The AST does not have the right info to print the CheckerSymbol node. It seems
  to be missing the connection to the body (CheckerInstanceBodySymbol)
* The TypeAliasType has the type_str member that has the string representation
  of the typedef. This prohibits the AST printer to do his own string formatting
* Interfaces are substituted when invoked. So I don't now how to recover to
  original interface again
* Parameters get resolved in the AST. This prohibits you from outputting the
  correct parameters again
* I noticed a difference in the PatternCaseStatement AST node compared to a
  couple of months ago. Previously if the case statement just contained a single
  expression, the PatternCaseStatement had a pointer to an ExpressionStatment
  object. But right now it always points to a BlockStatement. Which confused the
  AST printer a bit since this resulted in extra "begin end" delimiters. I don't
  know if this was the intention but it's interesting to notify
* In the AST printer we encountered some difficulties with printing the Symbol
  names. Maybe we're missing something here but this part is still difficult to
  get right. Perhaps an AST update could solve this
* As the testing of the AST printer we added a file with unittests. It has the
  following test flow:
  * We take the source text and transform it to the AST
  * We transform the AST to JSON with the current JSON serialiser
  * We transform the AST to SystemVerilog code with the new AST printer
  * From the new SystemVerilog code we read it back in and transform it to AST
  * This new AST gets transformed to JSON and compared to the first JSON output
  * This worked fine as a first set of tests and quickly checked that the AST
     printer is working correctly. I assume we later want to update the testing
     setup to not depend anymore on the JSON serialisation
* We encountered some problems with printing types. You have some options for
  type printing but this made the unittest rather complex to compare with JSON.
  We added an "includeTypes" option to the ASTSerializer.h so we can remove this
  info from the JSON output. This may be reverted in the future or handle
  differently
* At this point we loose the comments when printing the AST. It would be a great
  benefit to have a way of retaining comments or having some knowledge in the
  AST so we can also have comments in the AST printer output
* The AST printer .cpp files were split into a couple of files to reduce the
  amount of contents into 1 file and the recompile time. Maybe you have a
  different opinion/better solution on this matter

As a side note I don't have full knowledge of the slang project, let along much
knowledge about parsers and AST trees. So there are a lot of tasks left and
probably some things that must be reworked since I overlooked/misunderstood
something. I hope this first step may help in bringing up a mature AST
printer/formatter


inigo-easics:
> I agree with your take on outputting tokens (or maybe more likely a new concrete syntax tree using the SyntaxNode derived classes?) and then running a formatter over it so you don't have to worry about spacing everywhere in the printer. That would be a generally useful thing outside this printer, since you could run it on any parsed syntax tree and get standardized formatted output back out.

Yes a new concrete syntax tree might be a good solution. Then we can reuse a the formatter on any concrete syntax tree that we create. So either from source code or generated from the AST

> A lot of your other points touch on the fact that you're operating on the abstract syntax tree. What makes it abstract is that it's already been normalized and so loses information about what exactly the user originally typed. Printing it should give you a valid SystemVerilog design, but not necessarily looking the same as the original code. The AST represents the elaborated design and so reversing that is going to be hard. Examples here evaluated parameters, unrolled generate loops, etc.

I agree that not all information is retained in the AST. But I think we need at least enough information to have sensible creation of SystemVerilog code. For example, with a port list we still have the info to output an Ansi style list or not. You could have maybe removed this information and make a generic port list that does not need to know anything whether it originated from Ansi or not. The same I would say for the Generate statements. At this points it's not possible for the user to output any generate statement from the AST. This seems like a big drawback for what SystemVerilog code he wants to generate.

So yeah, I think these pieces of metadata must be looked at case by case and decided whether you want to abstract them away or keep some info you at least know which code/style to generate when you transform it to SystemVerilog code

> It's not clear to me what problems you were having with symbol names or serialized types.

Right now we keep some extra info in the AST printer to remeber the last symbol. This work was done by our intern at the time so I don't know the full details/trade-off he made. Probably you can pass along this info to the lower "handle" function another way. Or maybe it would be handier if the AST node had a pointer to the accompanied symbol

> I think I asked this on the previous PR, but can you express again what end goal you're trying to achieve here? In my mind the point of printing the AST would be to get the normalized form, but from your comments above it seems like you're maybe hoping for something else instead.

Getting the normalized form is definitely one of the goals. Ideally I would be able to do the following things:
* Read in source code. Change a couple of things. Write it back out as sensible SystemVerilog. For changing these things you often need the elaboration step to be done. So then we need to get back to SystemVerilog from the AST. Preferably as humanly readable as possible and also with the comments retained otherwise the output code can't really replace the original
* Easy code generation. At first glance I would say this is much easier with the AST. Since the concrete syntax tree is rather bloated for quick code generation

Some common tool examples I have in mind are:
* An intelligent formatter. For example, for some formatting rules you would need the elaboration step. Like wanting all the register names to have the "r_" prefix. For this you to know what becomes a signal. I guess you could split it up into a set of formatting rules that can be purely done on the concrete syntax tree. This would limit the rules a bit. And for the more complicated rules, these are checked with a Linter that just report them but the user needs to fix it themself
* If I would like to read in SystemVerilog code and change the reset style from Asynchornous to Synchronous
* If I would like to read in 2 SystemVerilog modules. Create a hierarchy level that instantiates both modules and automatically connect the ports with the same name. Kind of an automatic structural hierarchy generator
* If I would like to read in SystemVerilog code and transform it to Verilog code due to EDA tool limitations. Preferably we can retail having all the output also be in original files but having 1 big output file would maybe be ok for this goal
* If I have a memory map tool or any other code generation tool then it seems much easier to generate the AST compared to a concrete syntax tree. This may be solved with some extra wrapper

So I guess for most of the application there are a couple of important decisions to make:
* Do we want to output code into separate files so they can replace the original source code. Or can we just output the AST code into 1 big file and see it as an intermediate representation which is never going to be read. For some of the application I would argue that you want them to replace the original source code. So I would like this to be supported
* What to do with pre-processer statements. I think you want the option to right out macro's and include statements when generating code. So again I think this would need to be supported if the output code is ever going to be checked in

MikePopoloski:
I think your examples are all good things to be able to do, but I think most of them really want to operate on the CST so that what you output looks as close to possible as the original source code the user wrote, just with whatever modifications you want performed. This is essentially what the SyntaxRewriter class is supposed to help do, though probably there is a bunch of work to add helpful APIs on top to make working with creating and formatting new nodes on the fly easier to deal with.

I think your comment about ansi vs non-ansi port style is a good example of the kind of thing that the AST loses. The information about the original port style can only be obtained from the CST syntax nodes, which are sort of an optional part of the AST nodes (we set them right now because the AST is always constructed from one or more CSTs, but nothing stops someone from fabricating AST nodes internally or from some per-serialized source or whatever, which then would not have CST syntax nodes attached).

Note that just because you're working with the CSTs (to do the actual edits and serialization) doesn't mean you can't construct the AST and use that to help make decisions about what to edit. Like for your example of adding a "r_" prefix automatically to register names, you can construct the AST, build a map from CST -> AST nodes, and then walk the CST and rewrite the nodes that correspond to a "register" using whatever criteria you want. If you don't do the serialization from the CST though, your output will probably always be different enough from the original source that nobody will want to use it as a refactoring tool.

inigo-easics:
I get your point that most of the time you want to stay with the CST since it's the easiest to reconstruct the original code. Although I thought this was more of a second step:
* The first step might have been to transform the whole CST to AST. Do your changes. Then transform the AST back to CST. Then print it with a nice formatter
* A second step could be to transform the code to CST. Have the AST on the side to extract the necessary info. Create some extra CST or AST nodes. Transform the AST nodes back to CST nodes and insert them into the correct place of the original CST. This way you change less of the original CST. But juggling both CST, AST and conversion between the two seemed rather daunting at first and a lot of extra API's. But I may be wrong on this part and it may be simpler depending on the application

I guess if you want to work with the CST as much as possible I would still suggest the following:
* An API to transform the AST nodes back to CST ones. Preferably you still want to be able to generate all the useful CST nodes from the AST. Like generate statement I would find rather weird to not being able to make them from AST nodes
* A formatter for the CST nodes. This could pretty print all the CST nodes with some useful options. Like expand macro's or not, use the original source text if a pointer exists, take some kind of style rules, ...

MikePopoloski:
Yeah, there's already some examples in the VisitorTests.cpp file where a "SemanticModel" class handles the mapping from AST node back to CST node which is used to resolve some type info and rewrite the CST. It's clunky because not much effort has been put into making it nice, but it works. A more general AST -> CST transformer is going to run into exactly the same set of issues as this printer does, which is that there isn't a way to get the "original" look of the source code, be that ansi vs non-ansi ports, exact way that certain types are specified (since they are uniquified), macros pre-expansion, etc. It should be possible recover generates, since they result in GenerateBlockSymbols in the AST. You won't be able to see untaken generate branches though for conditional generates.

I agree that a general formatter for the CST would be really nice. There's already a printer in the library called `SyntaxPrinter` with options to do things like expand macros or not, but it doesn't do any kind of reformatting / styling of the code. Code formatting is a known Hard Problem but there's plenty of work on it for other languages (clang-format, etc) that there might be some nice reusable pieces these days.

MikePopoloski:
Yeah, unfortunately the way SystemVerilog works, you're basically required to do full elaboration and constant evaluation in order to be able to do name lookup and type checking correctly in all cases, so the current AST design is essentially the only reasonable next step after the CST. I agree that it doesn't match a typical AST you'd find for a software language -- perhaps the closest analogue is templates in C++, where you can't really do type checking and name resolution without expanding templates, and so the AST really needs to operate on the post-expansion form.

Anyway, I agree that the two other things you mention would be unambiguously good to have. In particular SystemVerilog codebases could really benefit from a top tier code formatter and I think the slang CST is a good starting point for that.

MikePopoloski:
Going to close this since it seems the discussion has run its course.

PR #1355, opened 2025-05-14
Add flags for depfile generation

AndrewNolte:
This is analagous to [Verilator's flag](https://verilator.org/guide/latest/exe_verilator.html?highlight=mmd#cmdoption-MMD) and [clang's flag](https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-MMD) for getting the used include files.

We opted to make the naming more similar to [Icarus's flags](https://steveicarus.github.io/iverilog/usage/command_line_flags.html#general#-M:~:text=after%20that%20date.-,%2DM%3Cmode%3E%3D%3Cpath%3E,-Write%20into%20the)

For support in ninja/make systems, there's the `--depfile-target` flag

### DIFF
--- a/bindings/python/SyntaxBindings.cpp
+++ b/bindings/python/SyntaxBindings.cpp
@@ -255,6 +255,13 @@ void registerSyntax(py::module_& m) {
              })
         .def("__str__", &SyntaxNode::toString);
 
+    py::class_<IncludeMetadata>(m, "IncludeMetadata")
+        .def(py::init<>())
+        .def_readonly("syntax", &IncludeMetadata::syntax)
+        .def_readonly("path", &IncludeMetadata::path)
+        .def_readonly("buffer", &IncludeMetadata::isSystem)
+        .def_readonly("isSystem", &IncludeMetadata::isSystem);
+
     py::class_<SyntaxTree, std::shared_ptr<SyntaxTree>>(m, "SyntaxTree")
         .def_readonly("isLibraryUnit", &SyntaxTree::isLibraryUnit)
         .def_static(
@@ -321,7 +328,7 @@ void registerSyntax(py::module_& m) {
         .def_property_readonly("root", py::overload_cast<>(&SyntaxTree::root))
         .def_property_readonly("options", &SyntaxTree::options)
         .def_property_readonly("sourceLibrary", &SyntaxTree::getSourceLibrary)
-        .def_property_readonly("getIncludeDirectives", &SyntaxTree::getIncludeDirectives)
+        .def("getIncludeDirectives", &SyntaxTree::getIncludeDirectives)
         .def_static("getDefaultSourceManager", &SyntaxTree::getDefaultSourceManager, byref);
 
     py::class_<LexerOptions>(m, "LexerOptions")
--- a/pyslang/tests/test_basics.py
+++ b/pyslang/tests/test_basics.py
@@ -85,6 +85,15 @@ source:3:23: warning: arithmetic between operands of different types ('logic' an
     )
 
 
+def test_include_metadata():
+    tree = pyslang.SyntaxTree.fromText(
+        """
+    `include "some_file.svh"
+    """
+    )
+    assert len(tree.getIncludeDirectives()) == 1
+
+
 def test_script_session():
     session = pyslang.ScriptSession()
     session.eval("""integer arr[string] = '{"Hello":4, "World":8, default:-1};""")


tools/driver/slang_main.cpp
```diff
@@ -154,45 +161,52 @@ int driverMain(int argc, TArgs argv) {
...
-                                            obfuscateIds == true);
+                return driver.runPreprocessor(includeComments == true, includeDirectives == true,
+                                              obfuscateIds == true);
             }
-            else if (onlyMacros == true) {
+            if (onlyMacros == true) {
                 driver.reportMacros();
+                return true;
+            }
+            if (outputIncludeDepfile) {
```
AndrewNolte, line 174: MMD is typically a side-output of a normal compile, which is why I rearranged this so it could be used with both `--parse-only` and with normal compiles

source/parsing/Preprocessor.cpp
```diff
@@ -212,6 +212,15 @@ std::vector<const DefineDirectiveSyntax*> Preprocessor::getDefinedMacros() const
     return results;
 }
 
+std::vector<IncludeMetadata> Preprocessor::getIncludeDirectives() const {
+    // return includeDirectives;
+    std::vector<IncludeMetadata> results;
+    for (auto& include : includeDirectives) {
+        results.push_back(include);
+    }
+    return results;
```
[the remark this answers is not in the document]
MikePopoloski: Actually since this is only ever used to return a std::vector it should probably just be stored as a std::vector.

tools/driver/slang_main.cpp
```diff
@@ -117,6 +116,30 @@ int driverMain(int argc, TArgs argv) {
         driver.cmdLine.add("--ast-json-detailed-types", serializeDetailedTypes,
                            "When dumping AST to JSON, expand out all type information");
 
+        // TODO: trim extra files based on parser metadata (no compilation)
```
MikePopoloski, line 119: Remove these commented out options until you're ready to add them.

include/slang/parsing/ParserMetadata.h
```diff
@@ -10,7 +10,9 @@
 #include <optional>
 #include <vector>
 
+#include "slang/parsing/Preprocessor.h"
 #include "slang/parsing/Token.h"
+#include "slang/syntax/AllSyntax.h"
```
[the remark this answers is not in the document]
AndrewNolte: I think clangd was trying to be smart when I added code, I should be able to remove this.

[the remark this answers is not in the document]
AndrewNolte: Oh actually I remember this change- this wasn't meant to go in this PR. The ModuleDeclarationSyntax* is more accurate, but I guess the large header was the reason this wasn't more narrow.

include/slang/driver/Driver.h
```diff
@@ -331,6 +331,15 @@ class SLANG_EXPORT Driver {
     /// Prints all macros from all loaded buffers to stdout.
     void reportMacros();
 
+    std::vector<std::string> getPreprocessorIncludes();
```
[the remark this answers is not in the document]
AndrewNolte: getPreprocessorIncludes() can't be const because it runs the preprocessor directly (going off of reportMacros's impl). Since this is likely to be used in a full compilation, should I change the impl to grab the includes from those syntax trees instead of preprocessing twice?

source/driver/Driver.cpp
```diff
@@ -738,6 +742,95 @@ void Driver::reportMacros() {
...
+void Driver::reportFilelist(std::string& outputName, std::vector<std::string>&& sortedFiles,
+                            bool useDepfileFormat) {
+    std::ostringstream contentsStream;
+
+    if (useDepfileFormat) {
+        contentsStream << outputName << ": ";
+        for (auto& file : sortedFiles)
+            contentsStream << file << " ";
+        contentsStream << "\n";
+        OS::writeFile(outputName + ".d", contentsStream.str());
```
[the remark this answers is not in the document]
AndrewNolte: I was going off the convention because there's really two names that need to be specified- the make target and the name of the dep file. I initially didn't want two flags for this, but now that there's --dep-file needed, the target name can be stored in that

source/driver/Driver.cpp
```diff
@@ -701,25 +696,29 @@ bool Driver::runPreprocessor(bool includeComments, bool includeDirectives, bool
     return true;
 }
 
-void Driver::reportMacros() {
-    Bag optionBag;
-    addParseOptions(optionBag);
-
+class PreprocessorRunner {
```
AndrewNolte, line 699: Thoughts on name/location? I'm assuming it should go somewhere else, and maybe call it <something>Driver instead of runner

include/slang/driver/Driver.h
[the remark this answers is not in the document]
MikePopoloski: Hmm yeah, probably best to just always take from the syntax tree's stored includes. Thus is you ever run with a -M option you will always need to at least perform parsing, but in the common case you won't preprocess everything twice.

source/driver/Driver.cpp
MikePopoloski: I suppose if you follow the other comment chain and always pull the include files from the parsed syntax trees you won't need to make this preprocessor change at all. Otherwise the name / location seems fine to me, it's just a small internal helper class.

source/driver/Driver.cpp
```diff
@@ -636,12 +636,9 @@ static std::string generateRandomAlphanumericString(TGenerator& gen, size_t len)
 
 bool Driver::runPreprocessor(bool includeComments, bool includeDirectives, bool obfuscateIds,
                              bool useFixedObfuscationSeed) {
-    Bag optionBag;
-    addParseOptions(optionBag);
-
     BumpAllocator alloc;
     Diagnostics diagnostics;
-    Preprocessor preprocessor(sourceManager, alloc, diagnostics, optionBag);
+    Preprocessor preprocessor(sourceManager, alloc, diagnostics, createOptionBag());
```
MikePopoloski, line 641: Looks like you meant for this to be `createParseOptionBag` instead?

tools/driver/slang_main.cpp
```diff
@@ -117,6 +115,25 @@ int driverMain(int argc, TArgs argv) {
         driver.cmdLine.add("--ast-json-detailed-types", serializeDetailedTypes,
                            "When dumping AST to JSON, expand out all type information");
 
+        std::optional<std::string> depfileTarget;
+        driver.cmdLine.add(
+            "--depfile-target", depfileTarget,
+            "Use depfile (.d) format for generated dependency files, meaning --M<mode> <name> will "
+            "generate <depfileTarget>.d with <name> as the makefile target.");
```
MikePopoloski, line 122: The .d is no longer automatically added so the string should be updated here.

source/driver/Driver.cpp
```diff
@@ -738,11 +735,61 @@ void Driver::reportMacros() {
...
+
+    std::sort(paths.begin(), paths.end());
+
+    FormatBuffer buffer;
+    if (depfileTarget) {
+        buffer.format("{}:", *depfileTarget);
+        for (const auto& file : paths) {
+            buffer.format(" {}", file);
+        }
+        buffer.format("\n");
```
MikePopoloski, line 780: Use `append` when there is no formatting to be done.

include/slang/driver/Driver.h
```diff
@@ -331,6 +331,11 @@ class SLANG_EXPORT Driver {
     /// Prints all macros from all loaded buffers to stdout.
     void reportMacros();
 
+    std::vector<std::filesystem::path> getDepFiles(bool includesOnly = false) const;
+
+    std::string writeDepfiles(const std::vector<std::filesystem::path>& files,
```
MikePopoloski, line 336: I don't love the name because it's not actually writing to anything anymore. Perhaps something like "serializeDepFile" instead? Not great either to be honest but maybe better. Also should be consistent that "DepFile" is two "words", like in getDepFiles above.

MikePopoloski (commented):
Ok, this is looking pretty good. Please also add Doxygen comments and Python bindings for new API surfaces.

PR #1360, opened 2025-05-20
driver: generate compliant SV identifiers when obfuscating

Sustrak:
SystemVerilog identifiers shall not start with a digit, hence this commit removes the generation of numbers in the random string that replaces the identifiers when obfuscating


tests/unittests/DriverTests.cpp
```diff
@@ -3,6 +3,7 @@
 
 #include "Test.h"
 #include <fmt/core.h>
+#include <iostream>
```
MikePopoloski, line 6: Why was this extra include added?

Sustrak: Ups... Added a `cout` to get the output of a tests and didn't remember to remove it. Fixed :)

PR #1376, opened 2025-05-25
slang: Add AST Matchers

Sustrak:
Add the capability to match the AST Nodes ala clang matchers.

Combining different Matchers and using the MatchFinder, the user will be able to match specific nodes in the AST and execute a callback function to evaluate the matched AST node

### DIFF
--- a/include/slang/matchers/MatchFinder.h
+++ b/include/slang/matchers/MatchFinder.h
@@ -1,0 +1,76 @@
+//------------------------------------------------------------------------------
+//! @file MatchFinder.h
+//! @brief Class that tries to find the nodes that matches the rules provided
+//
+// SPDX-FileCopyrightText: Michael Popoloski
+// SPDX-License-Identifier: MIT
+//------------------------------------------------------------------------------
+#pragma once
+
+#include "MatchersInternal.h"
+
+namespace slang::ast::matchers {
+
+class MatchResult {
+public:
+    explicit MatchResult(const BoundNodesMap& boundNodes, ASTContext& context) :
+        boundNodes(boundNodes), context(context) {}
+
+    template<typename T>
+    std::optional<const T*> getNodeAs(const std::string& id) const {
+        if (const auto it = boundNodes.find(id); it != boundNodes.end()) {
+            return it->second->as<const T>();
+        }
+        return {};
+    }
+
+private:
+    const BoundNodesMap& boundNodes;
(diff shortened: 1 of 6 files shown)


include/slang/matchers/MatchFinder.h
```diff
@@ -0,0 +1,76 @@
...
+    ASTContext& context;
+};
+
+class MatcherCallback {
+public:
+    virtual ~MatcherCallback() = default;
+    virtual void run(const MatchResult& result) = 0;
+};
+
+class MatchFinder : public ASTVisitor<MatchFinder, true, true> {
```
Sustrak, line 38: Hi @MikePopoloski I would like to ask you for feedback on the `MatchFinder` class. Although its function is to traverse the AST and match its nodes I think that inherit from `ASTVisitor` and "override" the `visit()` function, might not be the best solution. What do you think about it? Thanks

MikePopoloski: Ok, I didn't do a line by line review but the general idea here seems ok to me and does seem modeled after what clang does. Do you have some good driving use cases for this to help prove out the design?

Sustrak: Hi @MikePopoloski I've been taking a closer look at the architecture of clang's ASTMatchers and I think there is a first step that needs to be done which is the creation  of a generic container to store the different kinds of Nodes in the AST. So far my solution only works with `Symbol` but I'm guessing we would like to match `Statements` and `Expressions` too. Clang has solved this with its [DynTypedNode](https://github.com/llvm/llvm-project/blob/78eafb14f7c91a40d493b29a3ce977f327c13892/clang/include/clang/AST/ASTTypeTraits.h#L257). And I'm guessing we could take the same approach.

> Do you have some good driving use cases for this to help prove out the design?

I think `slang-tidy` is the perfect candidate for this feature, matchers would make writing linting rules so much less verbose than what it is now with the visitors

MikePopoloski: Ok, that makes sense. A generic container for AST nodes is pretty easy, just use a std::variant<Expression, Statement, Symbol>. You can wrap it up if you want something a bit fancier but not hard.

MikePopoloski:
Going to close this out since it's incomplete / just a draft. Feel free to reopen if you get to a more complete PR.

PR #1377, opened 2025-05-26
Pyslang: Implement `.to_dict()` and `.to_json()` methods

parker-research:
* Changed the pybind11 module name to `_pyslang`, as recommended in https://github.com/pybind/pybind11/issues/1004
* Implemented rudimetary node_to_str function, which seems to work effectively on many/most node types
* Monkey-patched that in as a method in Python

I'm curious to see how the docs build goes with this (esp. with the `_pyslang` existence now); fingers crossed.

### DIFF
--- a/pyslang/pyslang/__init__.py
+++ b/pyslang/pyslang/__init__.py
@@ -1,0 +1,76 @@
+# Import Pybind11 bindings
+from _pyslang import *
+
+import json
+
+# Generic conversion from node to dict
+def node_to_dict(node):
+    if isinstance(node, (int, float, bool, str)):
+        return node
+    
+    if node is None:
+        return None
+    
+    cls_name = type(node).__name__
+    
+    if isinstance(node, SyntaxKind | TokenKind | TriviaKind):
+        return str(node).split('.')[-1]
+    
+    if isinstance(node, SourceLocation):
+        return {
+            "type": cls_name,
+            # "buffer": node.buffer, # TODO: Figure out how to serialize buffer
+            "offset": node.offset,
+        }
+    
+    if isinstance(node, list):
+        return [node_to_dict(item) for item in node]
+    
+    if isinstance(node, dict):
+        return {k: node_to_dict(v) for k, v in node.items()}
+    
+    text = str(node) if hasattr(node, '__str__') else None
+
+    def custom_sort_key(item):
+        # Put start before end, and sort others by name
+        priority = {'kind': -100, 'start': 0, 'end': 1}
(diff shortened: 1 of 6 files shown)


pyslang/pyslang/__init__.py
```diff
@@ -0,0 +1,88 @@
+import json
+
+# Import Pybind11 bindings
+from _pyslang import *
+
+__all__ = [
```
MikePopoloski, line 6: This limits the exported names to just this list right? So all of the other stuff in _pyslang is no longer pulled in by import *?

pyslang/pyslang/__init__.py
```diff
@@ -0,0 +1,88 @@
+import json
```
MikePopoloski, line 1: Please include the copyright header.

pyslang/pyslang/__init__.py
```diff
@@ -0,0 +1,88 @@
...
+
+    if node is None:
+        return None
+
+    cls_name = type(node).__name__
+
+    if isinstance(node, (SyntaxKind, TokenKind, TriviaKind)):
+        return str(node).split(".")[-1]
+
+    if isinstance(node, SourceLocation):
```
MikePopoloski, line 32: Including locations in the output is going to make it extremely voluminous.

pyslang/pyslang/__init__.py
```diff
@@ -0,0 +1,88 @@
...
+
+    return result
+
+
+# Extend relevant classes with `to_dict` and `to_json`
+def _extend_with_serialization(cls):
+    def to_dict(self):
+        return node_to_dict(self)
+
+    def to_json(self, **kwargs):
```
MikePopoloski, line 79: I don't think you need to clutter the API with a to_json method. to_dict is sufficient, the JSON serialization is just a one liner at that point.

pyslang/tests/test_syntax_tree_to_json.py
```diff
@@ -0,0 +1,45 @@
...
+endmodule
+"""
+
+
+def test_to_dict_conversion():
+    tree = pyslang.SyntaxTree.fromText(testfile)
+    assert tree.root.kind == pyslang.SyntaxKind.ModuleDeclaration
+
+    json_output = tree.root.to_dict()
+    assert isinstance(json_output, dict)
```
MikePopoloski, line 37: This should probably actually check the results of the dict.

MikePopoloski:
Any updates here?

MikePopoloski:
Closing since there's been no followup.

PR #1381, opened 2025-05-27
Give type alias symbols a syntax

AndrewNolte:
Looking up a type parameter results in a type alias symbol with no syntax, and there aren't other consistent means of getting that syntax from the symbol.


MikePopoloski:
Looks like your test fails to compile with MSVC btw.

PR #1382, opened 2025-05-28
Change all.sv test methodology

AndrewNolte:
Rather than counting syntaxes, change to count the unique nodes seen of each type.

### DIFF
--- a/tests/unittests/parsing/VisitorTests.cpp
+++ b/tests/unittests/parsing/VisitorTests.cpp
@@ -734,6 +734,16 @@ class C; endclass
     }
 }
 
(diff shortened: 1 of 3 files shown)


tests/unittests/parsing/VisitorTests.cpp
```diff
@@ -745,14 +783,30 @@ TEST_CASE("Visit all file") {
     Compilation compilation;
     compilation.addSyntaxTree(*tree);
 
-    int count = 0;
-    compilation.getRoot().visit(makeVisitor([&](auto& v, auto& elem) {
```
MikePopoloski, line 749: It would be nice to keep a `makeVisitor` call somewhere, since that specific method has been broken in the past.

PR #1395, opened 2025-06-04
[slang][analysis] Fix unique_ptr UAF bug

likeamahoney:
Hi, all!

There is a Use-after-free [bug](https://github.com/MikePopoloski/slang/blob/master/source/analysis/AnalyzedProcedure.cpp#L205) when moved `unique_ptr` [used](https://github.com/MikePopoloski/slang/blob/master/source/analysis/AnalyzedProcedure.cpp#L210) lately through raw [ptr](https://github.com/MikePopoloski/slang/blob/master/source/analysis/AnalyzedProcedure.cpp#L204) which had gotten using `get()`.

It's UB exploitation due to standard.

I have't built any minimal example which can help you to reproduce it.

But you can run `slang` directly on [file](https://github.com/olgirard/openmsp430/blob/92c883abb4518dbc35b027e6cad5ffef5b2fbb81/fpga/altera_de0_nano_soc/bench/verilog/altsyncram.v) from this open repo and you will get a segmentation fault on release and debug builds.

### DIFF
--- a/source/analysis/AnalyzedProcedure.cpp
+++ b/source/analysis/AnalyzedProcedure.cpp
@@ -201,8 +201,8 @@ void AnalyzedProcedure::addFunctionDrivers(AnalysisContext& context, const CallE
     auto analysis = context.manager->getAnalyzedSubroutine(subroutine);
     if (!analysis) {
         analysisMem = std::make_unique<AnalyzedProcedure>(context, subroutine);
-        analysis = analysisMem.get();
         context.manager->addAnalyzedSubroutine(subroutine, std::move(analysisMem));
+        analysis = context.manager->getAnalyzedSubroutine(subroutine);
     }
 
     // For each driver in the function, create a new driver that points to the


MikePopoloski:
This explanation is not correct; it's perfectly fine to use the underlying memory owned by the unique_ptr after moving -- moving doesn't destroy the object. The actual problem here is that addAnalyzedSubroutine might *not* move the memory if another thread raced and got its entry in the table first. I put a better fix in 011efea49f1dfe80b55334df3031076654858f39

likeamahoney:
> This explanation is not correct; it's perfectly fine to use the underlying memory owned by the unique_ptr after moving -- moving doesn't destroy the object. The actual problem here is that addAnalyzedSubroutine might _not_ move the memory if another thread raced and got its entry in the table first. I put a better fix in [011efea](https://github.com/MikePopoloski/slang/commit/011efea49f1dfe80b55334df3031076654858f39)

Okay, thank you for clearing it!

Issue #1403, opened 2025-06-08
Add warning for when no latch is inferred in an always_latch


MikePopoloski:
Winferred-latch works for always_comb -- the opposite case, when no latch is inferred inside an always_latch, should also be diagnosed.

MikePopoloski:
Added in 66fa94a4862e6dd49c32036b57eaa1d3a76b6c29

PR #1404, opened 2025-06-09
[slang][analysis] Add checking no latch for always_latch

likeamahoney:
Closes https://github.com/MikePopoloski/slang/issues/1403


MikePopoloski:
This check should be per-variable, not per-procedure, just like the existing latch check. That way the message can include exactly which variable (or really longest static prefix of a variable) is not inferring a latch.

likeamahoney:
> This check should be per-variable, not per-procedure, just like the existing latch check. That way the message can include exactly which variable (or really longest static prefix of a variable) is not inferring a latch.

sorry for the late response

the fix is done

MikePopoloski:
I had a slightly more efficient approach in mind that I ended up implementing in 66fa94a4862e6dd49c32036b57eaa1d3a76b6c29 which also incorporates your unit test changes, so thanks for that.

PR #1413, opened 2025-06-16
[AstVisitor] check for non-leaf classes

AndrewNolte:
Stack on https://github.com/MikePopoloski/slang/pull/1382

- Add `final` to all Ast leaf classes
- Add `final` check in visitDefault()- this can be a point of confusion when making visitors- when visiting a non-leaf class, visitDefault() will not dispatch properly, and needs to be handled differently

`VariableSymbol` is both a leaf and a parent, so we can solve this by either
- special casing, as is done on this pr atm
- splitting VariableSymbol into VariableSymbol and VariableSymbolBase


include/slang/ast/ASTVisitor.h
```diff
@@ -126,6 +138,17 @@ class ASTVisitor {
...
             t.body.visit(DERIVED);
         }
+        // check if not final and if there's a visit method for the derived type
+        static_assert((std::is_final_v<T> || std::is_same_v<VariableSymbol, T>),
+                      "Non-leaf class was passed to visitDefault() - use "
+                      "`std::derived_from<> auto& node` or visitDerived(node) instead");
+    }
+
+    template<typename T>
+    void visitDerived(const T& t) {
```
MikePopoloski: Let's not include this visitDerived thing for now since it will otherwise be unused / untested.

MikePopoloski:
Ok, looks good, please rebase.

PR #1415, opened 2025-06-17
Remove netlist tool

jameshanlon:
This tool will now live in its own downstream repository.


MikePopoloski:
Looks like you need to remove the reference to the README from `docs/tools.dox`

udif:
Could we have an alternative location published and documented (in the slang docs, for the benefit of existing users) **before** removing the current tool?

jameshanlon:
> Could we have an alternative location published and documented (in the slang docs, for the benefit of existing users) **before** removing the current tool?

Apologies @udif, I will setup a new repository ASAP and once that's done @MikePopoloski will add a link in the README/docs.

udif:
Thanks @jameshanlon and thanks again for all your work on the netlist parts (and a big thanks ofcourse to @MikePopoloski whose enormous work on all this project made this possible in the first place).

PR #1417, opened 2025-06-20
ConstantRange: add a constructor from std::pair

jameshanlon:
To allow simple conversions from `IntervalMap` ranges to `ConstantRange`.

### DIFF
--- a/include/slang/numeric/ConstantValue.h
+++ b/include/slang/numeric/ConstantValue.h
@@ -314,6 +314,7 @@ struct SLANG_EXPORT ConstantRange {
     int32_t right = 0;
 
     ConstantRange() = default;
+    ConstantRange(std::pair<uint32_t, uint32_t> range) : left(range.second), right(range.first) {}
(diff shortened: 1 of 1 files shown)


include/slang/numeric/ConstantValue.h
```diff
@@ -314,6 +314,7 @@ struct SLANG_EXPORT ConstantRange {
     int32_t right = 0;
 
     ConstantRange() = default;
+    ConstantRange(std::pair<int32_t, int32_t> range) : left(range.second), right(range.first) {}
```
MikePopoloski: Hmm, it's pretty confusing to swap the left and right like this. It seems like the following lines should be equivalent yeah?
```
ConstantRange(1, 2)
ConstantRange(std::pair(1, 2))
```

jameshanlon: I was thinking of left and right as being syntactical, corresponding to the way ranges are written in SV.
Although the class does support that, I can see it's more straightforward to match the other constructor.

PR #1422, opened 2025-06-25
Assert that general always blocks have some timing control

AndrewNolte:
From 9.2.2.1 General purpose always procedure:

"""
The always keyword represents a general purpose always procedure, which can be used to represent
repetitive behavior such as clock oscillators. The construct can also be used with proper timing controls to
represent combinational, latched, and sequential hardware behavior.

The general purpose always procedure, because of its looping nature, is only useful when used in
conjunction with some form of timing control. If an always procedure has no control for simulation time to
advance, it will create a simulation deadlock condition.

The following code, for example, creates a zero-delay infinite loop:
`always areg = ~areg;`
Providing a timing control to the preceding code creates a potentially useful description as shown in the
following:
`always #half_period areg = ~areg;`
"""

I've held off on updating all the failing tests because I want to get feedback on implementation first regarding where it should go:
  - Parser, where it is in this pr. The diag here seems a bit a premature, but the other option I saw was to place it in block symbols. This impl only operates on syntax, and so would either do redundant checks or need to cache results.
  - AnalysisManager, since technically that's correct, although this implementation is at the syntax level. It could also catch more cases with constant folding, but the major foot-guns are fairly simple as in the first example above.

### DIFF
--- a/scripts/diagnostics.txt
+++ b/scripts/diagnostics.txt
@@ -1150,6 +1150,7 @@ error MultipleUDNTDrivers "net '{}' with user-defined nettype '{}' cannot have m
 error InputPortAssign "cannot assign to input port '{}'"
 error ClockVarTargetAssign "cannot write to '{}' with a continuous assignment since it's associated with an output clocking signal"
 error LocalFormalVarMultiAssign "cannot bind local variable '{}' to more than one output or inout local variable formal argument"
+error AlwaysWithoutTimingControl "'always' procedure does not advance time and so will create a simulation deadlock."
 note NoteRequiredHere "required here"
 note NoteClockHere "clock here"
 warning unused-def UnusedDefinition "{} definition is unused"
--- a/source/analysis/AnalysisManager.cpp
+++ b/source/analysis/AnalysisManager.cpp
@@ -236,6 +236,42 @@ void AnalysisManager::getFunctionDrivers(const CallExpression& expr, const Symbo
         getFunctionDrivers(*call, containingSymbol, visited, drivers);
 }
 
+void AnalysisManager::getTaskTimingControls(const CallExpression& expr,
+                                            SmallSet<const SubroutineSymbol*, 2>& visited,
+                                            std::vector<const ast::Statement*>& controls) {
+    if (expr.getSubroutineKind() != SubroutineKind::Task || expr.isSystemCall()) {
+        return;
+    }
+
+    auto& subroutine = *std::get<const SubroutineSymbol*>(expr.subroutine);
+    if (subroutine.flags.has(MethodFlags::Pure | MethodFlags::InterfaceExtern |
+                             MethodFlags::DPIImport | MethodFlags::Randomize |
+                             MethodFlags::BuiltIn)) {
+        return;
+    }
+
+    // If we've already visited this task then we don't need to
+    // analyze it again.
+    if (!visited.insert(&subroutine).second)
+        return;
+
+    // Get analysis for the task.
+    auto analysis = getAnalyzedSubroutine(subroutine);
+    if (!analysis) {
+        auto proc = std::make_unique<AnalyzedProcedure>(getState().context, subroutine);
+        analysis = addAnalyzedSubroutine(subroutine, std::move(proc));
+    }
+
+    // Add timing controls from the task to our list.
+    auto taskTimingControls = analysis->getTimingControls();
+    controls.insert(controls.end(), taskTimingControls.begin(), taskTimingControls.end());
+
+    // If this task has any calls, we need to recursively add timing controls
+    // from those calls as well.
+    for (auto call : analysis->getCallExpressions())
+        getTaskTimingControls(*call, visited, controls);
+}
+
 DriverList AnalysisManager::getDrivers(const ValueSymbol& symbol) const {
     return driverTracker.getDrivers(symbol);
 }
--- a/source/analysis/AnalyzedProcedure.cpp
+++ b/source/analysis/AnalyzedProcedure.cpp
@@ -54,6 +54,11 @@ AnalyzedProcedure::AnalyzedProcedure(AnalysisContext& context, const Symbol& ana
             hasSampledValueCalls = true;
     }
 
+    // Store timing control statements from data flow analysis
+    auto dfaTimedStatements = dfa.getTimedStatements();
+    timingControls.insert(timingControls.end(), dfaTimedStatements.begin(),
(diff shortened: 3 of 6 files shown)


MikePopoloski:
Yeah, this needs to be further down the stack because of things like task calls:
```
module m;
  int i, j;
  task t;
    @* i = j;
  endtask
  
  always t();
  
  initial begin
    $monitor(i);
    #10 j = 3;
  end
endmodule
```
Probably it should go in the analysis pass.

AndrewNolte:
So would that be recording timings in handleTiming(), then checking at the end if there were any? Or does the call expression visitor not enter in yet and analyze the subproc?

MikePopoloski:
Probably in AnalyzedProcedure, if it's an `always` block, you want to check if there is no timing and if so, check the collected CallExpressions to see if any called subroutines have timing. The getFunctionDrivers method gets AnalyzedProcedures for the called subroutines so you can extend something off of that.

AndrewNolte:
I couldn't reuse getFunctionDrivers() since it only looks at functions not tasks- but the method I added ended up having a lot of the same logic. Should I split that logic out to analyze all calls first, then filter for functions for driver tracing, and then filter for tasks for the timing check?

source/analysis/AnalysisManager.cpp
```diff
@@ -236,6 +236,46 @@ void AnalysisManager::getFunctionDrivers(const CallExpression& expr, const Symbo
         getFunctionDrivers(*call, containingSymbol, visited, drivers);
 }
 
+void AnalysisManager::getTaskTimingControls(const CallExpression& expr,
+                                            SmallSet<const SubroutineSymbol*, 2>& visited,
+                                            std::vector<const ast::Statement*>& controls) {
+    if (expr.getSubroutineKind() != SubroutineKind::Task) {
```
MikePopoloski: You also need to skip system calls here or else you'll crash on trying to access the wrong element of the variant below.

source/analysis/AnalysisManager.cpp
```diff
@@ -236,6 +236,46 @@ void AnalysisManager::getFunctionDrivers(const CallExpression& expr, const Symbo
         getFunctionDrivers(*call, containingSymbol, visited, drivers);
 }
 
+void AnalysisManager::getTaskTimingControls(const CallExpression& expr,
```
MikePopoloski, line 239: I think this is different enough from getFunctionDrivers that it's ok to just have it separate, even though at first glance there is some overlap.

source/analysis/AnalysisManager.cpp
```diff
@@ -236,6 +236,46 @@ void AnalysisManager::getFunctionDrivers(const CallExpression& expr, const Symbo
...
+        return;
+    }
+
+    // If we've already visited this task then we don't need to
+    // analyze it again.
+    if (!visited.insert(&subroutine).second)
+        return;
+
+    // Get analysis for the task.
+    auto& context = getState().context;
```
MikePopoloski: Move this access of the context inside the branch below where it's needed.

source/analysis/AnalysisManager.cpp
```diff
@@ -236,6 +236,46 @@ void AnalysisManager::getFunctionDrivers(const CallExpression& expr, const Symbo
...
+    auto analysis = getAnalyzedSubroutine(subroutine);
+    if (!analysis) {
+        auto proc = std::make_unique<AnalyzedProcedure>(context, subroutine);
+        analysis = addAnalyzedSubroutine(subroutine, std::move(proc));
+    }
+
+    // Add timing controls from the task to our list.
+    auto taskTimingControls = analysis->getTimingControls();
+    controls.reserve(controls.size() + taskTimingControls.size());
```
MikePopoloski: This could just be `controls.insert(controls.end(), taskTimingControls.begin(), taskTimingControls.end());`

scripts/diagnostics.txt
```diff
@@ -1150,6 +1150,7 @@ error MultipleUDNTDrivers "net '{}' with user-defined nettype '{}' cannot have m
 error InputPortAssign "cannot assign to input port '{}'"
 error ClockVarTargetAssign "cannot write to '{}' with a continuous assignment since it's associated with an output clocking signal"
 error LocalFormalVarMultiAssign "cannot bind local variable '{}' to more than one output or inout local variable formal argument"
+error AlwaysWithoutTimingControl "general always procedures must have timing control, or else it will create a simulation deadlock condition."
```
MikePopoloski: Wording on this: `'always' procedure does not advance time and so will create a simulation deadlock`

source/analysis/AnalyzedProcedure.cpp
```diff
@@ -165,6 +172,33 @@ AnalyzedProcedure::AnalyzedProcedure(AnalysisContext& context, const Symbol& ana
         SmallSet<const SubroutineSymbol*, 2> visited;
         for (auto call : dfaCalls)
             context.manager->getFunctionDrivers(*call, analyzedSymbol, visited, drivers);
+        if (analyzedSymbol.kind == SymbolKind::ProceduralBlock) {
```
MikePopoloski: Can you move this new check up to line 122 as an else where we're already checking `procedure.procedureKind == ProceduralBlockKind::AlwaysComb`

source/analysis/AnalyzedProcedure.cpp
```diff
@@ -54,6 +54,13 @@ AnalyzedProcedure::AnalyzedProcedure(AnalysisContext& context, const Symbol& ana
             hasSampledValueCalls = true;
     }
 
+    // Store timing control statements from data flow analysis
+    auto dfaTimedStatements = dfa.getTimedStatements();
+    timingControls.reserve(dfaTimedStatements.size());
```
MikePopoloski: Since the loop doesn't do anything except add to the vector you can just do `timingControls.insert(timingControls.end(), dfaTimedStatements.begin(), dfaTimedStatements.end());`

MikePopoloski (commented):
This looks pretty good in terms of structure. I guess one remaining thought is that this doesn't account for intra-assignment timing controls. Like for example:
```
module m;
  always clk = #1 ~clk;
endmodule
```

AndrewNolte:
> This looks pretty good in terms of structure. I guess one remaining thought is that this doesn't account for intra-assignment timing controls. Like for example:
> 
> ```
> module m;
>   always clk = #1 ~clk;
> endmodule
> ```

Just added a test for this, and it does seem to handle these

PR #1430, opened 2025-07-03
Clang 15 fixes

akashlevy:
I'm trying to compile an file-list pruning application https://github.com/Silimate/prunefl with AppleClang 15. This application links slang and is running on macOS (Apple M3 pro processor). I get the following errors:
```
prunefl/third_party/slang/source/analysis/DriverTracker.cpp:26:43: error: reference to local binding 'drivers' declared in enclosing function 'slang::analysis::DriverTracker::add'
            for (auto& [driver, bounds] : drivers) {
```
```
prunefl/third_party/slang/source/analysis/DriverTracker.cpp:118:43: error: reference to local binding 'drivers' declared in enclosing function 'slang::analysis::DriverTracker::add'
            for (auto& [driver, bounds] : drivers) {
```
```
prunefl/third_party/slang/source/../include/slang/util/BumpAllocator.h:41:54: error: no matching constructor for initialization of 'slang::ast::Compilation::NetAlias'
        return new (allocate(sizeof(T), alignof(T))) T(std::forward<Args>(args)...);
                                                     ^ ~~~~~~~~~~~~~~~~~~~~~~~~
```
This PR resolves these issues by add binding references and a constructor as necessary.

### DIFF
--- a/include/slang/ast/Compilation.h
+++ b/include/slang/ast/Compilation.h
@@ -1009,6 +1009,9 @@ private:
         AliasBitRange range;
         const Expression* firstExpr;
         const Expression* secondExpr;
+        
+        NetAlias(const Symbol* s, AliasBitRange r, const Expression* fe, const Expression* se)
+            : sym(s), range(r), firstExpr(fe), secondExpr(se) {}
     };
 
     using AliasIntervalMap = IntervalMap<uint64_t, const NetAlias*>;
--- a/include/slang/ast/Compilation.h
+++ b/include/slang/ast/Compilation.h
@@ -1009,9 +1009,9 @@ private:
         AliasBitRange range;
         const Expression* firstExpr;
         const Expression* secondExpr;
-        
-        NetAlias(const Symbol* s, AliasBitRange r, const Expression* fe, const Expression* se)
-            : sym(s), range(r), firstExpr(fe), secondExpr(se) {}
+
+        NetAlias(const Symbol* s, AliasBitRange r, const Expression* fe, const Expression* se) :
+            sym(s), range(r), firstExpr(fe), secondExpr(se) {}
     };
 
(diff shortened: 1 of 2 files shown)


likeamahoney:
From [sv-lang.com](https://sv-lang.com/building.html#build-requirements):

> Minimum supported compiler versions:
> GCC 11
> **clang 17**

akashlevy:
Why not just extend the support to clang 15? The current state prevents slang from being compiled out of the box on macOS. And it's a simple fix

MikePopoloski:
Hmm, this is not actually clang 15 but "AppleClang" 15, installed with Xcode 15, which is actually based on upstream clang 16? When I bumped the Xcode requirement to 16 I remember there being other sources of incompatibilities that were fixed, not just the ones included in this PR. Are you not using the libc++ included with Xcode?

Regardless, I only support building against the latest production release of Xcode (same story with Visual Studio releases on Windows). While the changes here are minimal, I'm only going to be moving forward as Xcode does new releases, taking advantage of the things they fix and the features they add. For example, I'll be switching the C++ language version to c++23 at some point in the next year or so.

I'm not sure what your comment about building out of the box means though -- macOS is my daily driver. You need to install Xcode anyway, why not just install the latest version? Apple forces you to upgrade quickly anyway as they deprecate old versions pretty quickly.

akashlevy:
Hi @MikePopoloski my machine is saying that my XCode is up-to-date for my macOS version (14.4). It's very likely I just need to update my macOS version (to 15.5), which will give access to the latest XCode version with AppleClang 16. I'll give that a try and close this out if that fixes this. Thanks for your time

akashlevy:
Yeah updating macOS allowed me to update XCode, which now uses AppleClang 17. Closing this!

PR #1434, opened 2025-07-10
Rewriter with macro and include expansion

AndrewNolte:
I wanted to add these options to the syntax printer for use in the rewriter and language servers.

These new options conflict with `includePreprocessed` and `includeDirectives`, but these options are widely used in tests, so I'm wondering if I should make a version of SyntaxPrinter for focused on rewriting- specifically spitting out equivalent but different code- for example these expansions, filtering out non-synthesizable constructs, getting rid of package star imports, etc. 

A lot of the options in the existing printer aren't too useful for the rewriter- for example setting includeTrivia to false isn't useful, since it takes out all whitespace and therefore prints invalid code.

Macros invoked in includes weren't being classified as being from includes- so I modified isincludedfileloc to address this case.

Eventually I want to allow for selectively expanding macros- either a single macro usage or a macro name, which I think should be possible without extra metadata. But doing a single macro usage without printing the whole file (language server use case) could be more efficient if ranges of the usage expansion were stored in the syntax tree.

### DIFF
--- a/external/CMakeLists.txt
+++ b/external/CMakeLists.txt
@@ -12,7 +12,7 @@ set(catch2_min_version "3.8")
 # --- fmt lib ---
 set(find_pkg_args "")
 if(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.24.0")
-  set(find_pkg_args "OVERRIDE_FIND_PACKAGE" "${fmt_min_version}")
+  set(find_pkg_args "FIND_PACKAGE_ARGS" "${fmt_min_version}")
 endif()
 
 set(FMT_SYSTEM_HEADERS
--- a/external/CMakeLists.txt
+++ b/external/CMakeLists.txt
@@ -12,7 +12,7 @@ set(catch2_min_version "3.8")
 # --- fmt lib ---
 set(find_pkg_args "")
 if(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.24.0")
-  set(find_pkg_args "FIND_PACKAGE_ARGS" "${fmt_min_version}")
+  set(find_pkg_args "OVERRIDE_FIND_PACKAGE" "${fmt_min_version}")
 endif()
 
 set(FMT_SYSTEM_HEADERS
--- a/include/slang/syntax/SyntaxPrinter.h
+++ b/include/slang/syntax/SyntaxPrinter.h
@@ -69,6 +69,8 @@ public:
     }
 
     /// Sets whether to include preprocessor directives when printing syntax.
+    /// For controlling expansion of macros and includes, use `setExpandIncludes` and
+    /// `setExpandMacros`.
     /// @return a reference to this object, to allow chaining additional method calls.
     SyntaxPrinter& setIncludeDirectives(bool include) {
         includeDirectives = include;
@@ -86,13 +88,6 @@ public:
     /// @return a reference to this object, to allow chaining additional method calls.
     SyntaxPrinter& setExpandMacros(bool expand) {
         expandMacros = expand;
-        return *this;
-    }
-
-    /// Sets conflicting settings for use when expanding macros or includes.
-    /// @return a reference to this object, to allow chaining additional method calls.
-    SyntaxPrinter& setExpansionMode() {
-        includeDirectives = true;
         return *this;
     }
 
--- a/include/slang/syntax/SyntaxPrinter.h
+++ b/include/slang/syntax/SyntaxPrinter.h
@@ -75,13 +75,6 @@ public:
         return *this;
     }
 
-    /// Sets whether to include preprocessor-expanded tokens when printing syntax.
-    /// @return a reference to this object, to allow chaining additional method calls.
-    SyntaxPrinter& setIncludePreprocessed(bool include) {
-        includePreprocessed = include;
-        return *this;
-    }
-
     /// Sets whether to expand include directives when printing syntax.
     /// @return a reference to this object, to allow chaining additional method calls.
     SyntaxPrinter& setExpandIncludes(bool expand) {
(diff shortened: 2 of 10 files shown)


external/CMakeLists.txt
```diff
@@ -12,7 +12,7 @@ set(catch2_min_version "3.8")
 # --- fmt lib ---
 set(find_pkg_args "")
 if(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.24.0")
-  set(find_pkg_args "FIND_PACKAGE_ARGS" "${fmt_min_version}")
+  set(find_pkg_args "OVERRIDE_FIND_PACKAGE" "${fmt_min_version}")
```
MikePopoloski: This change doesn't make any sense to me -- I'm not calling find_package so this just breaks the ability for users to override the fmt dependency with a locally installed version.

include/slang/syntax/SyntaxPrinter.h
```diff
@@ -82,6 +82,27 @@ class SLANG_EXPORT SyntaxPrinter {
...
+    /// Sets whether to expand macro directives when printing syntax.
+    /// @return a reference to this object, to allow chaining additional method calls.
+    SyntaxPrinter& setExpandMacros(bool expand) {
+        expandMacros = expand;
+        return *this;
+    }
+
+    /// Sets conflicting settings for use when expanding macros or includes.
+    /// @return a reference to this object, to allow chaining additional method calls.
+    SyntaxPrinter& setExpansionMode() {
```
MikePopoloski: This method is confusing. What exactly is expansion mode? Why can't users just set the flags they want directly?

AndrewNolte: Yeah this confusion was why I think it makes more sense to make a different printer for the rewriter. The types should really be more like {includeDirectives, includePreproc} | {expandIncludes, expandMacros}. But since the syntax printer is used widely I can't easily change the defaults for these.

AndrewNolte: I can't set these settings just in the expansion methods, because the default behavior for the rewriter should be to not expand anything, but the default behavior for the rewriter is to expand everything

MikePopoloski: I still don't get it. This method is equivalent to `printer.setIncludePreprocessed(false).setIncludeDirectives(true)` so why not let callers just set the flags they care about? You can argue what the defaults should rightly be but it doesn't matter that much because it's not like it's hard to flip some flags.

external/CMakeLists.txt
AndrewNolte: I had some issues here with slang finding the fmt library from an old install, causing the build to fail. I can remove this

include/slang/syntax/SyntaxPrinter.h
AndrewNolte: In my view a syntax printer should print out equivalent code- the existence of includePreprocessed and includeDirectives violates this. If they're both true, it writes a macro usage and its expanded tokens, doubling the code. Or in the case that both includeDirectives and expandIncludes are true, should that print the directive? It's confusing because one property is saying that you write it out, and the other is saying it was expanded, so I wouldn't expect users to know the right combo. The includeDirectives/Preproc knobs and the expandIncludes/Macros knobs just aren't compatible. I think I should not try to awkwardly force this desired property on SyntaxPrinter, and instead make a new class for this with rewriter-focused knobs.

AndrewNolte: I thought about it a bit more and was able to make it work by removing includePreprocessed, and having includeDirectives control non-macro/include directives.

external/CMakeLists.txt
MikePopoloski: This only fixes such a problem because it breaks the ability to find system packages entirely. The right fix for that is probably to tell cmake not to look in install directories for packages with some command line flag.

source/syntax/SyntaxPrinter.cpp
```diff
@@ -175,4 +178,50 @@ SyntaxPrinter& SyntaxPrinter::append(std::string_view text) {
     return *this;
 }
 
+bool SyntaxPrinter::shouldPrint(SourceLocation loc) {
```
MikePopoloski: Function should be `const`

source/syntax/SyntaxPrinter.cpp
```diff
@@ -175,4 +178,50 @@ SyntaxPrinter& SyntaxPrinter::append(std::string_view text) {
...
+        return !sourceManager->isIncludedFileLoc(loc);
+    }
+    else if (sourceManager->isIncludedFileLoc(loc)) {
+        return expandIncludes;
+    }
+    // Not a preprocessed location, so we should print it.
+    return true;
+}
+
+bool SyntaxPrinter::shouldPrint(SyntaxNode& /* DirectiveSyntax& */ syntax) {
```
MikePopoloski: Function and syntax argument should both be `const`. Also I don't understand the comment.

AndrewNolte: It's saying it can be more specific (`DirectiveSyntax&`), but we choose not to so that we don't include `AllSyntax.h`

external/CMakeLists.txt
AndrewNolte: Yeah- I spent an hour trying to find this knob, and every LLM just hallucinated cmake flags. I think a bazel migration might be worth the better hermeticity and documentation. I saw someone else mention bazel on an issue, and one user made a bazel version of slang- https://github.com/hankhsu1996/slang-bazel

MikePopoloski: Pretty sure you can control it by setting https://cmake.org/cmake/help/latest/variable/CMAKE_FIND_USE_INSTALL_PREFIX.html#variable:CMAKE_FIND_USE_INSTALL_PREFIX or https://cmake.org/cmake/help/latest/variable/CMAKE_FIND_NO_INSTALL_PREFIX.html#variable:CMAKE_FIND_NO_INSTALL_PREFIX

MikePopoloski: The find_package docs are pretty clear about what is searched and in what order: https://cmake.org/cmake/help/latest/command/find_package.html

Better to just read the docs than to waste time having an AI lie to you for an hour.

Issue #1439, opened 2025-07-14
Does not build on WebAssembly WASI


whitequark:
**Is your feature request related to the C++ slang project or the pyslang Python bindings?**
C++ slang

**Is your feature request related to a problem? Please describe.**
I would like slang to be ported to WASI, a POSIX subset for WebAssembly.

To be clear, I plan to port it myself. I'm opening this issue as a way to discuss some of the issues I encounter in the process, as there are bits of Slang internals I don't understand.

whitequark:
@MikePopoloski Right off the bat (after fixing fmt https://github.com/fmtlib/fmt/pull/4497 and a weird `boost::uint16_t` typename in boost_unordered.hpp) I get this:

```
[  3%] Building CXX object third_party/slang/source/CMakeFiles/slang_slang.dir/AllSyntax.cpp.obj
In file included from /home/whitequark/Projects/yowasp/yosys/yosys-slang-build/third_party/slang/source/AllSyntax.cpp:8:
In file included from /home/whitequark/Projects/yowasp/yosys/yosys-slang-build/third_party/slang/source/slang/syntax/AllSyntax.h:10:
/home/whitequark/Projects/yowasp/yosys/yosys-slang-src/third_party/slang/source/../include/slang/parsing/Token.h:123:15: error: static assertion failed due to requirement 'sizeof(slang::parsing::Trivia) == 16'
  123 | static_assert(sizeof(Trivia) == 16);
      |               ^~~~~~~~~~~~~~~~~~~~
/home/whitequark/Projects/yowasp/yosys/yosys-slang-src/third_party/slang/source/../include/slang/parsing/Token.h:123:30: note: expression evaluates to '12 == 16'
  123 | static_assert(sizeof(Trivia) == 16);
      |               ~~~~~~~~~~~~~~~^~~~~
/home/whitequark/Projects/yowasp/yosys/yosys-slang-src/third_party/slang/source/../include/slang/parsing/Token.h:224:15: error: static assertion failed due to requirement 'sizeof(slang::parsing::Token) == 16'
  224 | static_assert(sizeof(Token) == 16);
      |               ^~~~~~~~~~~~~~~~~~~
/home/whitequark/Projects/yowasp/yosys/yosys-slang-src/third_party/slang/source/../include/slang/parsing/Token.h:224:29: note: expression evaluates to '12 == 16'
  224 | static_assert(sizeof(Token) == 16);
      |               ~~~~~~~~~~~~~~^~~~~
[...]

PR #1440, opened 2025-07-16
Address portability to multi-threaded WebAssembly/WASI

whitequark:
Part of addressing #1439.

### DIFF
--- a/source/parsing/ParserBase.cpp
+++ b/source/parsing/ParserBase.cpp
@@ -109,7 +109,7 @@ Token ParserBase::expect(TokenKind kind) {
             // become unbalanced and flush it.
             openDelims.clear();
         }
-        lastPoppedDelims = {};
+        lastPoppedDelims = {Token(), Token()};
     }
 
     Token result = Token::createExpected(alloc, getDiagnostics(), peek(), kind, window.lastConsumed,
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -31,6 +31,8 @@ jobs:
             os: windows-latest
           - preset: win64-debug-noexcept
             os: windows-latest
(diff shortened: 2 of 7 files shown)


whitequark:
I'm going to try and add CI support for WASI.

whitequark:
OK, I've added CI support. It doesn't build tools or tests because that's a bunch of additional work I'm not sure I can justify doing right now (you'd need a Wasm runtime, Catch2 doesn't even build under WASI as it is, etc), and the assumption is that people building for WASI would only be interested in Slang as a library.

source/parsing/ParserBase.cpp
```diff
@@ -109,7 +109,7 @@ Token ParserBase::expect(TokenKind kind) {
             // become unbalanced and flush it.
             openDelims.clear();
         }
-        lastPoppedDelims = {};
```
whitequark, line 112: I have no idea why this is needed but this line causes an error under Clang. (I don't have the exact error recorded but it was something about an ambiguous overload.)

whitequark:
It looks like everything works with multi-threaded WASI target. Unfortunately until a solution for https://github.com/WebAssembly/wasi-sdk/issues/546 is a part of a WASI SDK release (most likely WASI SDK 26 or 27, possibly 28) it will not be possible to build the follow-up PR introducing SLANG_USE_THREADS on a single-threaded WASI target.

whitequark:
@MikePopoloski To confirm, the only blocker for merging this PR is a release of the fmt library, is that correct?

MikePopoloski:
Yes, that's right.

whitequark:
I've made the PR use the latest {fmt} release, as suggested by an {fmt} maintainer. Should be fine now!

whitequark:
Thanks!

PR #1441, opened 2025-07-16
Address portability to single-threaded WebAssembly/WASI

### LINKED ISSUE #1439: Does not build on WebAssembly WASI

whitequark:
Depends on #1440.
Fixes #1439.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -70,6 +70,7 @@ option(SLANG_CI_BUILD "Enable longer running tests for CI builds" OFF)
 option(SLANG_FUZZ_TARGET "Enables changes to make binaries easier to fuzz test"
        OFF)
 option(BUILD_SHARED_LIBS "Generate shared libraries instead of static" OFF)
+option(SLANG_USE_THREADS "Enable use of threads" ON)
 option(SLANG_USE_MIMALLOC "Enable use of the mimalloc library" ON)
 option(SLANG_USE_CPPTRACE "Enable use of the cpptrace library" OFF)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -145,6 +145,11 @@ if(CMAKE_CXX_COMPILER_ID MATCHES "MSVC")
   add_compile_options(/utf-8 /bigobj /permissive-)
 endif()
 
+if(CMAKE_SYSTEM_NAME MATCHES "WASI")
+  add_compile_definitions(_WASI_EMULATED_SIGNAL)
+  add_link_options(-lwasi-emulated-signal)
+endif()
+
 if(SLANG_FUZZ_TARGET)
   add_compile_definitions(SLANG_ASSERT_ENABLED)
 endif()
--- a/external/CMakeLists.txt
+++ b/external/CMakeLists.txt
@@ -22,7 +22,7 @@ set(FMT_SYSTEM_HEADERS
 FetchContent_Declare(
   fmt
   GIT_REPOSITORY https://github.com/fmtlib/fmt.git
-  GIT_TAG 11.2.0
+  GIT_TAG 553ec11ec06fbe0beebfbb45f9dc3c9eabd83d28
   GIT_SHALLOW ON
   ${find_pkg_args})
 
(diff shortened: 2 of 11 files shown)


external/CMakeLists.txt
```diff
@@ -22,7 +22,7 @@ set(FMT_SYSTEM_HEADERS
 FetchContent_Declare(
   fmt
   GIT_REPOSITORY https://github.com/fmtlib/fmt.git
-  GIT_TAG 11.2.0
+  GIT_TAG 553ec11ec06fbe0beebfbb45f9dc3c9eabd83d28
```
MikePopoloski: This is fine for exposition but I don't want to land this until there's an official release containing the fix.

external/boost_unordered.hpp
```diff
@@ -5728,11 +5728,11 @@ struct group15 {
     inline int match(std::size_t hash) const { return match_impl(reduced_hash(hash)); }
 
     inline bool is_not_overflowed(std::size_t hash) const {
-        return !(reinterpret_cast<const boost::uint16_t*>(m)[hash % 8] & 0x8000u);
+        return !(reinterpret_cast<const uint16_t*>(m)[hash % 8] & 0x8000u);
```
MikePopoloski: Technically should be fixed in the upstream project but this is fine, I'll just remember to check this the next time I update for a boost release.

CMakeLists.txt
```diff
@@ -70,6 +70,7 @@ option(SLANG_CI_BUILD "Enable longer running tests for CI builds" OFF)
 option(SLANG_FUZZ_TARGET "Enables changes to make binaries easier to fuzz test"
        OFF)
 option(BUILD_SHARED_LIBS "Generate shared libraries instead of static" OFF)
+option(SLANG_USE_THREADS "Enable use of threads" ON)
```
MikePopoloski, line 73: Can you add some documentation for this option to building.dox?

MikePopoloski (commented):
I'm still surprised you don't get compile errors for the use of the thread pool in the Driver. To answer your question from elsewhere, the Driver class is a wrapper around common command line options, source file loading, and running the compiler through the various steps (parsing, compilation, analysis, etc). Not necessary since you can use all the library components on their own, but I think most people who use slang as a library do use the Driver.

MikePopoloski:
Oops, that review was meant to be on the other PR.

whitequark:
> I'm still surprised you don't get compile errors for the use of the thread pool in the Driver.

For my local build, this is down to turning off tools for the WASI build.

For the CI build, it's because right now I don't have a single-threaded build enabled, due to WASI SDK not shipping such a build. (Right now it requires patching the build scripts, and there isn't a consensus yet on how to ship it exactly, though we're close to getting it.)

CMakeLists.txt
whitequark: Done

whitequark:
> I'm still surprised you don't get compile errors for the use of the thread pool in the Driver. To answer your question from elsewhere, the Driver class is a wrapper around common command line options, source file loading, and running the compiler through the various steps (parsing, compilation, analysis, etc). Not necessary since you can use all the library components on their own, but I think most people who use slang as a library do use the Driver.

Okay, I think I understand what's going on here. Let me try to explain.

Since WebAssembly/wasi-sdk#546 has been fixed, the WASI SDK makes concurrency-related headers like `<pthread.h>` and `<thread>` and `<mutex>` available even in single-thread-only builds, and all of the concurrency operations except for spawning and joining threads succeed with their specified semantics. So it's actually OK to `#include <BS_thread_pool.hpp>` just like it's OK to include `<mutex>`, you just can't spawn any threads.

It looks like `Driver.cpp` doesn't even use anything from `BS_thread_pool.hpp`, so I went ahead and removed the `#include`. It would be fine to remove the `#ifdef` guard in `AnalysisManager.h`, but I left it in since everything else threadpool-related is compiled out in single-threaded builds as well.

Also, I just realized that I can build both Slang tools and Slang unit tests for Wasm, so let me also add a commit that enables both and runs tests on CI.

whitequark:
@MikePopoloski OK, I consider this PR to be complete. All functionality works, all tests pass, I've also verified manually that the tools (slang, slang-tidy, slang-hier, slang-reflect, rewriter) all function as expected with a small testcase.

whitequark:
Thank you!

Issue #1433, opened 2025-07-09
Spurious unresolved hierarchical path error


garrinkimmell:
**Is your bug report related to the C++ slang project or the pyslang Python bindings?**
C++ slang

**Describe the bug**

A reference to a seemingly-valid signal is raising an error.

```
$ slang-hier --version
slang version 8.1.314+4e3f4140
$ slang-hier --top top --single-unit --color-diagnostics=false repro.sv
Module="top" Instance="top" File="repro.sv"
Module="example" Instance="top.dut" File="repro.sv"
Module="inner" Instance="top.dut.gg_inner[0].genblk1.i" File="repro.sv"
Module="inner" Instance="top.dut.gg_inner[1].genblk1.i" File="repro.sv"
Top level design units:
    top

repro.sv:34:40: error: could not resolve hierarchical path name 'genblk1'
   wire reference = top.dut.gg_inner[0].genblk1.i.innermost;
                                       ^~~~~~~~

Build failed: 1 error, 0 warnings
```

**To Reproduce**

The code below should be sufficient to reproduce this:

```
module inner(
             input logic clk,
             input logic resetn);
   // Simple module, inner wire declaration which we will reference from top.
   wire innermost = '1;

endmodule; // inner

module example(
               input logic clk,
               input logic resetn
               );

   // Module with a silly generate loop that includes an generate-time `if`.
   generate
      for (genvar g_inner=0; g_inner<2; g_inner++) begin : gg_inner
           if (g_inner == 0) begin
              inner i(.*);
           end else if (g_inner == 1) begin
              inner i(.*);
           end;
       end : gg_inner
   endgenerate;
endmodule; // example

module top();

    wire clk;
    wire  resetn;

   example dut(.*);

   // The reference to this hierarchial path fails.
   wire reference = top.dut.gg_inner[0].genblk1.i.innermost;


endmodule; // top
```

MikePopoloski:
`genblk1` is a system generated name for an anonymous generate block. You can't reference an anonymous block from within the Verilog code.

garrinkimmell:
Got it. I realize that it's dodgy to rely on the generated name, but unfortunately it's a pattern that appears several times in a design I'm verifying. 

I'll close the issue. Thanks again for taking a look, and for all your work on slang.

MikePopoloski:
It's possible to add a compatibility flag to allow this, if you have an existing design that works with some commercial tools that support this extension.

toddstrader:
[...]

PR #1445, opened 2025-07-22
Enable lookupName() access to unnamed generate blocks

toddstrader:
This demonstrates the inability to lookup unnamed generate blocks via `lookupName()`.  It is true that the RTL itself cannot hierarchically refer to such a scope, but `27.6` outlines how external interfaces may refer to it.  Assuming the slang API is an external interface, it would seem good to allow for access to these scopes as well via `lookupName()`.

The problem (if you'll call it that) is that during elaboration `Scope::insertMember()` gets a symbol for the unnamed generate block symbol with no name and therefore does not add it to `nameMap`.  Slang obviously knows how to construct the external names via `getExternalName()`.  So I'm wondering if the correct approach is to alter the `name` of the symbol once this is computed and before it is inserted as a member of its parent.  Also, is this behavior generaically desired or should it be hidden behind an elaboration parameter?  Alternatively we could also make this a `LookupFlags` option, but then we'd need to build a second name map to house external name.


MikePopoloski:
Makes sense to want this. It also sort of came up recently in this issue: https://github.com/MikePopoloski/slang/issues/1433

I think the right approach is to use the external name if no name is provided, when the symbol is created. Let it exist in the name map like any other symbol. Upon lookup, if there is a manually added LookupFlags::AllowUnnamedGenerate or if there is a compilation flag set (for allowing command line-level compatibility settings) then allow it, otherwise give a descriptive error.

toddstrader:
Doesn't add flags yet, just want to make sure this is the right track.  a) are flags actually needed? b) would we need both elaboration and lookup flags?

source/ast/Scope.cpp
```diff
@@ -1165,6 +1177,26 @@ void Scope::elaborate() const {
         }
     }
 
+    // Unnamed generate blocks must be handled last to detect genblk name collisions
+    for (auto symbol : unnamedGenblks) {
```
AndrewNolte, line 1195: I think it should be gated behind a flag in lookup, maybe ::ExternalLookup. Then it can add a better diag if a reference was attempted, telling the user to name the block.

We could also add a test on all.sv which queries all symbols, and verify that they are reachable from an external lookup via their hierarchical path. I think typedefs will also struggle to pass this test, and may need a check for this flag as well.

toddstrader: Went with `AllowUnnamedGenerate` as Mike suggested.  Did not do the all.sv test, I'll leave that for a subsequent PR.

include/slang/ast/Lookup.h
```diff
@@ -83,6 +83,9 @@ enum class SLANG_EXPORT LookupFlags {
     /// the $unit compilation unit scope.
     DisallowUnitReferences = 1 << 14,
 
+    /// Allow unnamed generate blocks to be looked up by their external name.
+    AllowUnnamedGenerate = 1 << 15,
+
```
MikePopoloski, line 88: You need to update the SLANG_BITMASK macro below to reference this since it is now the highest valued flag. Also this should be manually added to the Python bindings in SymbolBindings.cpp

source/ast/Scope.cpp
```diff
@@ -1165,6 +1177,26 @@ void Scope::elaborate() const {
...
+        auto updateName = [&](Symbol* symbol, const std::string& externalName) {
+            char* mem = (char*)getCompilation().allocate(externalName.size(), 1);
+            memcpy(mem, externalName.data(), externalName.size());
+            symbol->name = std::string_view(mem, externalName.size());
+            nameMapEmplace(symbol, true);
+        };
+        if (auto block = symbol->as_if<GenerateBlockSymbol>()) {
+            if (block->name.empty())
```
MikePopoloski: The name is always empty here since you checked above before adding to the list, so you don't need to check again.

source/ast/Scope.cpp
```diff
@@ -1165,6 +1177,26 @@ void Scope::elaborate() const {
...
+                block->isUnnamed = true;
```
MikePopoloski: It seems like isUnnamed should be settable in the GenerateBlockSymbol constructor so you don't need to mess with it elsewhere.

source/ast/Scope.cpp
```diff
@@ -620,6 +620,16 @@ static bool canLookupByName(SymbolKind kind) {
     }
 }
 
+void Scope::nameMapEmplace(const Symbol* member, bool isElaborating) const {
```
MikePopoloski: I don't think you need to factor this out; when you call this for the genblks you already know the name is not empty, it canBeLookedUpByName, and that there won't be any name conflicts.

source/ast/Lookup.cpp
```diff
@@ -457,6 +457,24 @@ bool lookupDownward(std::span<const NamePlusLoc> nameParts, NameComponents name,
 
         auto& scope = symbol->as<Scope>();
         symbol = scope.find(name.text);
+        if (symbol) {
```
MikePopoloski: This needs to be checked up above, prior to descending into the next name component, otherwise you won't error if the genblk is the first element in the name chain. Probably you should put this into the checkClassParams helper and rename it to something more appropriate.

Hmm, also when this happens you need to null out the result and then return true from this function so that lookup can continue in other scopes, otherwise you break upward name lookup for previously valid cases. For example:
```
module m;
    int a;
endmodule

module top;
    m genblk1();
    n n1();
endmodule

module n;
    if (1) begin
        wire a = 1;
    end

    int b = genblk1.a;
endmodule
```
This should still pass even with your change. Note that passing your new flag is a behavior change, and influences whether b is assigned from the wire in the genblk or the variable in the module up the hierarchy.

I'm also worried about a similar problem for unqualified name lookup. This example still needs to work too:
```
struct { int a; } genblk1;

module m;
    if (1) begin end

    int b = genblk1.a;
endmodule
```

MikePopoloski: Or even just:
```
int genblk1;

module m;
    if (1) begin end

    int b = genblk1;
endmodule
```

toddstrader: Added the tests.  The genblk-first was definitely a problem.  The other two are fine because no lookup will yield the unnamed generate blocks without the new flag.  Presumably we never want `genblk#` in the verilog to reference a generate block that only has an external name.

MikePopoloski:
I would still expect these tests to fail because you don't qualify on the flag inside Lookup's unqualifiedImpl method. I was confused why they did not, and then realized it's because you only implemented the handling for loop generates, not if and case generates in Scope::elaborate.

toddstrader:
Yeah, you're right.  I think that's all resolved now.

One bit of wonkiness with the warning (I changed it from an error to a warning) is that it means different things depending on whether you're using the API or compiling.  If you're using the API it means "yeah, the thing you're looking for exists but it's under an unnamed generate so I'm not going to give it to you because you didn't supply `LookupFlags::AllowUnnamedGenerate`".  But if you're compiling it means "you're referencing something which has the name of an unnamed generate block if you were coming from an external interface, but you're not so I'm just going to warn you about this regardless of whether your code is technically correct or not".  I tried to capture some of this in the warning message.

The latter should not fail to compile (as you pointed out with a few examples) so I think this has to be a warning.

toddstrader:
I added some words locally, but I used the simplest example for the docs:
```
int genblk1;

module m;
    if (1) begin end

    int b;
    always_comb b = genblk1;
endmodule
```
and ran into a different (possibly preexisting problem).  More to follow later.

toddstrader:
The problem was mine and not preexisting.  That's fixed now that I'm checking for unnamed generates in both `lookupDownward()` and `Lookup::unqualifiedImpl()`.  I could reproduce the docs failure I was seeing earlier and have gotten past that step (running the warnings examples) but am now hitting some `doxygen.py` issue so I'm just going to throw it back to the GitHub CI and hope for the best.

toddstrader:
```
Build with Pip (ubuntu-latest, 3.11)
The hosted runner lost communication with the server. Anything in your workflow that terminates the runner process, starves it for CPU/Memory, or blocks its network access can cause this error.
```
I think that python failure may just need to be rerun, but I don't see a way to do it myself.  I've also tried to run this locally but am running into different (probably environmental) problems.

PR #1446, opened 2025-07-23
Add --depfile-trim flag for depfile generation

AndrewNolte:
This trims unreferenced modules that are passed in, and does a topo sort. This topo sort of packages is needed for most tools, so this enabled just using slang to order files correctly


include/slang/parsing/ParserMetadata.h
```diff
@@ -60,6 +61,20 @@ struct SLANG_EXPORT ParserMetadata {
...
+    /// Visits top level symbols declared in this metadata, calling the provided function with the
+    /// name. This includes modules, classes, packages, and interfaces.
+    void visitDeclaredSymbols(const std::function<void(std::string_view)>& func) const;
+
+    /// Returns a list of all symbols from visitReferencedSymbols
+    std::vector<std::string_view> getReferencedSymbols() const;
+
+    /// Visits all top level symbols referenced/used by this metadata, calling the provided
+    /// function for each symbol name.
+    void visitReferencedSymbols(const std::function<void(std::string_view)>& func) const;
```
AndrewNolte: I could use ranges/views here, but the iterator code is pretty ugly. I think it's better to wait for C++23 generators

PR #1447, opened 2025-07-23
[rewriter] Add --squash-blanklines

AndrewNolte:
The issue with --squash-newlines is that when macros are expanded and directives are ignored, the blank lines with tabs add up and don't get squashed, since they're all coming from different directives, resulting in sometimes a massive amount of blank lines at the beginning. This solves that by doing a pass at the end to squash these.

This is a use case for the rewriter as running the preprocessor in slang for other tools, and could be set as default for `-E,--preprocess` as well.

### DIFF
--- a/include/slang/syntax/SyntaxPrinter.h
+++ b/include/slang/syntax/SyntaxPrinter.h
@@ -105,8 +105,15 @@ public:
         return *this;
     }
 
+    /// Sets whether to squash blank lines down into one when printing syntax.
+    /// @return a reference to this object, to allow chaining additional method calls.
+    SyntaxPrinter& setSquashBlanklines(bool squash) {
+        squashBlankLines = squash;
+        return *this;
+    }
+
     /// @return a copy of the internal text buffer.
-    std::string str() const { return buffer; }
+    std::string str() const;
 
     /// A helper method that assists in printing an entire syntax tree back to source
     /// text. A SyntaxPrinter with useful defaults is constructed, the tree is printed,
@@ -127,6 +134,7 @@ private:
     bool expandMacros = false;
     bool includeComments = true;
     bool squashNewlines = true;
+    bool squashBlankLines = false;
 };
 
 } // namespace slang::syntax
--- a/tools/rewriter/rewriter.cpp
+++ b/tools/rewriter/rewriter.cpp
@@ -62,8 +62,8 @@ int main(int argc, char** argv) {
 
         // Trivia options
         driver.cmdLine.add("--exclude-comments", excludeComments, "Exclude comments in output");
-        driver.cmdLine.add("--squash-newlines", squashNewlines,
-                           "Squash adjacent newlines into one");
+        driver.cmdLine.add("--squash-blanklines", squashNewlines,
+                           "Squash adjacent blank lines into one");
 
         // Missing/skipped node options
         driver.cmdLine.add("--include-missing", includeMissing,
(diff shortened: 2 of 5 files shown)


include/slang/syntax/SyntaxPrinter.h
```diff
@@ -105,8 +105,15 @@ class SLANG_EXPORT SyntaxPrinter {
         return *this;
     }
 
+    /// Sets whether to squash blank lines down into one when printing syntax.
+    /// @return a reference to this object, to allow chaining additional method calls.
+    SyntaxPrinter& setSquashBlanklines(bool squash) {
```
MikePopoloski: This doesn't seem like it needs to be a flag. It's really just a utility function that you can run on the result of the printer -- it takes a string, removes extra newlines, returns the result. Kind of unrelated to syntax printing at all really.

AndrewNolte: Moved it to rewriter

tools/rewriter/rewriter.cpp
```diff
@@ -26,6 +27,58 @@ using namespace slang;
...
+            }
+            // For empty lines, don't include whitespace - just make it truly empty
+            lastLineWasEmpty = true;
+        }
+        // Skip subsequent empty/whitespace-only lines in sequence
+
+        firstLine = false;
+    }
+
+    // Handle the case where input doesn't end with newline
```
MikePopoloski: This comment seems like nonsense; you're handling the case where the input *does* end with a newline?

The whole `firstLine` thing seems very clunky. Just always append a newline char when you add a new line to the result, and then you don't need this branch down here either.

tools/rewriter/rewriter.cpp
```diff
@@ -26,6 +27,58 @@ using namespace slang;
...
+    std::string result;
+    std::string line;
+    bool lastLineWasEmpty = false;
+    bool firstLine = true;
+
+    result.reserve(input.size());
+
+    while (std::getline(stream, line)) {
+        // Check if line is empty or contains only whitespace
+        bool isEmpty = line.empty();
```
MikePopoloski: `const bool isEmpty = std::ranges::all_of(line, isWhitespace);`

tools/rewriter/rewriter.cpp
```diff
@@ -26,6 +29,49 @@ using namespace slang;
...
+        }
+    }
+
+    // Process remaining lines
+    while (std::getline(stream, line)) {
+        const bool isEmpty = std::ranges::all_of(line, isWhitespace);
+
+        if (!isEmpty) {
+            // Line has content, always include it with full indentation
+            result.push_back('\n');
```
MikePopoloski: If you just flip the order of these two lines you don't need the separate code blocks before and after the loop.

tools/rewriter/rewriter.cpp
```diff
@@ -26,6 +29,35 @@ using namespace slang;
...
+    // Process lines and write back in-place
+    while (std::getline(stream, line)) {
+        const bool isEmpty = std::ranges::all_of(line, isWhitespace);
+
+        if (!isEmpty) {
+            // Line has content, always include it with full indentation
+            std::copy(line.begin(), line.end(),
+                      text.begin() + static_cast<std::ptrdiff_t>(writePos));
+            writePos += line.size();
+            text[writePos++] = '\n';
```
MikePopoloski: Hmm, I think what you had before was fine (using a separate result variable instead of trying to reuse the input). I think now you have a potential buffer overrun if the input string does not end in a newline.

MikePopoloski:
Bump?

AndrewNolte:
rebased and dropped the inplace commit

Issue #1321, opened 2025-04-12
Add a `slang --cst-json` CLI argument/action


parker-research:
**Is your feature request related to the C++ slang project or the pyslang Python bindings?**
C++ slang (CLI tool)

**Is your feature request related to a problem? Please describe.**
As discussed in other recent issues, slang has `slang --ast-json <json_file>` to dump the AST (post-elaboration), but does not have functionality to dump the Concrete Syntax Tree (CST), aka "parse tree", which represents the `SyntaxNode` tree. 

For debugging visitors, building rewriters, etc., it would be helpful to be able to dump this early-in-the-pipeline representation. 

**Describe the solution you'd like**
I'd like to see the `slang --cst-json` CLI arg be implemented, similar to the `slang --ast-json` CLI arg.

**Describe alternatives you've considered**
The solution to #1320 (coming soon) is a viable alternative, albeit in Pyslang only.

**Additional context**
This feature may not be super commonly used/requested, and that's fine. Just wanted to share it anyway to centralize the discussion on its existence, in case anyone else ever wants it as well.

parker-research:
Thanks a lot, @AndrewNolte! Very neat!

PR #1448, opened 2025-07-25
Implement --cst-json

AndrewNolte:
Round trip verified with `build/bin/slang tests/regression/all.sv --cst-json all.json` then `python scripts/reconstruct_from_json.py all.json --compare tests/regression/all.sv`

- Addresses https://github.com/MikePopoloski/slang/issues/1321
- Added flags for different levels of verbosity with the cst output, defaulting to most verbose.
- Also added cmdline addEnum function for adding slang enums to arg parsers

### DIFF
--- a/docs/command-line-ref.dox
+++ b/docs/command-line-ref.dox
@@ -283,6 +283,25 @@ When dumping AST to JSON, include source line and file information.
 When dumping AST to JSON, include expanded type information in the output instead of
 condensing all types into a human-friendly string.
 
+`--cst-json <file>`
+
+Dump the concrete syntax trees (CST) in JSON format to the specified file, or '-' for stdout.
+The CST represents the parsed structure of the source code including all tokens, whitespace,
+and comments exactly as they appear in the source.
+
+`--cst-json-mode <mode>`
+
+Controls the level of detail included in the CST JSON output. The available modes in order of most to least detailed are:
+
+| Mode | Description |
+|------|-------------|
+| `full` | Full token objects with complete trivia information including whitespace, comments, and formatting details |
+| `simple-trivia` | Full token objects with trivia simplified to concatenated strings instead of detailed arrays and kinds |
+| `no-trivia` | Full token objects but with all trivia objects excluded from the output |
+| `simple-tokens` | Tokens represented as simple strings with no trivia or token kind information |
+
+The default mode is `full` if this option is not specified.
+
 @section compilation-limits Compilation
 
 `--top <name>`
--- a/include/slang/syntax/CSTSerializer.h
+++ b/include/slang/syntax/CSTSerializer.h
@@ -1,0 +1,39 @@
+//------------------------------------------------------------------------------
+//! @file CSTSerializer.h
+//! @brief Concrete Syntax Tree JSON serialization
+//
+// SPDX-FileCopyrightText: Michael Popoloski
+// SPDX-License-Identifier: MIT
+//------------------------------------------------------------------------------
+#pragma once
+
+#include "slang/syntax/SyntaxNode.h"
+#include "slang/text/Json.h"
+
+namespace slang::syntax {
+
+class SyntaxTree;
+
+#define MODE(x) x(Full) x(SimpleTrivia) x(NoTrivia) x(SimpleTokens)
+SLANG_ENUM(CSTJsonMode, MODE)
+#undef MODE
+
+/// Converts concrete syntax trees to JSON format for debugging and analysis
+class SLANG_EXPORT CSTSerializer {
+public:
+    explicit CSTSerializer(JsonWriter& writer, CSTJsonMode mode = CSTJsonMode::Full);
+
+    /// Serialize a syntax tree to JSON
+    void serialize(const SyntaxTree& tree);
+
+    /// Serialize a syntax node to JSON
+    void serialize(const SyntaxNode& node);
+
+private:
+    void visitToken(parsing::Token token);
+    void writeTokenTrivia(parsing::Token token);
+
+    JsonWriter& writer;
+    CSTJsonMode mode;
+};
+} // namespace slang::syntax
--- a/pyslang/tests/test_cst_json.py
+++ b/pyslang/tests/test_cst_json.py
@@ -1,0 +1,238 @@
+# SPDX-FileCopyrightText: Michael Popoloski
+# SPDX-License-Identifier: MIT
+
+import json
+from functools import cache
+from typing import Any, Dict, List, Union
+
+import pyslang
+
+
+def to_dict(tree, mode: pyslang.CSTJsonMode):
+    json_str = tree.to_json(mode)
+    return json.loads(json_str)
+
+
+def get_enum_names(enum_class) -> set:
+    return set([k for k in enum_class.__members__])
+
+
+SYNTAX_KINDS = get_enum_names(pyslang.SyntaxKind)
+TOKEN_KINDS = get_enum_names(pyslang.TokenKind)
+TRIVIA_KINDS = get_enum_names(pyslang.TriviaKind)
+
+
+class CSTValidator:
+    """Validates CST JSON structure properties based on serialization mode."""
+
+    def __init__(self, mode: pyslang.CSTJsonMode):
+        self.mode = mode
+        self.errors = list[str]()
+
+    def validate(self, json_data: Any, path: str = "root") -> bool:
+        """Validate CST JSON data and return True if valid."""
+        self.errors = []
+        self._validate_node(json_data, path)
+        return len(self.errors) == 0
+
+    def get_errors(self) -> List[str]:
+        """Get list of validation errors."""
+        return self.errors.copy()
+
+    def _error(self, message: str, path: str):
+        """Record a validation error."""
+        self.errors.append(f"{path}: {message}")
+
+    def _validate_node(self, node: Any, path: str):
+        """Validate a single node in the CST."""
+        if not isinstance(node, dict):
+            self._error(f"Expected dict, got {type(node)}", path)
+            return
+
+        # Every node must have a 'kind' field
+        if "kind" not in node:
+            self._error("Missing 'kind' field", path)
+            return
+
+        kind = node["kind"]
+
+        # Validate kind value
+        if not isinstance(kind, str):
+            self._error(f"'kind' must be string, got {type(kind)}", path)
+            return
+
+        # Handle special cases that are not in the enum
+        special_kinds = {"SyntaxTree"}
+
+        if kind not in (SYNTAX_KINDS | TOKEN_KINDS | special_kinds):
+            self._error(f"Unknown kind '{kind}'", path)
+
+        # Validate based on node type
+        if kind in TOKEN_KINDS:
+            self._validate_token(node, path)
+        elif kind in SYNTAX_KINDS or kind in special_kinds:
+            self._validate_syntax_node(node, path)
+
+        # Validate trivia constraints based on mode
+        self._validate_trivia_constraints(node, path)
+
+    def _validate_token(self, token: Dict[str, Any], path: str):
+        """Validate a token node."""
+        kind = token["kind"]
+
+        if self.mode == pyslang.CSTJsonMode.SimpleTokens:
+            # In SimpleTokens mode, some tokens might be collapsed to strings
+            return
+
+        # Token must have 'text' field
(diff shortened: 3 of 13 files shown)


scripts/syntax_gen.py
```diff
@@ -1255,5 +1257,250 @@ def generatePyBindings(builddir, alltypes):
...
+    try:
+        os.makedirs(headerdir)
+    except OSError:
+        pass
+
+    # Start the header file.
+    outf = open(os.path.join(headerdir, "CSTJsonConverter.h"), "w")
+    outf.write(
+        """//------------------------------------------------------------------------------
+//! @file CSTJsonConverter.h
```
MikePopoloski: I would prefer that this be called CSTSerializer to match ASTSerializer. Also it looks like this class doesn't need to be generated. You can just put it in the library proper and only generate the CSTJsonVisitor into its own header that gets included by CSTSerializer.cpp.

tools/driver/slang_main.cpp
```diff
@@ -102,6 +124,12 @@ int driverMain(int argc, TArgs argv) {
             "Dump the compiled AST in JSON format to the specified file, or '-' for stdout",
             "<file>", CommandLineFlags::FilePath);
 
+        std::optional<std::string> cstJsonFile;
+        driver.cmdLine.add(
+            "--cst-json", cstJsonFile,
```
MikePopoloski, line 130: Please add documentation for this.

MikePopoloski:
You should add your test case to the regress CTest list. Also probably want to expose to Python.

tools/driver/slang_main.cpp
```diff
@@ -214,6 +249,36 @@ int driverMain(int argc, TArgs argv) {
...
+                        cstMode = syntax::CSTJsonMode::SimpleTrivia;
+                    }
+                    else if (*cstJsonMode == "no-trivia") {
+                        cstMode = syntax::CSTJsonMode::NoTrivia;
+                    }
+                    else if (*cstJsonMode == "simple-tokens") {
+                        cstMode = syntax::CSTJsonMode::SimpleTokens;
+                    }
+                    else {
+                        std::cerr << "Error: Invalid value for --cst-json-mode: " << *cstJsonMode
```
MikePopoloski: Please don't use `cerr` -- follow the convention used in this file.

tools/driver/slang_main.cpp
```diff
@@ -214,6 +249,36 @@ int driverMain(int argc, TArgs argv) {
                               driver.serializeDepfiles(driver.getDepfiles(), depfileTarget));
             }
 
+            if (cstJsonFile) {
+                TimeTraceScope timeScope("cstSerialization"sv, ""sv);
+                syntax::CSTJsonMode cstMode = syntax::CSTJsonMode::Full;
+
+                // Handle CST JSON mode
+                if (cstJsonMode) {
+                    if (*cstJsonMode == "full") {
```
MikePopoloski: You don't need the `*` since std::optional's operator== will delegate to the underlying string.

tests/regression/CMakeLists.txt
```diff
@@ -10,6 +10,26 @@ add_test(NAME regression_wire_module
...
                  "--ast-json=-" "--ast-json-detailed-types")
+add_test(NAME regression_all_file_syntax
+         COMMAND slang::driver "${CMAKE_CURRENT_LIST_DIR}/all.sv"
+                 "--cst-json=-")
+
+# CST JSON round trip test
+add_test(
+  NAME regression_cst_json_roundtrip
+  COMMAND
+    python3 ${CMAKE_SOURCE_DIR}/scripts/reconstruct_from_json.py
```
MikePopoloski: Use `${Python_EXECUTABLE}` instead of `python3`

tests/regression/CMakeLists.txt
```diff
@@ -10,6 +10,26 @@ add_test(NAME regression_wire_module
 add_test(NAME regression_all_file
          COMMAND slang::driver "${CMAKE_CURRENT_LIST_DIR}/all.sv"
                  "--ast-json=-" "--ast-json-detailed-types")
+add_test(NAME regression_all_file_syntax
```
MikePopoloski: It seems like this test duplicates the one below that save the results to a file?

source/syntax/CSTSerializer.cpp
```diff
@@ -0,0 +1,202 @@
...
+// SPDX-License-Identifier: MIT
+//------------------------------------------------------------------------------
+#include "slang/syntax/CSTSerializer.h"
+
+#include "slang/parsing/Token.h"
+#include "slang/syntax/AllSyntax.h"
+#include "slang/syntax/SyntaxTree.h"
+#include "slang/syntax/SyntaxVisitor.h"
+
+// This file contains CST JSON conversion functionality.
```
MikePopoloski: This comment is pointless, it's already mentioned in the header comment.

docs/command-line-ref.dox
```diff
@@ -283,6 +283,10 @@ When dumping AST to JSON, include source line and file information.
 When dumping AST to JSON, include expanded type information in the output instead of
 condensing all types into a human-friendly string.
 
+`--cst-json <file>`
```
MikePopoloski, line 286: You're missing documentation for the mode flag.

include/slang/syntax/CSTSerializer.h
```diff
@@ -0,0 +1,52 @@
+
```
MikePopoloski: Remove random newline.

include/slang/syntax/CSTSerializer.h
```diff
@@ -0,0 +1,52 @@
...
+#include "slang/syntax/SyntaxNode.h"
+#include "slang/text/Json.h"
+
+namespace slang::syntax {
+
+class SyntaxTree;
+
+/// CST JSON output formatting modes
+enum class SLANG_EXPORT CSTJsonMode {
```
MikePopoloski: Can you not just use the `SLANG_ENUM` system instead of manually defining all the traits stuff?

pyslang/tests/test_cst_json.py
```diff
@@ -0,0 +1,91 @@
...
+    json_full = root_node.to_json(pyslang.CSTJsonMode.Full)
+    json_simple = root_node.to_json(pyslang.CSTJsonMode.SimpleTokens)
+
+    assert isinstance(json_full, str)
+    assert isinstance(json_simple, str)
+    assert "ModuleDeclaration" in json_full
+    assert "ModuleDeclaration" in json_simple
+
+
+def test_cst_json_mode_enum():
```
MikePopoloski: This test seems like nonsense...

pyslang/tests/test_cst_json.py
```diff
@@ -0,0 +1,91 @@
...
+        initial $display("Hello");
+    endmodule
+    """
+
+    tree = pyslang.SyntaxTree.fromText(test_code)
+
+    # Test default serialization (full mode)
+    json_str = tree.to_json()
+    assert isinstance(json_str, str)
+    assert "SyntaxTree" in json_str
```
MikePopoloski: These tests are not very good. The output is deterministic, why not assert what the actual JSON string should be? Otherwise maybe the output is just garbage.

pyslang/tests/test_cst_json.py
```diff
@@ -0,0 +1,91 @@
...
+    # Test full mode (default)
+    json_full = tree.to_json(pyslang.CSTJsonMode.Full)
+    assert "trivia" in json_full
+
+    # Test simple trivia mode
+    json_simple_trivia = tree.to_json(pyslang.CSTJsonMode.SimpleTrivia)
+    assert isinstance(json_simple_trivia, str)
+
+    # Test no trivia mode
+    json_no_trivia = tree.to_json(pyslang.CSTJsonMode.NoTrivia)
```
MikePopoloski: Same with all of these.

source/CMakeLists.txt
```diff
@@ -100,6 +101,7 @@ add_library(
   syntax/SyntaxPrinter.cpp
   syntax/SyntaxTree.cpp
   syntax/SyntaxVisitor.cpp
+  syntax/CSTSerializer.cpp
```
MikePopoloski: nit: please keep the file listing in alphabetical order

source/syntax/CSTSerializer.cpp
```diff
@@ -0,0 +1,202 @@
...
+    SLANG_UNREACHABLE;
+}
+
+CSTSerializer::CSTSerializer(JsonWriter& writer, CSTJsonMode mode) : writer(writer), mode(mode) {
+}
+
+void CSTSerializer::serialize(const SyntaxTree& tree) {
+    writer.startObject();
+    writer.writeProperty("kind");
+    writer.writeValue(std::string_view("SyntaxTree"));
```
MikePopoloski: Seems weird to always write "SyntaxTree" here... also why the explicit string_view constructor call?

source/syntax/CSTSerializer.cpp
```diff
@@ -0,0 +1,202 @@
...
+        writer.writeProperty("kind");
+        writer.writeValue(std::string_view(toString(node.kind)));
+    }
+
+    void writePropertyAndToken(std::string_view name, parsing::Token token) {
+        writer.writeProperty(name);
+        visitToken(token);
+    }
+
+    void writePropertyAndNode(std::string_view name, const SyntaxNode* node) {
```
MikePopoloski: If the node is not optional here it should be passed by reference.

source/syntax/CSTSerializer.cpp
```diff
@@ -0,0 +1,202 @@
...
+    CSTJsonVisitor(JsonWriter& w, CSTJsonMode m) : writer(w), mode(m) {}
+
+    // Helper methods for common patterns
+    void startSyntaxObject(const SyntaxNode& node) {
+        writer.startObject();
+        writer.writeProperty("kind");
+        writer.writeValue(std::string_view(toString(node.kind)));
+    }
+
+    void writePropertyAndToken(std::string_view name, parsing::Token token) {
```
MikePopoloski: The naming of these functions is weird... like the token is the property value yes? And then the optional variants below are not PropertyAnd***. It seems like you don't need both anyway. Just always check for optionality?

source/syntax/CSTSerializer.cpp
```diff
@@ -0,0 +1,202 @@
...
+    JsonWriter& writer;
+    CSTJsonMode mode;
+
+    CSTJsonVisitor(JsonWriter& w, CSTJsonMode m) : writer(w), mode(m) {}
+
+    // Helper methods for common patterns
+    void startSyntaxObject(const SyntaxNode& node) {
+        writer.startObject();
+        writer.writeProperty("kind");
+        writer.writeValue(std::string_view(toString(node.kind)));
```
MikePopoloski: Another annoying explicit call to std::string_view.

source/syntax/CSTSerializer.cpp
```diff
@@ -0,0 +1,202 @@
...
+    void writeOptionalToken(std::string_view name, parsing::Token token) {
+        if (token) {
+            writer.writeProperty(name);
+            visitToken(token);
+        }
+    }
+
+    void writeTokenList(std::string_view name, const TokenList& tokenList) {
+        writer.writeProperty(name);
+        if (tokenList.empty()) {
```
MikePopoloski, line 88: If the list is empty should you just skip writing the property altogether? That's what you're doing in the other optional calls above.

source/syntax/CSTSerializer.cpp
```diff
@@ -0,0 +1,202 @@
...
+    void visitToken(parsing::Token token) {
+        // If token text is empty, write null instead of an object
+        if (token.rawText().empty()) {
+            writer.writeNull();
+            return;
+        }
+
+        // If simple-tokens mode, just write the text value
+        if (mode == CSTJsonMode::SimpleTokens) {
+            writer.writeValue(std::string_view(token.rawText()));
```
MikePopoloski: rawText() already returns a string_view, why is it repeated here again?

scripts/syntax_gen.py
```diff
@@ -1255,5 +1257,91 @@ def generatePyBindings(builddir, alltypes):
         outf.write("}\n")
 
 
+def generateCSTJson(builddir, alltypes):
+    # Make sure the output directory exists.
```
MikePopoloski: Vestigial comment.

PR #1457, opened 2025-08-06
Fix compilation corrupting parent pointers for type params passed to insts

AndrewNolte:
The compilation flow sometimes makes new syntax nodes; if any children are also syntax nodes, this repoints the parent pointers at the compilation alloced one, which is bad. This is fixed for this one case by deep cloning the child, but there may be more cases.

- Adds visitor test for all.sv to validate those parent pointers
- Removes template from syntax visitor's visitDefault(), which increased build time, and on some clang builds also increases binary size.
- Adds makeSyntaxVisitor impls for simple single method handlers

### DIFF
--- a/include/slang/syntax/SyntaxVisitor.h
+++ b/include/slang/syntax/SyntaxVisitor.h
@@ -83,7 +83,9 @@ auto makeSyntaxVisitor(Functions... funcs) {
     return Result(std::move(funcs)...);
 }
 
-/// Simple visitors with kinds that are left to the user to evaluate.
+/// @brief Simple visitors with kinds that are left to the user to evaluate.
+/// This is preferable over makeSyntaxVisitor([](auto, auto){...}), as the compile time cost is
+/// lower, and the compiler errors are better.
 struct AllSyntaxVisitor {
     AllSyntaxVisitor(std::function<void(const SyntaxNode&)> func) : handler(std::move(func)) {}
 
@@ -99,22 +101,6 @@ struct AllSyntaxVisitor {
 private:
     std::function<void(const SyntaxNode&)> handler;
 };
-
-/// @brief Creates a SyntaxVisitor with the provided handler function.
-///
-/// @param func The handler function to called during the preorder traversal of the syntax tree.
-///
-/// For example:
-/// @code
-/// auto visitor = makeSyntaxVisitor([&](const SyntaxNode& node) {
-///     std::cout << node.kind << std::endl;
-/// });
-/// visitor.visit(tree.root());
-/// @endcode
-///
-inline AllSyntaxVisitor makeAllSyntaxVisitor(std::function<void(const SyntaxNode&)> func) {
-    return AllSyntaxVisitor(std::move(func));
-}
 
 namespace detail {
 
--- a/include/slang/syntax/SyntaxVisitor.h
+++ b/include/slang/syntax/SyntaxVisitor.h
@@ -7,9 +7,11 @@
 //------------------------------------------------------------------------------
 #pragma once
 
+#include <functional>
 #include <vector>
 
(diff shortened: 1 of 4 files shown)


include/slang/syntax/SyntaxVisitor.h
```diff
@@ -82,6 +83,39 @@ auto makeSyntaxVisitor(Functions... funcs) {
...
+///
+/// For example:
+/// @code
+/// auto visitor = makeSyntaxVisitor([&](const SyntaxNode& node) {
+///     std::cout << node.kind << std::endl;
+/// });
+/// visitor.visit(tree.root());
+/// @endcode
+///
+inline AllSyntaxVisitor makeAllSyntaxVisitor(std::function<void(const SyntaxNode&)> func) {
```
MikePopoloski: Hmm, what is the point of this helper function? Normally you use one of these to get template inference, but here there's no template parameter. You can just substitute any call to this function with a call to the constructor instead.

source/ast/symbols/ParameterBuilder.cpp
```diff
@@ -182,10 +182,8 @@ const ParameterSymbolBase& ParameterBuilder::createParam(
             // a type parameter, so fix it up into a NamedTypeSyntax to get a type from it.
             tt.addFlags(DeclaredTypeFlags::TypeOverridden);
             if (NameSyntax::isKind(newInitializer->kind)) {
-                // const_cast is ugly but safe here, we're only going to refer to it
```
MikePopoloski, line 185: I guess the const_cast was not safe. Oops!

tests/unittests/parsing/VisitorTests.cpp
```diff
@@ -733,6 +736,65 @@ class C; endclass
...
+            for (size_t i = 0; i < node.getChildCount(); i++) {
+                auto child = node.childNode(i);
+                if (!child)
+                    continue;
+                if (child->parent != &node) {
+                    valid = false;
+                    auto parentKind = toString(child->parent->kind);
+                    auto childKind = toString(child->kind);
+                    auto expectedParent = toString(node.kind);
+                    fmt::print("Parent pointer mismatch with `{}`. {}.parent should be {}, "
```
MikePopoloski: Prefer using catch2's `INFO` macro so this gets categorized under the correct test case.

tests/unittests/parsing/VisitorTests.cpp
```diff
@@ -733,6 +736,65 @@ class C; endclass
...
+
+    fs::path path = findTestDir();
+    path /= "../../regression/all.sv";
+    auto mTree = SyntaxTree::fromFile(path.string());
+    REQUIRE(mTree);
+    auto tree = *mTree;
+
+    REQUIRE(validateParents(*tree));
+
+    // Store the printed text
```
MikePopoloski: This comment doesn't make sense to me.

include/slang/syntax/SyntaxVisitor.h
AndrewNolte: Ah true

include/slang/syntax/SyntaxVisitor.h
```diff
@@ -82,6 +83,39 @@ auto makeSyntaxVisitor(Functions... funcs) {
     return Result(std::move(funcs)...);
 }
 
+/// Simple visitors with kinds that are left to the user to evaluate.
```
AndrewNolte: I realized while amending that the makeSyntaxVisitor and makeVisitor code examples don't make sense... they don't call node.visit on the created visitors

MikePopoloski: Eh, that's assumed context. The example snippet is very small. If you wanted to be very complete you'd need a whole example program in there which seems unnecessary.

PR #1465, opened 2025-08-13
[slang] Fix eval script session segmentation fault

likeamahoney:
This bug was founded using `slang` with this command line:

```
slang "-GA=fork=L:for" ../top.sv
```

It's not depends on content of `.sv` file. It is related to the fact that syntax trees in `SciptSession` are not validated before ast building at the moment.

I provide a fix for this but idk is it needs to show diagnostics from syntax parser in `ScriptSession` or not?


MikePopoloski:
Better fix in 352c532a6b316456d457fc326e32a18dbcbabe5b

Issue #1356, opened 2025-05-15
Generic `interconnect` bit width is always 0 with `Type::getBitstreamWidth()`


akashlevy:
**Is your bug report related to the C++ slang project or the pyslang Python bindings?**
C++ slang and `yosys-slang`

**Describe the bug**
There does not seem to be a way to dynamically resolve the width of an `UntypedType`. Calling `Type::getBitstreamWidth()` on an `UntypedType` always returns 0.

To fix this, you would probably need to implement dynamic resolution in a similar fashion to how SV `class` width is implemented. I don't currently see any methods to resolve currently. Curious to hear your thoughts.

**To Reproduce**
See https://github.com/povik/yosys-slang/issues/146

**Additional context**
Context is that the `interconnect` width needs to be resolved for `yosys-slang` to be able to dump into RTLIL.

MikePopoloski:
I think the idea is that interconnect nets aren't real and don't ever have an actual width. They don't end up in the final simulation or synthesized output, you have to resolve them to a real net somewhere.

23.3.3.7.1 describes how they get merged (in not very great detail). In particular:
> It shall be illegal for the type of a simulated net (see 23.3.3.7) at the end of elaboration to be an
interconnect net.

I think slang is probably not implementing that check right now. There is logic to do the merging of net types in `PortSymbol::checkSimulatedNetTypes` that could be rearranged into something you can use more generally. Note that every "bit" of the connection can be a different net type!

akashlevy:
I was not aware of `PortSymbol::checkSimulatedNetTypes`, thanks for the helpful pointer.

MikePopoloski:
I was looking at this some more (because of the linked PR) and I think there's more to do here beyond just checks for interconnect nets, so here are some notes for later. slang implements the dissimilar net type checks listed in the LRM but only the checks; what you really want is a way to query what the actual net type of each net should be, which because of the port collapsing rules means it may not be the net type that was written on the declaration.

For example:
```
module netlist;
  wire w;
  dut1 child1(w);
  dut2 child2(w);
  initial #10 $display(w);
endmodule
module dut1(wand w);
  assign w = 1;
endmodule
module dut2(wor w);
  assign w = 0;
endmodule
```
[...]

PR #1466, opened 2025-08-13
[slang] Add a check for simulated interconnect nets

likeamahoney:
Closes https://github.com/MikePopoloski/slang/issues/1356


MikePopoloski:
See extended discussion in #1356 -- this only solves a small piece of the problem; the whole simulated net thing needs to be reworked a bit.

PR #1470, opened 2025-08-18
[slang-tidy][style] Ignore null ports for OnlyANSIPortDecl checker

likeamahoney:
`OnlyANSIPortDecl` checker segfaults on modules with null ports

### DIFF
--- a/tools/tidy/src/style/OnlyANSIPortDecl.cpp
+++ b/tools/tidy/src/style/OnlyANSIPortDecl.cpp
@@ -16,7 +16,7 @@ struct MainVisitor : public TidyVisitor, ASTVisitor<MainVisitor, true, false, fa
     void handle(const PortSymbol& port) {
         NEEDS_SKIP_SYMBOL(port)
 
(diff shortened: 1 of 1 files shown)


tools/tidy/src/style/OnlyANSIPortDecl.cpp
```diff
@@ -16,7 +16,7 @@ struct MainVisitor : public TidyVisitor, ASTVisitor<MainVisitor, true, false, fa
     void handle(const PortSymbol& port) {
         NEEDS_SKIP_SYMBOL(port)
 
-        if (!port.isAnsiPort) {
+        if (!port.isAnsiPort && !port.isNullPort) {
```
MikePopoloski: This should check `internalSymbol` directly, since there are other cases that can lead to a non-ansi port with no connected internal symbol.

likeamahoney: fixed

PR #1477, opened 2025-08-20
check for loops always within generate

perrintong325:


perrintong325:
whoops my bad should be for my own fork

PR #1478, opened 2025-08-20
[slang-tidy] Fix --code option validating

likeamahoney:
In this PR I will try to fix `slang-tidy` `--code` option possible validating issue.

Because at the moment, it gives unclear errors or crashes with an exception. For example for this case:

```
slang-tidy --code -h
```

### DIFF
--- a/include/slang/util/String.h
+++ b/include/slang/util/String.h
@@ -39,7 +39,8 @@ std::optional<int32_t> strToInt(std::string_view str, size_t* pos = nullptr, int
 /// source string that was not part of the conversion.
 ///
 /// @return the converted value, or nullopt if conversion fails (invalid string, overflow, etc).
-std::optional<uint32_t> strToUInt(std::string_view str, size_t* pos = nullptr, int base = 10);
+SLANG_EXPORT std::optional<uint32_t> strToUInt(std::string_view str, size_t* pos = nullptr,
+                                               int base = 10);
 
 /// Parses the provided @a str into a floating point value.
 /// If @a pos is non-null, it will be set to point to the first character in the


tools/tidy/src/tidy.cpp
```diff
@@ -96,21 +96,28 @@ int main(int argc, char** argv) {
 
         // Get the ID and kind from the check code string
         auto hypenPos = infoCode->find('-');
-        if (hypenPos == std::string::npos) {
+        if (hypenPos == std::string::npos || (!infoCode->empty() && infoCode->at(0) == '-')) {
```
MikePopoloski: Could just be `|| hypenPos == 0`

Also please rename this variable to be correctly spelled while you're here: `hyphen`

tools/tidy/src/tidy.cpp
```diff
@@ -96,21 +96,28 @@ int main(int argc, char** argv) {
...
         auto kind = tidyKindFromStr(kindStr);
-        auto id = stoull(infoCode->substr(hypenPos + 1));
-
         if (!kind) {
             OS::printE(fmt::format("Check kind {} does not exist\n", kindStr));
             return 1;
         }
 
+        uint64_t id;
+        SLANG_TRY {
```
MikePopoloski: Using a method that throws exceptions for cases like this isn't great. You should use one of the standard library methods that reports errors more directly. Alternatively, String.h already includes a helper method `strToUInt` to wrap that for you.

tools/tidy/src/tidy.cpp
likeamahoney: fixed and variable renamed

tools/tidy/src/tidy.cpp
likeamahoney: great suggestion

I tried to keep the original idea of ​​the author with calling this library function, but I also thought that throwing exceptions outside was a bad idea

fix is done

PR #1486, opened 2025-08-21
disable MSAN instrumentation for SmallVector destructor

cc10512:
Because of the destruction order when handling small vectors, MSAN is throwing a `use-of-uninitialized-value` error. The best explantation we came up with is that:
- first `SmallVector` releases the stackBase memory
- MSAN poisons the freed memory
- the `SmallVectorBase` destructor calls cleanup, which does a check for `isSmall()` which accesses the poisoned memory locations.

This "fix" essentially creates an explicit destructor for SmallVector which is annotated to disable MSAN checks because we are quite confident that the underlying memory released is still available for the cleanup check.

### DIFF
--- a/include/slang/util/SmallVector.h
+++ b/include/slang/util/SmallVector.h
@@ -636,6 +636,18 @@ public:
         }
     }
 
+#if defined(__has_feature)
+#    if __has_feature(memory_sanitizer)
+    // Elide msan check for SmallVector destructor.
+    //
+    // There is an issue with the destruction order: SmallVector frees the
+    // underlying storage (stackBase) and then cleanup accesses data in that
+    // region but MSAN has poisoned it already.
+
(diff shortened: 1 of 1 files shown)


MikePopoloski:
Thanks for the PR. Perhaps a better fix is to make SmallVector's destructor be the one to call `std::ranges::destroy(*this);` -- I suspect this is what is actually triggering MSAN. `isSmall` only accesses members of the base class so should be safe even after the child class is destructed.

So basically:
```
~SmallVectorBase() {
    if (!isSmall())
            ::operator delete(data_);
}
```
and
```
~SmallVector() {
    std::ranges::destroy(*this);
}
```

cc10512:
Thank you Mike! 

Your proposed solution is indeed better. It passes the MSAN tests so I updated the PR.

PR #1511, opened 2025-09-10
Add accessors for modport and instance tracking to the driver tracker

jameshanlon:


include/slang/analysis/AnalysisManager.h
```diff
@@ -140,6 +140,13 @@ class SLANG_EXPORT AnalysisManager {
     /// Returns all of the known drivers for the given symbol.
     DriverList getDrivers(const ast::ValueSymbol& symbol) const;
 
+    /// Returns all of the known drivers for the given modport symbol.
+    DriverList getModportDrivers(const ast::ValueSymbol& symbol) const;
```
MikePopoloski: Instaed of adding a new method it probably makes more sense to just merge the functionality into `getDrivers` by having it explicitly check for a ModportPort and returning from the other map if so.

include/slang/analysis/DriverTracker.h
```diff
@@ -33,6 +33,20 @@ class AnalyzedProcedure;
 /// A helper class that tracks drivers for all symbols in a thread-safe manner.
 class DriverTracker {
 public:
+    // State tracked per canonical instance.
+    struct InstanceState {
```
MikePopoloski: Given that this is now part of the public analysis API it should move out of DriverTracker and probably rename to something like `InstanceDriverState`. Also needs to be marked `SLANG_EXPORT` and use Doxygen triple comments for docs.

source/analysis/DriverTracker.cpp
```diff
@@ -241,6 +241,26 @@ DriverList DriverTracker::getDrivers(const ValueSymbol& symbol) const {
...
+        for (auto& [driverSymbol, bounds] : item.second) {
+            drivers.emplace_back(driverSymbol, bounds);
+        }
+    });
+    return drivers;
+}
+
+std::optional<DriverTracker::InstanceState> DriverTracker::getInstanceState(
+    const InstanceBodySymbol& symbol) const {
+    if (!instanceMap.contains(&symbol)) {
```
MikePopoloski: You can skip the whole `contains` check by making `state` be of type `std::optional<InstanceState>` and then just returning it after the cvisit call.

include/slang/analysis/AnalysisManager.h
jameshanlon: Fixed.

include/slang/analysis/DriverTracker.h
jameshanlon: Fixed.

source/analysis/DriverTracker.cpp
jameshanlon: Fixed.

source/analysis/DriverTracker.cpp
```diff
@@ -234,13 +235,32 @@ void DriverTracker::addDrivers(AnalysisContext& context, DriverAlloc& driverAllo
 
 DriverList DriverTracker::getDrivers(const ValueSymbol& symbol) const {
     DriverList drivers;
+
+    // If the symbol is a modport member, then return from the modport drivers.
+    if (symbol.isKind(SymbolKind::Modport)) {
```
MikePopoloski: This should be `SymbolKind::ModportPort`

jameshanlon: Not confusing at all!

I checked it and thought I had it the right way...

jameshanlon:
I will update the PR with a renaming of `InstanceState` to `InstanceDriverState`.

PR #1518, opened 2025-09-15
[slang] Fix number parser segmentation fault

likeamahoney:
Hello, `slang` number parser segfaults on this example:

```verilog
`pragma D's 11111111111111111110's 1.`pragma I's
```

Which happens due to this [code](https://github.com/MikePopoloski/slang/blob/master/include/slang/parsing/NumberParser.h#L117) in which `peek` can rewrite the value of `valid` value from `false` to `true` in nested calls to `numberParser` method which leads to avoiding of next number parser error handling process.

I don't know should I to add a test for this and which test category should I choose for this

### DIFF
--- a/include/slang/parsing/NumberParser.h
+++ b/include/slang/parsing/NumberParser.h
@@ -91,10 +91,16 @@ public:
         Token next = first;
         firstLocation = first.location();
 
+        // To restore previously matched number parsing status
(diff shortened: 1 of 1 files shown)


likeamahoney:
reimplemented my fix - I think this variant is ​​more correct

MikePopoloski:
Fixed in df1bd4e00b294ff84a2f46e0309f5d626b76511e

PR #1542, opened 2025-10-17
Split out text output functionality from Driver

AndrewNolte:
This will allow non-text based clients to:
  - receive errors from the driver
  - Not add cli options specific to text formatting,  json output, or depfiles
  - Not instantiate text or json diag clients and have the engine spend runtime sending to those clients.
 
The constructor takes a single diagnostics client to ensure at least one client is seeing the cli parse errors.

In a followup pr, the json client will be amended  to emit these parse errors in its output.

This was done in two commits to more easily verify the split. The diff is best viewed from the second commit

### DIFF
--- a/include/slang/driver/BaseDriver.h
+++ b/include/slang/driver/BaseDriver.h
@@ -1,7 +1,6 @@
 //------------------------------------------------------------------------------
-//! @file Driver.h
-//! @brief Top-level handler for processing arguments and
-//! constructing a compilation for a CLI tool.
+//! @file BaseDriver.h
+//! @brief Base driver class for argument parsing and compilation.
 //
 // SPDX-FileCopyrightText: Michael Popoloski
 // SPDX-License-Identifier: MIT
@@ -52,25 +51,13 @@ namespace slang::driver {
 SLANG_ENUM(CompatMode, COMPAT)
 #undef COMPAT
 
-/// @brief A top-level class that handles argument parsing, option preparation,
+/// @brief Base driver class that handles argument parsing, option preparation,
 /// and invoking various parts of the slang compilation process.
 ///
 /// This is exposed as a convenience wrapper around the various components
 /// that could otherwise be used on their own.
 ///
-/// A typical compilation flow using the driver looks as follows:
-///
-/// @code{.cpp}
-/// Driver driver;
-/// driver.addStandardArgs();
-/// if (!driver.parseCommandLine(someStr)) { ...error }
-/// if (!driver.processOptions()) { ...error }
-/// if (!driver.parseAllSources()) { ...error }
-/// if (!driver.runFullCompilation()) { ...error }
-/// else { ...success }
-/// @endcode
-///
-class SLANG_EXPORT Driver {
+class SLANG_EXPORT BaseDriver {
 private:
     // This exists to ensure we get a Compilation object created prior to anything else,
     // such as the DiagnosticEngine, which wants a Compilation to register callbacks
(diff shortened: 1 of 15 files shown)


source/diagnostics/TextDiagnosticClient.cpp
```diff
@@ -360,4 +361,33 @@ void TextDiagnosticClient::formatDiag(SourceLocation loc, std::span<const Source
...
+            break;
+        default:
+            OS::printE(message);
+            OS::printE("\n");
+            return;
+    }
+
+    OS::printE(fg(color), prefix);
+    OS::printE(message);
+    OS::printE("\n");
```
AndrewNolte, line 390: This does match the previous behavior, but now makes the text diag client aware of the OS. Not sure why the other diags need to be stored in an intermediate string; I think it's nice to print them out as they come.

source/driver/BaseDriver.cpp
```diff
@@ -443,17 +389,18 @@ void Driver::addStandardArgs() {
...
 }
 
-[[nodiscard]] bool Driver::parseCommandLine(std::string_view argList,
-                                            CommandLine::ParseOptions parseOptions) {
+[[nodiscard]] bool BaseDriver::parseCommandLine(std::string_view argList,
+                                                CommandLine::ParseOptions parseOptions) {
     if (!cmdLine.parse(argList, parseOptions)) {
         for (auto& err : cmdLine.getErrors())
-            OS::printE(fmt::format("{}\n", err));
+            printError(err);
```
AndrewNolte, line 401: This would be the one change in output behavior, printing `error: slang: ...` instead of `slang: ...` on parse errors. We could add another enum to the severity to account for this if we want to keep the old format.

source/driver/BaseDriver.cpp
```diff
@@ -0,0 +1,858 @@
...
+    Diagnostics optionDiags = diagEngine.setWarningOptions(options.warningOptions);
+    for (auto& diag : optionDiags)
+        diagEngine.issue(diag);
+
+    return true;
+}
+
+bool BaseDriver::parseAllSources() {
+    syntaxTrees = sourceLoader.loadAndParseSources(createParseOptionBag());
+    if (!reportLoadErrors())
```
AndrewNolte, line 710: I also noticed that this branch will never get hit; `processOptions()` checks this, so `parseAllSources()` won't get called.

AndrewNolte:
Updated basedriver constructor to require a diag client

include/slang/driver/BaseDriver.h
```diff
@@ -426,18 +325,38 @@ class SLANG_EXPORT Driver {
...
     bool anyFailedLoads = false;
     flat_hash_set<std::filesystem::path> activeCommandFiles;
     std::vector<std::tuple<std::string_view, std::string_view, std::string_view>>
         translateOffFormats;
-    std::unique_ptr<JsonWriter> jsonWriter;
+};
+
+template<typename ClientT>
+    requires std::derived_from<ClientT, DiagnosticClient>
+class ClientOwner {
```
MikePopoloski, line 345: I think I'd prefer the original way of not having a constructor argument if it means we don't need this stuff here.

source/diagnostics/TextDiagnosticClient.cpp
```diff
@@ -360,4 +361,33 @@ void TextDiagnosticClient::formatDiag(SourceLocation loc, std::span<const Source
     buffer->append("\n"sv);
 }
 
+void TextDiagnosticClient::reportMessage(DiagnosticSeverity severity, const std::string& message) {
+    fmt::terminal_color color;
+    std::string_view prefix;
+
+    switch (severity) {
```
MikePopoloski, line 368: There are already functions for getSeverityString and getSeverityColor that should be used instead of duplicating here.

source/diagnostics/TextDiagnosticClient.cpp
MikePopoloski: It's nice if there are many diagnostics all issued at once to buffer them up and output in one go rather than an individual syscall to output each one by one.

This method should use the buffer just like all other methods in the class, otherwise the usage is very confusing. I could see the utility of factoring out the stringification logic and having different clients, one that outputs messages as they come and one that buffers first, but I don't think it's super necessary for the Driver use cases. Probably just keeping the existing buffering logic will be fine, the delay shouldn't be large because the cases where the Driver outputs a message are usually due to some config error that is going to cause an early exit.

AndrewNolte:
Will reimplement with a virtual printError on the driver class

PR #1544, opened 2025-10-20
Put tree loading functionality into loadTrees()

AndrewNolte:
Allows for reusing this logic when providing alternate ways to find a symbol name.
Also exposes `findBuffer()`, allowing for those users to also use the standard -y flag.

### DIFF
--- a/include/slang/driver/SourceLoader.h
+++ b/include/slang/driver/SourceLoader.h
@@ -10,7 +10,6 @@
 #include <deque>
 #include <filesystem>
 #include <memory>
-#include <mutex>
 #include <optional>
 #include <span>
 #include <vector>
@@ -19,6 +18,7 @@
 #include "slang/text/Glob.h"
 #include "slang/text/SourceLocation.h"
 #include "slang/util/FlatMap.h"
+#include "slang/util/Function.h"
 #include "slang/util/Util.h"
 
 namespace slang {
@@ -151,6 +151,16 @@ public:
     /// Gets a pointer to the source library with the given name, or adds it if
     /// it does not exist. Returns nullptr if @a name is empty.
     SourceLibrary* getOrAddLibrary(std::string_view name);
+
+    /// Find a source buffer by searching through libraries added via -y
+    SourceBuffer findBuffer(std::string_view name) const;
+
+    /// Load trees using a custom buffer finder function
+    /// Result is stored in the same syntaxTree list
+    static void loadTrees(
+        SyntaxTreeList& syntaxTrees, function_ref<SourceBuffer(std::string_view)> findBufferFunc,
+        SourceManager& sourceManager, const Bag& optionBag,
+        std::span<const syntax::DefineDirectiveSyntax* const> inheritedMacros = {});
 
 private:
     // One entry per unit of files + options to compile them.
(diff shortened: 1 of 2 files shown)


include/slang/driver/SourceLoader.h
```diff
@@ -152,6 +159,64 @@ class SLANG_EXPORT SourceLoader {
     /// it does not exist. Returns nullptr if @a name is empty.
     SourceLibrary* getOrAddLibrary(std::string_view name);
 
+    /// Find a source buffer by searching through libraries added via -y
+    SourceBuffer findBuffer(std::string_view name);
+
+    /// Load trees using a custom buffer finder function (templated with concept constraint)
+    template<BufferFinder FindBufferFunc>
+    static void loadTrees(
```
MikePopoloski, line 160: I think I'd prefer that you use a `function_ref` argument instead of templating on the callback so that the implementation can remain hidden. I think the tiny perf overhead here doesn't matter.

include/slang/driver/SourceLoader.h
```diff
@@ -152,6 +159,64 @@ class SLANG_EXPORT SourceLoader {
...
+        SyntaxTreeList& syntaxTrees, FindBufferFunc findBufferFunc, SourceManager& sourceManager,
```
MikePopoloski: const SyntaxTreeList&

source/driver/SourceLoader.cpp
```diff
@@ -184,6 +184,31 @@ std::vector<SourceBuffer> SourceLoader::loadSources() {
     return results;
 }
 
+SourceBuffer SourceLoader::findBuffer(std::string_view name) {
```
MikePopoloski: Looks like this function can be marked `const`

source/driver/SourceLoader.cpp
```diff
@@ -184,6 +184,31 @@ std::vector<SourceBuffer> SourceLoader::loadSources() {
...
+        path /= name;
+
+        for (auto& ext : searchExtensions) {
+            path.replace_extension(ext);
+            if (!sourceManager.isCached(path)) {
+                // This file is never part of a library because if
+                // it was we would have already loaded it earlier.
+                auto readResult = sourceManager.readSource(path, /* library */ nullptr);
+                if (readResult) {
+                    buffer = *readResult;
```
MikePopoloski: Can just do a `return` here and skip all the `break`s.

include/slang/driver/SourceLoader.h
```diff
@@ -238,6 +303,7 @@ class SLANG_EXPORT SourceLoader {
     std::vector<std::string> errors;
     SyntaxTreeList libraryMapTrees;
 
+public:
```
MikePopoloski: Why was this made public?

AndrewNolte: Was using this for the indexer and source loader; I'll revert this and make other constants.

include/slang/driver/SourceLoader.h
AndrewNolte: Additional trees are added to this list though.

MikePopoloski: Oh yeah sorry, missed that.

PR #1551, opened 2025-10-27
Add syntaxes for empty instance arrays and primitive symbols

AndrewNolte:


include/slang/ast/symbols/InstanceSymbols.h
```diff
@@ -247,8 +247,9 @@ class SLANG_EXPORT InstanceArraySymbol final : public Symbol, public Scope {
     /// the root array's name. Otherwise returns the name of this symbol itself.
     std::string_view getArrayName() const;
 
-    static InstanceArraySymbol& createEmpty(Compilation& compilation, std::string_view name,
-                                            SourceLocation loc);
+    static InstanceArraySymbol& createEmpty(
```
MikePopoloski: You can just call setSyntax() on the result of the call, no need for the optional parameter.

PR #1554, opened 2025-10-28
Install boost_concurrent if Boost is not found

furkanusta:
If you are interested, I can try to add a CMake target that runs a simple program with the installed files similar to what simdjson is doing https://github.com/simdjson/simdjson/blob/0112be86b0dec7fd979d10f0cf405c1197181769/tests/installation_tests/find/CMakeLists.txt


MikePopoloski:
Thanks for the PR. There is an install test here: https://github.com/MikePopoloski/slang_package_example but it doesn't try to include this specific header so it didn't catch the issue.

PR #1558, opened 2025-10-30
Handle wildcard-import diamonds that re-export the same declaration

ebrevdo:
# Summary
Teach the lookup engine to recognise "diamond" wildcard imports that ultimately forward the same
declaration. When every leg of the import tree resolves to an identical origin—whether via explicit
imports, multi-hop localparams, or typedef chains—we now accept the reference instead of hard
failing, while leaving the existing diagnostic intact for genuine conflicts. New fixtures and driver
coverage exercise the previously failing shapes (explicit re-export, multi-hop forward, typedef),
and the mixed-case test still confirms that real discrepancies are rejected. This keeps the language
semantics strict by default yet unblocks common layered-package patterns that were formerly
impossible to model.

# Details
- Compute canonical import origins in `Lookup::unqualifiedImpl` by recursively following explicit imports, conversions, and typedef targets.
- Treat a wildcard name as safe when all candidates share that canonical symbol; otherwise emit the
  original `AmbiguousWildcardImport` error with the full note fan-out.
- Add dedicated fixture packages for explicit re-export, multi-hop forward, and typedef diamonds, plus the mixed collision case.
- Extend driver tests to cover the three newly supported diamonds and the remaining failing scenario.

### DIFF
--- a/source/ast/Lookup.cpp
+++ b/source/ast/Lookup.cpp
@@ -2096,8 +2096,8 @@ void Lookup::unqualifiedImpl(const Scope& scope, std::string_view name, LookupLo
                         flat_hash_set<const Symbol*> noted;
                         for (size_t i = 0; i < imports.size(); i++) {
                             const auto& pair = imports[i];
-                            const Symbol* target =
-                                canonicalTargets[i] ? canonicalTargets[i] : pair.imported;
+                            const Symbol* target = canonicalTargets[i] ? canonicalTargets[i]
+                                                                       : pair.imported;
                             if (!noted.emplace(target).second)
                                 continue;
 
@@ -2120,8 +2120,8 @@ void Lookup::unqualifiedImpl(const Scope& scope, std::string_view name, LookupLo
                 if (!selectedImport)
                     selectedImport = &imports[0];
 
-                const Symbol* finalSymbol =
-                    canonicalSymbol ? canonicalSymbol : selectedImport->imported;
+                const Symbol* finalSymbol = canonicalSymbol ? canonicalSymbol
+                                                            : selectedImport->imported;
 
                 if (symbol && sourceRange) {
                     // The existing symbol might be an import for the thing we just imported
--- a/tests/unittests/DriverTests.cpp
+++ b/tests/unittests/DriverTests.cpp
@@ -4,8 +4,8 @@
 #include "Test.h"
 #include <fmt/core.h>
 #include <fstream>
-#include <regex>
 #include <iostream>
+#include <regex>
 
 #include "slang/ast/symbols/CompilationUnitSymbols.h"
 #include "slang/ast/symbols/InstanceSymbols.h"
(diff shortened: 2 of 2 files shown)


tests/unittests/ast/LookupTests.cpp
ebrevdo, line 1106: i think the issue here might be that warnings don't show up in the diagnostics; i wasn't sure the perfect way to update this test.  LMK if you prefer a different approach here.

MikePopoloski:
Can you say which tool allows this behavior? Every tool I've tried fails with an error for this situation; VCS, Questa, Xcelium, and RivieraPro.

While I'm happy in general to add compatibility flags to work with existing real codebases, it seems like not a great idea to promote coding styles that don't work with literally any commercial simulator. Surely at some point you want your design to be simulated by one of the big three?

ebrevdo:
> Can you say which tool allows this behavior? Every tool I've tried fails with an error for this situation; VCS, Questa, Xcelium, and RivieraPro.
> 
> While I'm happy in general to add compatibility flags to work with existing real codebases, it seems like not a great idea to promote coding styles that don't work with literally any commercial simulator. Surely at some point you want your design to be simulated by one of the big three?

Thanks for the feedback!  This is a large legacy codebase and I've not dug deeply into the original synthesis flow.  Nevertheless, I think I'd rather try to fix behavior to something more reasonably accepted by the community.

As such I've modified the PR to handle wildcard import "diamonds": when the same symbol is wildcard-imported via multiple paths.  In this case, there's no ambiguity: it's the exact same symbol.  We just have to do a bit more work to check.

MikePopoloski:
The concept of a "diamond" import here does not seem well defined. You've kind of just made up the idea that if a parameter happens to resolve to the same constant value that it's the same symbol? That's not how the LRM specifies the language works. For example, the LRM says this:
> It shall be illegal if the wildcard import of more than one package within the same scope defines the same
potentially locally visible identifier and a search for a reference matches that identifier.

I also again can't find a tool that supports this behavior. I am more likely to accept a compatibility flag more like your original proposal; this new one seems very ad-hoc and opaque and hard to explain to users. I'd still like to know which tool we're trying to emulate with this non-standard behavior though.

ebrevdo:
> The concept of a "diamond" import here does not seem well defined. You've kind of just made up the idea that if a parameter happens to resolve to the same constant value that it's the same symbol? That's not how the LRM specifies the language works. For example, the LRM says this:
> 
> > It shall be illegal if the wildcard import of more than one package within the same scope defines the same
> > potentially locally visible identifier and a search for a reference matches that identifier.
> 
> I also again can't find a tool that supports this behavior. I am more likely to accept a compatibility flag more like your original proposal; this new one seems very ad-hoc and opaque and hard to explain to users. I'd still like to know which tool we're trying to emulate with this non-standard behavior though.

From a quick investigation, seems likely to be Synopsys; however in that case I'd also expect to have seen failures there.  Regardless, after this investigation I agree with your assessment and will just do some targeted rewriting of the legacy codebase.  I'll withdraw this PR.

The other PR I sent over addresses an issue that would be trickier for me to manually fix, and seems more likely to be a useful feature for slang.  Hope we can focus on that one.

Issue #1409, opened 2025-06-15
Generate and ship Python sources to enable autocomplete


gussmith23:
**Is your feature request related to the C++ slang project or the pyslang Python bindings?**
Pyslang Python bindings

**Is your feature request related to a problem? Please describe.**

I'd like autocomplete to work with Pyslang.

**Describe the solution you'd like**

I may be missing something, but it seems like no Python files are actually generated when pyslang is built. Is there a way to get Python autocomplete suggestions working with this kind of Python library?

I was eventually able to use `stubgen` (within `mypy`) to generate a `.pyi` for pyslang. Here's how I did it:

1. Cloned pyslang, cd'ed to the directory and did the standard `pip install .`
2. Ran `stubgen -m pyslang`
3. Copied the resulting `out/pyslang.pyi` to sit next to the `pyslang.*.so` file in my `site-packages` (in this case, inside the `.venv` folder in my project)

MikePopoloski:
Hmm, in theory the Python dist build does generate stubs and include them in the release: https://github.com/MikePopoloski/slang/blob/master/.github/workflows/python-dist.yml#L59

Is it not making it into the PyPI distribution somehow?

MikePopoloski:
Yeah, I just checked and it seems to be working for me in the Python REPL.

MikePopoloski:
Ok well I'm going to close this since there's been no followup.

stefanpie:
For some reason, the stubs don't appear in the latest pyslang package (corresponding to the 9.1 slang release) even though it appears in CI yaml files. We (@ilthraim and I) would like to reopen the issue and can provide some more debugging information if needed. We are currently installing the package though uv with `uv add pyslang` and we have double-checked the package is the 9.1 release.

In parallel, we are also attempting to maybe integrate this into the cmake build process or port the binding generation from using pybind11 to using [nanobind](https://nanobind.readthedocs.io/en/latest/) which supports annotation generation as well as doc generation.

MikePopoloski:
FWIW I'd be happy to see the bindings ported to nanobind. When I took a brief look a while back it wasn't clear whether it would be possible given that slang uses multiple inheritance for Symbols / Scope derived classes. The Python side doesn't need to know about or use the multiple inheritance, but it seemed like I couldn't even expose those classes with nanobind.

stefanpie:
[...]

PR #1561, opened 2025-11-03
Integrate pybind11 stub generation into CMake build

stefanpie:
This pull requests integrate generating Python stubs for the pyslang bindings into the CMake build process. Previously, the sub generation was done only in the CI which seemed like a brittle way to do it and prevents it from being included if building the package locally (ex. when running `uv build`). Also, on our end, the 9.1 release installed from PyPi did not have stubs. Hopefully, this pull request should also fix that issue.

Not sure what is the best way to test this. After my changes, I was able to run `uv build`, unpack the generated wheel, and manually check the stubs were included. I also installed the wheel into a fresh venv and the stubs are also in there was well. Not sure how to best test the CI integration, but that seems best left up to the maintainers.

### DIFF
--- a/bindings/CMakeLists.txt
+++ b/bindings/CMakeLists.txt
@@ -73,6 +73,4 @@ install(
   DESTINATION .)
 
 # Generate Python stubs as a post-install step
-install(
-  SCRIPT ${CMAKE_CURRENT_SOURCE_DIR}/generate_stubs.cmake
-  COMPONENT pylib)
+install(SCRIPT ${CMAKE_CURRENT_SOURCE_DIR}/generate_stubs.cmake COMPONENT pylib)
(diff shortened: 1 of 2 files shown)


stefanpie:
I created this pull request as an attempt to address the issue I describe in #1409

Since most of these changes are with the build and release process, it would be greatly appreciated for others to be able to test these changes independently of me to be sure they are correct and don't break any current build or release flows.

MikePopoloski:
Thanks for the PR. I'm in favor of the direction, though it looks like the stub generation is failing for reasons unclear to me. The original implementation was a contribution so I don't have a lot of knowledge about the process to begin with.

stefanpie:
Thanks for the feedback!
I think the issue is that the bindings are generated in a pyslang subfolder inside the pyslang wheel instead of the root. This means that this will appear to work `from pyslang import pyslang` and this won't work `import pyslang`. Should be easy to fix and test.

MikePopoloski:
Are you planning to fix up this PR? If not we should close it out.

MikePopoloski:
Closing since there's been no followup.

stefanpie:
Sorry for the delay, I shifted to some other work for a while. If we are able to fix up this issue in the future for a PR, we can ask to reopen or open another PR. For now, we are just working off our fork until we have more time for a fix to merge back into here.

Issue #1568, opened 2025-11-10
Verilog compatibility


povik:
**Is your feature request related to the C++ slang project or the pyslang Python bindings?**
C++ slang

**Is your feature request related to a problem? Please describe.**
yosys-slang cannot always read Verilog sources as they use SystemVerilog keywords to label nets or for other identifiers.

**Describe the solution you'd like**
Add Verilog compatibility enabled via `--std=`. I think I can make this option disable all SV keywords. This wouldn't be a Verilog linter as it would still allow all SV features which don't involve new keywords. Would this be fine?

**Describe alternatives you've considered**

MikePopoloski:
Yes, that's fine with me. Preferably using `-std=` with an older version would actually change the semantics to match that particular version, but that's quite an undertaking to look through the old specs and see what has changed. We can address issues on an ad-hoc basis as people run into them and report them.

slang already supports `` `begin_keywords `` and `` `end_keywords `` so just changing the keyword handling will be very easy.

likeamahoney:
> Yes, that's fine with me. Preferably using `-std=` with an older version would actually change the semantics to match that particular version, but that's quite an undertaking to look through the old specs and see what has changed. We can address issues on an ad-hoc basis as people run into them and report them.
> 
> slang already supports `` `begin_keywords `` and `` `end_keywords `` so just changing the keyword handling will be very easy.

Hi! I suggest such kind a solution.

Can we add an `--verilog` option for `slang` which can be specified before passing `verilog` sources in CLI. In this case `slang` can cover `SourceBuffer`s of this sources with `begin..end_keywords` macros implicitly (or setting for this files preprocessor keyword version directly). It would be helpful for developers.

povik:
> slang already supports `begin_keywords and `end_keywords so just changing the keyword handling will be very easy.

Indeed this has turned out to require little work. I skimmed the standards and grepped the SV one for mentions of "Verilog" to see if it points out any critical differences. In the hour or two I have invested into it I haven't found any which seemed relevant.

MikePopoloski:
> > Yes, that's fine with me. Preferably using `-std=` with an older version would actually change the semantics to match that particular version, but that's quite an undertaking to look through the old specs and see what has changed. We can address issues on an ad-hoc basis as people run into them and report them.
> > slang already supports `` `begin_keywords `` and `` `end_keywords `` so just changing the keyword handling will be very easy.
> 
> Hi! I suggest such kind a solution.
> 
[...]

PR #1576, opened 2025-11-18
Add --map-keyword-version option

likeamahoney:
Here I have added an `--emit-verilog-by-ext` option which forces `slang` preprocessor to emit `verilog IEEE 1364-2005` keywords when it meets files with `.v` and `.vh` extensions according to this discussion https://github.com/MikePopoloski/slang/issues/1568#issuecomment-3533567818

May be it needs to specify extensions as option arguments instead of just boolean option but idk most of cases which I saw contained only this extensions in the designs and filelists with old verilog IPs blocks usage.

### DIFF
--- a/source/parsing/Preprocessor.cpp
+++ b/source/parsing/Preprocessor.cpp
@@ -1071,7 +1071,8 @@ Trivia Preprocessor::handleEndKeywordsDirective(Token directive) {
                 break;
             }
         }
-        keywordVersionStack.erase(stackIt.base());
+
+        keywordVersionStack.erase(stackIt.base() - 1);
     }
 
     return createSimpleDirective(directive);


likeamahoney:
Removing draft status after fixing found issues with headers.

likeamahoney:
@povik Hello! Would it be useful for your `yosys-slang`? 

Please, Can you provide some suggestions if you will not satisfied with such changes?

povik:
Hello @likeamahoney, so far I only stumbled on cases solvable with a global flag such as the one implemented in https://github.com/MikePopoloski/slang/pull/1574. Your per file option might be useful to someone but I haven't had the need yet.

MikePopoloski:
I don't love how specific this is to certain extensions and keyword versions. I think if we want to have something like this then it should be more general; probably an option that allows mapping a specific extension or a set of files (via glob perhaps) to a keyword version to use, and not limit it to just that one Verilog version.

likeamahoney:
> I don't love how specific this is to certain extensions and keyword versions. I think if we want to have something like this then it should be more general; probably an option that allows mapping a specific extension or a set of files (via glob perhaps) to a keyword version to use, and not limit it to just that one Verilog version.

I re-implemented the option by made it more general. For now the option us named `--map-keyword-version`. It forces slang to interpret files which satisfy the passed patterns with specified keywords version (avaliable keywords is standart `slang` keyword version set - "1364-1995, 1364-2001-noconfig, 1364-2001, 1364-2005, 1800-2005, 1800-2009, 1800-2012, 1800-2017, 1800-2023").

Example of usage:
`--map-keyword-version=1364-2005:*.v,*.vh` or `--map-keyword-version=1364-2005:/path/to/verilog.v`.

It would be useful for multi language designs with old  IP cores written on Verilog.

likeamahoney:
> Example of usage: `--map-keyword-version=1364-2005:*.v,*.vh` or `--map-keyword-version=1364-2005:/path/to/verilog.v`.

I Replaced `:` with `+` due to windows file system path naming convention

likeamahoney:
@MikePopoloski Hello, is there anything else that needs to be improved in the PR?

MikePopoloski:
I took a quick look and yes there are a number of things that need to be fixed or improved but I haven't had time to do a full review yet.

include/slang/util/CommandLine.h
```diff
@@ -307,6 +311,18 @@ class SLANG_EXPORT CommandLine {
...
+    ///
+    /// @param value a string containing a single plus-separated
+    ///              "keyword_version+file_pattern,[...]" pair, where the "keyword_version"
+    ///              can pass only such types of values:
+    ///              "1364-1995, 1364-2001-noconfig, 1364-2001, 1364-2005, 1800-2005,
+    ///              1800-2009, 1800-2012, 1800-2017, 1800-2023". After which starts
+    ///              the comma-separated list of file patterns.
+    /// @returns a string containing an error message if the @a value is malformed.
+    std::map<std::string, std::pair<parsing::KeywordVersion, std::optional<SmallVector<fs::path>>>>
+    parseMapKeywordVersion(std::string_view value, std::string& error);
```
MikePopoloski: The CommandLine class should remain generic and not have these project-specific parsing functions. Instead this should move to the Driver class.

source/parsing/Preprocessor.cpp
```diff
@@ -98,6 +99,26 @@ void Preprocessor::pushSource(std::string_view source, std::string_view name) {
 void Preprocessor::pushSource(SourceBuffer buffer) {
     SLANG_ASSERT(buffer.id);
 
+    if (!options.keywordMapping.empty()) {
```
MikePopoloski, line 111: This is very confusing; why do we sometimes use the keywordMapping and sometimes instead rely on the buffer's forcedKeywordVersion? It seems like you can remove all of the SourceLoader and forcedKeywordVersion stuff and just rely on the PreprocessorOptions to map buffers.

source/parsing/Preprocessor.cpp
```diff
@@ -572,6 +597,8 @@ Trivia Preprocessor::handleIncludeDirective(Token directive) {
             addDiag(diag::ExceededMaxIncludeDepth, fileName.range());
         }
         else if (includeOnceHeaders.find(buffer->data.data()) == includeOnceHeaders.end()) {
+            driver::SourceLoader::findBufferInKeywordMapping(buffer.value(), options.keywordMapping,
```
MikePopoloski: The Driver / SourceLoader API is at a higher level of abstraction than the Preprocessor, so there should not be a backward dependency like this.

source/driver/SourceLoader.cpp
```diff
@@ -510,6 +516,37 @@ void SourceLoader::createLibrary(const LibraryDeclarationSyntax& syntax, const f
...
+    std::map<std::string, std::pair<parsing::KeywordVersion, std::optional<SmallVector<fs::path>>>>&
+        keywordMapping,
+    const SourceManager& sourceManager) {
+    if (buf.forcedKeywordVersion.has_value())
+        return;
+
+    auto& pathToCheck = sourceManager.getFullPath(buf.id);
+    for (auto& [pattern, keywordWithSearchCache] : keywordMapping) {
+        auto& searchCache = keywordWithSearchCache.second;
+        if (!searchCache.has_value()) {
```
MikePopoloski: You don't need to (and shouldn't) actually glob the file system. `svGlobMatches` will check whether the path of the file matches the glob, so you can just compare each mapping to the pathToCheck and skip all this caching stuff.

include/slang/util/CommandLine.h
likeamahoney: done

source/parsing/Preprocessor.cpp
likeamahoney: removed

source/driver/SourceLoader.cpp
likeamahoney: Thank you for review! I removed all the `SourceLoader` stuff and filesystem globbing.

In the first implementation, I wanted to give the user the possibility to map both full and relative paths (which is what filesysten globbing is for - for example `slang --map-keyword-version 1364-2005+*.v` - which have to map all files with `.v` suffix in translation unit) ). Currently, mapping is only possible using the full path. For example:

`slang --map-keyword-version 1364-2005+${PWD}/tests/unittests/data/*.v tests/unittests/data/systemverilog.sv`

I hope this option will be just as convenient for users.

source/parsing/Preprocessor.cpp
likeamahoney: Fixed!

Now I am relying only on `options.keywordMapping` and then searching for the buffer keyword version if it present and else restoring keyword version from the stack

source/driver/SourceLoader.cpp
MikePopoloski: This should still work, they just have to write it like `.../*.v`

likeamahoney: I have checked this - yes it is works!

source/parsing/Preprocessor.cpp
```diff
@@ -9,6 +9,7 @@
 
 #include "slang/diagnostics/LexerDiags.h"
 #include "slang/diagnostics/PreprocessorDiags.h"
+#include "slang/driver/SourceLoader.h"
```
MikePopoloski: Remove unused include.

docs/command-line-ref.dox
```diff
@@ -86,6 +86,16 @@ Note that multithreading *parsing* cannot be supported when running
 with `--single-unit`, though other parts of the compilation process
 may still take advantage of threading where possible.
 
+`--map-keyword-version <keyword-version>+<file-pattern>[,...]`
+
+Forces slang to interpret files which satisfy the passed patterns with specified
+keywords version. For example '--map-keyword-version=1364-2005+*.v,*.vh'
```
MikePopoloski: Update examples with correct glob syntax.

include/slang/driver/Driver.h
```diff
@@ -134,6 +135,9 @@ class SLANG_EXPORT Driver {
         /// A set of options controlling translate-off comment directives.
         std::vector<std::string> translateOffOptions;
 
+        /// A map of file patterns to its keyword versions
+        std::map<std::string, parsing::KeywordVersion> keywordMapping;
```
MikePopoloski: I think this actually needs to be a std::vector<std::pair<>> everywhere. You want to maintain the order that the mappings are provided so that there is determinism in which ones get applied if more than one pattern could match a file.

source/driver/Driver.cpp
```diff
@@ -454,6 +472,57 @@ void Driver::addStandardArgs() {
...
+                fileNamePatterns.substr(startPos, nextComma - startPos));
+
+            startPos = nextComma + 1;
+        }
+        else {
+            extractedFilePattern = std::string(fileNamePatterns.substr(startPos, endPos));
+            startPos = endPos;
+        }
+
+        if (keywordMapping.contains(extractedFilePattern) &&
```
MikePopoloski: This error check doesn't make much sense to me... it's saying that it matched another keyword version but it actually checks if it matches the *same* keyword version? I think you can just remove the check altogether.

source/driver/Driver.cpp
```diff
@@ -454,6 +472,57 @@ void Driver::addStandardArgs() {
     return !anyFailedLoads;
 }
 
+std::map<std::string, parsing::KeywordVersion> Driver::parseMapKeywordVersion(
+    std::string_view value, std::string& error) {
+    const size_t firstPlusIndex = value.find_first_of('+');
+    const size_t lastPlusIndex = value.find_last_of('+');
+    std::map<std::string, parsing::KeywordVersion> keywordMapping;
+
+    if (firstPlusIndex == std::string_view::npos || firstPlusIndex != lastPlusIndex) {
```
MikePopoloski: I don't think you need to check the last '+' position. If we find the first one that seems sufficient. I suppose it's possible that paths contain a '+' character and then it will just work in that case too.

source/parsing/Preprocessor.cpp
```diff
@@ -95,9 +96,39 @@ void Preprocessor::pushSource(std::string_view source, std::string_view name) {
         sourceManager.addLineDirective(SourceLocation(buffer.id, 0), 2, name, 0);
 }
 
+std::optional<parsing::KeywordVersion> Preprocessor::findBufferInKeywordMapping(SourceBuffer& buf) {
```
MikePopoloski: This function should just take a `BufferID` since that's all it needs to do its work.

source/parsing/Preprocessor.cpp
```diff
@@ -117,6 +148,10 @@ bool Preprocessor::popSource() {
     if (includeDepth)
         includeDepth--;
 
+    if (!keywordVersionStack.empty() &&
```
MikePopoloski, line 151: I think you also need to update `handleEndKeywordsDirective` so that such a directive can't pop these forced entries.

source/parsing/Preprocessor.cpp
likeamahoney: done

docs/command-line-ref.dox
likeamahoney: added

include/slang/driver/Driver.h
likeamahoney: thank you for suggestion, fixed!

source/driver/Driver.cpp
likeamahoney: I removed this check

source/parsing/Preprocessor.cpp
likeamahoney: done

source/parsing/Preprocessor.cpp
likeamahoney: updated

source/driver/Driver.cpp
likeamahoney: removed it and also related test was removed

PR #1598, opened 2025-12-09
Change path string conversion to generic_string on windows

evanwporter:
See #1596

### DIFF
--- a/source/text/SourceManager.cpp
+++ b/source/text/SourceManager.cpp
@@ -392,6 +392,10 @@ SourceBuffer SourceManager::assignBuffer(std::string_view bufferPath, SmallVecto
     // first see if we have this file cached
     fs::path path(bufferPath);
 #ifdef _WIN32
+    // this is necessary on windows because fs::path::string returns a backslashed `\\` path.
+    // however weakly_canonical returns a forward slashed `/` path. If we don't do this then
+    // it conflicts with the other uses of weakly_canonical (namely in `isCached` and `openCache`)
+    // and we end up looking for a different key in `lookupCache`.
     auto pathStr = path.generic_string();
 #else
     auto pathStr = getU8Str(path);


evanwporter:
@MikePopoloski 

My understanding of `fs::path::generic_string` is that it returns a generic string which uses forward slashes. `weakly_caninical` also returns a generic string. `fs::path::string` returns the native string, which uses backward slashes on windows. 

See: https://stackoverflow.com/a/26178784

There may still be an issue with `assignBuffer` not using `weakly_canonical`, but that's outside the scope of the problem I'm trying to solve.

include/slang/util/String.h
```diff
@@ -80,20 +80,9 @@ inline char charToLower(char c) {
...
 inline std::string getU8Str(const std::filesystem::path& path) {
-    return path.string();
-}
-
-#else
-
-/// Gets a string representation of the given path, in UTF-8 encoding.
-inline const std::string& getU8Str(const std::filesystem::path& path) {
-    return path.native();
+    return path.generic_string();
```
MikePopoloski: I think you should keep the ifdef -- for non-Windows builds the native() method already returns the right thing and returns by reference, so we save an allocation + copy of the string. Changing the WIN32 side of the branch to use generic_string instead seems fine though.

evanwporter: >the native() method already returns the right thing and returns by reference, so we save an allocation + copy of the string

I didn't know that, that's cool!

evanwporter:
yikes, it failed.

MikePopoloski:
What I meant was that you should still change getU8Str, just for the Win32 branch. The method has to be consistent everywhere so you can't change just that one use in SourceManager.

evanwporter:
@MikePopoloski 

I did that in cd7dec69da08b012399bd623889e22df6150509e. But then I realized the tests failed for windows from the first commit, and if they ran for the second commit (cd7dec69da08b012399bd623889e22df6150509e) they would've failed as well.

So, my solution is to modify source manager directly.  I'm not really familiar with the codebase though, so I don't really know what the available alternatives are though.

These tests should pass though.

evanwporter:
I'm going to look into this a little more

Issue #1611, opened 2025-12-18
[feature] Xilinx Vivado parameter use before declare compatibility


likeamahoney:
Hi, all! I have found some designs that used parameters before its declarations. 

For example (from https://github.com/badgeteam/mch2022-firmware-ice40/blob/ce6473addcf6066cada7d80d8ba352f52173d01d/projects/Snake/rtl/common-verilog/stack2.v#L12):

```verilog
`default_nettype none

module stack2(
  input wire clk,
  output wire [WIDTH-1:0] rd, // use before declaration
  input wire we,
  input wire [1:0] delta,
  input wire [WIDTH-1:0] wd // use before declaration
);

  parameter DEPTH = 16;
  parameter WIDTH = 16;
  localparam BITS = (WIDTH * DEPTH) - 1;

  ...
endmodule
```

This is only accepted (it can be synthesed) by Xilinx Vivado FPGA synthesis tool (with warning but no error)  and no any other commercial tool.

And `yosys` open source logic synthesis tool also accepts this without warning.

@MikePopoloski  Can I add support for this in Slang? For example, as an option or with the `-compat vivado` cli option? There are plenty of such FPGA designs in open source. Or is this too special a case to be considered? I guess it would be useful to have such compatibility.

MikePopoloski:
Yes, we can add a compat flag. This might be tricky though, since the order of evaluation of parameters is delicate and I could imagine this introducing new ways to create cyclic dependencies where previously they could not exist.

likeamahoney:
> Yes, we can add a compat flag. 

Thank you for approving!

> This might be tricky though, since the order of evaluation of parameters is delicate and I could imagine this introducing new ways to create cyclic dependencies where previously they could not exist.

If you have time and any relevant examples come to mind, I would appreciate it if you could share some, preferably involving interesting cyclic dependency cases. I currently use a trivial implementation, and it works correctly because the slang default evaluation and lookup mechanisms are able to detect cyclic dependencies in symbol self-definitions on their own. For example such cases are working fine (slang correctly errored on these cyclic dependencies after I had unset declaration-before checking for params in symbols lookup -- without endless loop or segfault terminations):

```verilog
parameter P = P; // error
```

```verilog
parameter P = P1 // error;
parameter P1 = P;
```

```verilog
module top #(W = WIDTH)(
);

  parameter WIDTH = BITS * W; // error
  localparam BITS = (WIDTH * DEPTH) - 1; // error

endmodule
```

[...]

PR #1614, opened 2025-12-22
[compat][allow-use-before-declare] Add support for parameters and emit a warning instead of error or nothing

likeamahoney:
Hi, @MikePopoloski 

Currently, this PR may not fully address the issue (https://github.com/MikePopoloski/slang/issues/1611), and therefore may not be ready for closure. There are several aspects that are still unclear to me, and I would appreciate feedback based on your experience.

Current state:

1. I observed that `Xilinx Vivado` emits a warning for use-before-declaration of parameters and variables, rather than an error (or no diagnostic at all), whereas slang currently emits nothing unless the `--allow-use-before-declare` option is enabled.
2. Based on this observation, I chose to make slang emit a warning when `--allow-use-before-declare` is enabled, instead of remaining silent. At this point, this is not implemented as a dedicated warning category (e.g. `-W…`), since it is still unclear whether emitting a warning is the right behavior, or whether emitting nothing would be preferable - You can help with clarification of this for me.
3.  I disabled the `--allow-use-before-declare` compiler flag by default for `VCS` compatibility. In my testing with variables and nets, `VCS` reports errors for such cases. This change can be reverted if you prefer a different default.
4. I added an explicit check for self-dependent parameters when the option is enabled and tested it on several examples involving self-referential parameter definitions. I believe having a dedicated check for this case improves usability when --allow-use-before-declare is in effect.
5. I did not add a separate tool-compatibility mode (e.g. for `Vivado` or `Yosys`), as I assumed the behavior could be sufficiently controlled via the `--allow-use-before-declare` option. However, I can add such compatibility options if you think they are necessary.

### DIFF
--- a/source/ast/Lookup.cpp
+++ b/source/ast/Lookup.cpp
@@ -61,6 +61,9 @@ bool LookupResult::hasError() const {
         return true;
 
     for (auto& diag : diagnostics) {
+        if (diag.code == diag::UsedBeforeDeclared)
+            return true;
+
         if (diag.isError())
             return true;
     }
@@ -1767,6 +1770,7 @@ void Lookup::unqualifiedImpl(const Scope& scope, std::string_view name, LookupLo
     // Try a simple name lookup to see if we find anything.
     auto& nameMap = scope.getNameMap();
     const Symbol* symbol = nullptr;
+    bool reportDeclaredAfter = false;
     if (auto it = nameMap.find(name); it != nameMap.end()) {
         // If the lookup is for a local name, check that we can access the symbol (it must be
         // declared before use). Callables and block names can be referenced anywhere in the
@@ -1828,11 +1832,14 @@ void Lookup::unqualifiedImpl(const Scope& scope, std::string_view name, LookupLo
                     break;
                 case SymbolKind::Parameter:
                 case SymbolKind::Specparam:
-                case SymbolKind::EnumValue:
-                    // Constants can never be looked up before their declaration,
-                    // to avoid problems with recursive constant evaluation.
-                    flags &= ~LookupFlags::AllowDeclaredAfter;
+                case SymbolKind::Port:
+                case SymbolKind::Net:
+                case SymbolKind::Variable: {
+                    // It is needed to defer reporting of use before declared
+                    // because of wildcard import and class scope later resolving.
+                    reportDeclaredAfter = true;
                     break;
+                }
                 default:
                     break;
             }
@@ -1917,6 +1924,14 @@ void Lookup::unqualifiedImpl(const Scope& scope, std::string_view name, LookupLo
                     else {
                         result.found = symbol;
                     }
+                    break;
+                case SymbolKind::Parameter:
+                case SymbolKind::Specparam:
+                case SymbolKind::EnumValue:
+                    // If location doesn't match for constant then try to search
+                    // in outer scopes to find a better match.
+                    result.found = symbol;
+                    done = locationGood;
                     break;
                 default:
                     result.found = symbol;
@@ -2076,8 +2091,28 @@ void Lookup::unqualifiedImpl(const Scope& scope, std::string_view name, LookupLo
         return;
     }
 
-    return unqualifiedImpl(*location.getScope(), name, location, sourceRange, flags,
-                           outOfBlockIndex, result, originalScope, originalSyntax);
+    unqualifiedImpl(*location.getScope(), name, location, sourceRange, flags, outOfBlockIndex,
+                    result, originalScope, originalSyntax);
+    if (reportDeclaredAfter && !result.found) {
+        auto isViable = [&originalScope](const Symbol* s) {
+            return !(VariableSymbol::isKind(s->kind) &&
+                     s->as<VariableSymbol>().flags.has(VariableFlags::CompilerGenerated)) &&
+                   isVisibleFrom(*s, originalScope);
+        };
+        if (isViable(symbol) && sourceRange && !flags.has(LookupFlags::AllowDeclaredAfter)) {
+            auto& diag = result.addDiag(originalScope, diag::UsedBeforeDeclared, *sourceRange);
+            diag << name;
+            diag.addNote(diag::NoteDeclarationHere, symbol->location);
+
+            result.found = symbol;
+            auto declaredType = result.found->getDeclaredType();
+            if (declaredType && declaredType->isEvaluating()) {
+                reportRecursiveError(*result.found);
+                return;
+            }
+        }
+    }
+    return;
 }
 
 void Lookup::qualified(const ScopedNameSyntax& syntax, const ASTContext& context,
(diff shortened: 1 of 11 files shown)


likeamahoney:
I guess `wasi-sdk-pthread` pipeline was failed by reasons not related to my changes

MikePopoloski:
I don't think this is the right approach. If someone is using this flag it's because they have a codebase that requires it, so reporting an extra warning is not helpful -- they already know they're doing this and want it to work correctly.

We also can't change the VCS compatibility mode; this will break existing users. The latest version of VCS may do the right thing now but older versions did not and there is a lot of code out there that relies on this to work.

likeamahoney:
> I don't think this is the right approach. If someone is using this flag it's because they have a codebase that requires it, so reporting an extra warning is not helpful -- they already know they're doing this and want it to work correctly.
> 
> We also can't change the VCS compatibility mode; this will break existing users. The latest version of VCS may do the right thing now but older versions did not and there is a lot of code out there that relies on this to work.

Thank you for review! I will fix this!

likeamahoney:
> I don't think this is the right approach. If someone is using this flag it's because they have a codebase that requires it, so reporting an extra warning is not helpful -- they already know they're doing this and want it to work correctly.
> 
> We also can't change the VCS compatibility mode; this will break existing users. The latest version of VCS may do the right thing now but older versions did not and there is a lot of code out there that relies on this to work.

Hi!

I reimplemented my solution according to your feedback and reverted any changes that affected the existing logic.

I added two compatibility options: `Vivado` and `Yosys`. At the moment, these are just aliases, but I can keep only one if having two options is not necessary right now.

I tested the `Vivado` behavior: it allows the use of parameters, specparams, and variables before their declaration, issuing a warning rather than an error. I aligned my solution with this behavior.

I also noticed that `slang` reports an error for the following code:

```verilog
module paramTest1 ();
      specparam tAC = tBC;
      specparam tBC = 10;
endmodule
```

but does not report an error for this variant:

```verilog
module paramTest1 ();
      specify
        specparam tAC = tBC;
        specparam tBC = 10;
     endspecify
endmodule
```

I do not see a significant semantic difference between these two examples, so I could address this as part of this PR or handle it in a separate one.

source/ast/Lookup.cpp
```diff
@@ -1827,12 +1830,27 @@ void Lookup::unqualifiedImpl(const Scope& scope, std::string_view name, LookupLo
                     locationGood = true;
                     break;
                 case SymbolKind::Parameter:
-                case SymbolKind::Specparam:
+                case SymbolKind::Specparam: {
```
MikePopoloski: Probably the existing behavior for params can just be removed; I think I added that when other mechanisms for resolving recursive parameter definitions didn't exist. You can just merge all of the params + port / net / variable cases and handle them uniformly.

MikePopoloski: I don't think you need a separate lookup flag or compilation flag for this. It's very convoluted to be setting flags all over the place just to reuse a small part of the reportUndeclared function. Instead if you hit this case here you can just issue the warning directly right here and move on.

source/driver/Driver.cpp
```diff
@@ -669,6 +676,10 @@ bool Driver::processOptions() {
         diagEngine.setSeverity(diag::UnknownSystemName, DiagnosticSeverity::Error);
     }
 
+    if (options.compat == CompatMode::Vivado || options.compat == CompatMode::Yosys ||
+        options.compat == CompatMode::All)
+        diagEngine.setSeverity(diag::UsedBeforeDeclared, DiagnosticSeverity::Warning);
```
MikePopoloski: Going from Error -> Warning this way isn't correct; you will need to make this a proper warning in the diagnostics.txt and add documentation for it to warning_docs.txt, and then default it to an Error here when the compat mode is *not* set.

MikePopoloski (commented):
I agree that specparam example you showed seems like a bug. Probably worth handling in a separate diff.

source/ast/Lookup.cpp
likeamahoney: Hi! Sorry for the delay with the fixes. I have removed the previously added flags and eliminated the existing behavior for params. I've also unified the logic for params and variables.

> Instead if you hit this case here you can just issue the warning directly right here and move on.

I can't do the warning directly right mentioned point due to possible unresolved wildcard imports, which need to be handled later. Instead, I've made the reporting mechanism deferred until the resolving.

source/driver/Driver.cpp
likeamahoney: Done

likeamahoney:
> I agree that specparam example you showed seems like a bug. Probably worth handling in a separate diff.

Sorry, I realized that the SystemVerilog LRM addresses this in section 6.20.5:

> A specify parameter declared outside a specify block shall be declared before it is referenced.

So I think `slang` is actually behaving correctly by handling these cases differently.

source/ast/Lookup.cpp
MikePopoloski: Ah yeah, that's right. This still looks wrong to me but I'm not sure what it should look like exactly. I'll have to poke around in the code a bit to figure it out on my end.

MikePopoloski:
Ok, I pushed 56d0b585de8a73da3df57290a9dd5020aa1c6892 which has the basic fix to allow using parameters before they're declared, without the other potential regressions in behavior introduced by this PR.

It does not do:
- Add new compat modes. This could be done with a new PR based on this one, I'm still open to that.
- Issue a warning for cases where symbols are used before they are declared. This can't actually be done with the current flags available during lookup, because there are other cases besides the command line flag that legitimately allow using items before they're declared and we don't want that to suddenly start throwing warnings. I'm still not convinced we need a warning for this case though.

likeamahoney:
> Ok, I pushed [56d0b58](https://github.com/MikePopoloski/slang/commit/56d0b585de8a73da3df57290a9dd5020aa1c6892) which has the basic fix to allow using parameters before they're declared, without the other potential regressions in behavior introduced by this PR.
> 
> It does not do:
> 
> * Add new compat modes. This could be done with a new PR based on this one, I'm still open to that.
> * Issue a warning for cases where symbols are used before they are declared. This can't actually be done with the current flags available during lookup, because there are other cases besides the command line flag that legitimately allow using items before they're declared and we don't want that to suddenly start throwing warnings. I'm still not convinced we need a warning for this case though.

Thank you for the fix. What would you recommend I include in this PR at this stage?
For now, I can keep the logic that adds two new compatibility modes and increase the warning severity when it is not explicitly configured, as you mentioned above.

MikePopoloski:
I think the warning severity stuff is no longer relevant, since it won't apply in the cases where you want it to (the lookup will succeed and the warning will never be issued). If we really want to have warnings for these cases it'll have to be rethought, but again I'm not sure how useful they will be.

The new compat modes seem fine, though I'm not sure what they'll be setting exactly. It might be best to wait until there are more things to set for them?

MikePopoloski:
Going to close this for now since the original issue is resolved and it's unclear how useful the ancillary parts are.

Issue #1617, opened 2025-12-27
Adding "isSigned" to ScalarTypes in the AST JSON.


dinoruic:
**Is your feature request related to the C++ slang project or the pyslang Python bindings?**
C++ slang

**Is your feature request related to a problem? Please describe.**
When using the AST JSON with detailed types for analysis, I cannot get full type information of module arguments.

For example, using the following `test.v`:
```
module abc(
   input logic signed [4:0] a,
   output logic signed [4:0] b
);
   assign b = a;
endmodule : abc
```
I run the following command to obtain the AST will detailed types:
```
slang --ast-json-detailed-types -ast-json - test.v
```
which will show me the following for Port a:
```
            {
              "name": "a",
              "kind": "Port",
              "addr": 6248979568952,
              "type": {
                "name": "",
                "kind": "PackedArrayType",
                "addr": 6248979569744,
                "elementType": {
                  "name": "logic",
                  "kind": "ScalarType",
                  "addr": 94340151431744,
                },
                "range": "[4:0]"
              },
              "direction": "In",
              "internalSymbol": "6248979569080 a"
            },
```
There is no way to figure out that the packed array is storing signed values.

On the other hand, I can find that information when using non-detailed types. Running:
```
slang -ast-json - test.v
```
I get:
```
            {
              "name": "a",
              "kind": "Port",
              "addr": 4478916172088,
              "type": "logic signed[4:0]",
              "direction": "In",
              "internalSymbol": "4478916172216 a"
            },
```
which shows that the type is indeed signed.

**Describe the solution you'd like**
Adding the `"isSigned": <bool>` field to the `ScalarType` kind.

**Describe alternatives you've considered**
Parsing both the non-detailed for the signed keyword and the detailed type ast for other data. It would be much nicer if the detailed types would also output the signed keyword.

dinoruic:
Hi, I hope you're doing well!

Created a PR with a proposed solution #1618 but I'm not super familiar with SystemVerilog syntax or with the inner workings of slang, so please give feedback on whether any of this makes sense or if there is a better way.

Thanks!

dinoruic:
Thanks a lot!

PR #1618, opened 2025-12-27
Adding isSigned output field to ScalarTypes in the JSON AST.

dinoruic:
This is a suggestion for adding the `"isSigned" : <bool>` field to the AST json output of ScalarTypes.

More info in #1617 

I'm not really sure if this is the right way to do it as I'm not super familiar with the SystemVerilog syntax and how slang handles things. Please let me know any feedback.


MikePopoloski:
This seems reasonable, but the tests have to be updated to account for the change.

dinoruic:
Thank you for the quick reply -- I'll take care of it once I'm back at my laptop. Happy holidays!

PR #1621, opened 2025-12-31
Feature: Full SyntaxFactory bindings for CST Manipulation

### LINKED ISSUE #1620: pyslang bindings for full SyntaxFactory for CST manipulation
1. **Is your feature request related to the C++ slang project or the pyslang Python bindings?**:
    pyslang Python bindings

2. **Is your feature request related to a problem? Please describe.**: 
    Not necessarily a problem, but just a need to manipulate SV CSTs easily through Python. I understand pyslang's usecase is currently scoped to "load SystemVerilog source files, inspect the syntax tree, and evaluate expressions." according to [the docs](https://github.com/MikePopoloski/slang/blob/master/pyslang/docs/pages/index.rst), but I would really want to add nodes without having to write SV for it in strings and then pass them over

3. **Describe the solution you'd like**
    Working on a [fork right now](https://github.com/CheeksTheGeek/slang/tree/feat/pyslang-syntax-factory-full), seems like 98% of the work was already done on the C++ side too, and the API is really clean to work with pybind11, so the only thing was just registering its constructor via pybind11 and then adding some typecasters (look at the [patch](https://github.com/MikePopoloski/slang/compare/master...CheeksTheGeek:slang:feat/pyslang-syntax-factory-full.patch)). Will make a PR if the maintainers are open to it and the pyslang roadmap is flexible to taking inputs like these at the moment.

4. **Describe alternatives you've considered**
    The `pyslang/tests/test_syntax_rewriter.py` test shows exactly what I'm talking about, and I think it's a worthwhile feature to have the ability to create nodes on the fly without having to parse them from another tree.

5. **Additional context**
    The reason I can think of that this functionality is missing right now is either of 
    a. pyslang is meant to be limited to analysis and exploration and not be mirrored to full slang's cpp side abilities (in that case I would want to make a long proposal of the benefits of just adding python bindings and seeing the crazy projects and usecases people come up with. Slang has become an integral dependancy in so many FOSS projects and its intermingling with Python will bear great fruits!) 
    b. pyslang's hierarchy is still being worked on (from #1215), and will change in a big way, so adding this wasn't a high enough priority compared to that (in which case, I'll add an example of what the simple bridging-the-gap adding-the-bindings work on my work can open the doors to. Also, I'd want to work towards accelerating that too so pyslang's hierarchy can solidify with the next major slang release!)

CheeksTheGeek:
Resolves #1620 
### bindings
- factory methods are accessible via `SyntaxRewriter.factory` property

### helpers for SyntaxRewriter
- `makeToken(kind, text, trivia)` - create tokens with explicit text
- `makeToken(kind, trivia)` - for tokens with text inferred from kind (keywords/punctuation only)
- `makeId(text, trivia)` - for identifier tokens
- `makeComma()` - for comma separator tokens
- `makeTrivia(kind, text)` - for trivia with allocator-managed text
- `makeList(items)` - for `SyntaxList` from Python list
- `makeSeparatedList(items)` - for `SeparatedSyntaxList` from mixed nodes/tokens
- `makeTokenList(items)` - for `TokenList` from token list
- `clone(node)` / `deep_clone(node)` - for cloning syntax nodes

note for memory safety: token text and trivia will now be copied into the bump allocator to prevent dangling pointers when python strings are garbage collected

### DIFF
--- a/pyslang/tests/test_syntax_factory.py
+++ b/pyslang/tests/test_syntax_factory.py
@@ -2,35 +2,44 @@
 # SPDX-License-Identifier: MIT
 
 import pytest
+
 import pyslang
+
 
 class TestMakeToken:
     def test_makeToken_with_explicit_text(self):
         token_created = {}
+
         def handler(node, rewriter):
             if node.kind == pyslang.SyntaxKind.ModuleDeclaration:
                 token = rewriter.makeToken(pyslang.TokenKind.Semicolon, ";", [])
                 token_created["token"] = token
                 assert token.kind == pyslang.TokenKind.Semicolon
 
-        pyslang.rewrite(pyslang.SyntaxTree.fromText("module m; endmodule", "test.sv"), handler)
+        pyslang.rewrite(
+            pyslang.SyntaxTree.fromText("module m; endmodule", "test.sv"), handler
+        )
         assert "token" in token_created
 
     def test_makeToken_inferred_text(self):
         """Test creating a token where text is inferred from kind"""
         token_created = {}
+
         def handler(node, rewriter):
             if node.kind == pyslang.SyntaxKind.ModuleDeclaration:
                 token = rewriter.makeToken(pyslang.TokenKind.Semicolon)
                 token_created["token"] = token
                 assert token.kind == pyslang.TokenKind.Semicolon
 
-        pyslang.rewrite(pyslang.SyntaxTree.fromText("module m; endmodule", "test.sv"), handler)
+        pyslang.rewrite(
+            pyslang.SyntaxTree.fromText("module m; endmodule", "test.sv"), handler
+        )
         assert "token" in token_created
 
     def test_makeToken_requires_explicit_text_for_identifier(self):
         """Test that makeToken raises error for tokens needing explicit text."""
         error_raised = {}
+
         def handler(node, rewriter):
             if node.kind == pyslang.SyntaxKind.ModuleDeclaration:
                 try:
@@ -38,13 +47,16 @@ class TestMakeToken:
                 except Exception as e:
                     error_raised["error"] = str(e)
 
-        pyslang.rewrite(pyslang.SyntaxTree.fromText("module m; endmodule", "test.sv"), handler)
+        pyslang.rewrite(
+            pyslang.SyntaxTree.fromText("module m; endmodule", "test.sv"), handler
+        )
         assert "error" in error_raised
         assert "explicit text" in error_raised["error"]
 
     def test_makeId(self):
         """Test creating an identifier token."""
         token_created = {}
+
         def handler(node, rewriter):
             if node.kind == pyslang.SyntaxKind.ModuleDeclaration:
                 token = rewriter.makeId("my_identifier")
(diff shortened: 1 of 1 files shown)


scripts/syntax_gen.py
```diff
@@ -48,14 +54,24 @@ def main():
     parser = argparse.ArgumentParser(description="Diagnostic source generator")
     parser.add_argument("--dir", default=os.getcwd(), help="Output directory")
     parser.add_argument("--python-bindings", action="store_true")
+    parser.add_argument(
```
MikePopoloski: I don't think there will be a reason to generate bindings without including the factory stuff right? So we don't need a separate argument for it.

bindings/python/SyntaxBindings.cpp
```diff
@@ -463,7 +547,50 @@ void registerSyntax(py::module_& m) {
...
         .def("insert_at_back", &PySyntaxRewriter::py_insertAtBack, py::arg("list"),
              py::arg("newNode"), py::arg("separator") = Token())
-        .def_property_readonly("factory", &PySyntaxRewriter::getFactory);
+        .def_property_readonly("factory", &PySyntaxRewriter::getFactory,
+                               py::return_value_policy::reference_internal,
+                               "Get the SyntaxFactory for creating new syntax nodes")
+        .def_property_readonly("alloc", &PySyntaxRewriter::getAllocator,
+                               py::return_value_policy::reference_internal,
+                               "Get the allocator for creating tokens and trivia")
+        .def("make_token", &PySyntaxRewriter::py_makeToken, py::arg("kind"), py::arg("text"),
```
MikePopoloski: Probably the names of the methods here should use camelCase like all of the other methods do, for consistency.

bindings/python/SyntaxBindings.cpp
```diff
@@ -463,7 +547,50 @@ void registerSyntax(py::module_& m) {
              py::arg("newNode"), py::arg("separator") = Token())
         .def("insert_at_back", &PySyntaxRewriter::py_insertAtBack, py::arg("list"),
              py::arg("newNode"), py::arg("separator") = Token())
-        .def_property_readonly("factory", &PySyntaxRewriter::getFactory);
+        .def_property_readonly("factory", &PySyntaxRewriter::getFactory,
+                               py::return_value_policy::reference_internal,
+                               "Get the SyntaxFactory for creating new syntax nodes")
+        .def_property_readonly("alloc", &PySyntaxRewriter::getAllocator,
+                               py::return_value_policy::reference_internal,
```
MikePopoloski: All properties automatically have reference_internal policies so this is redundant.

bindings/python/SyntaxBindings.cpp
```diff
@@ -96,6 +98,88 @@ class PySyntaxRewriter : public SyntaxRewriter<PySyntaxRewriter> {
 
     SyntaxFactory& getFactory() { return factory; }
 
+    BumpAllocator& getAllocator() { return alloc; }
+
+    Token py_makeToken(TokenKind kind, std::string_view text, std::span<const Trivia> trivia = {}) {
+        char* textCopy = reinterpret_cast<char*>(alloc.allocate(text.size(), 1));
```
MikePopoloski: All of these copies of the text could be done with one-liner: `toStringView(alloc.copyFrom(text))`

bindings/python/pyslang.h
```diff
@@ -218,6 +218,173 @@ class type_caster<bitmask<type>> {
...
+
+        // SyntaxNode* -static-> SyntaxListBase* and then SyntaxListBase* -reinterpret->
+        // SyntaxList<T>*
+        auto* listBase = static_cast<SyntaxListBase*>(node);
+        // this is an intentional reinterpret_cast as layout verified by static_assert above
+#if defined(__clang__)
+#    pragma clang diagnostic push
+#    pragma clang diagnostic ignored "-Wreinterpret-base-class"
+#endif
+        ptr = reinterpret_cast<SyntaxList<T>*>(listBase);
```
MikePopoloski: Why is the reinterpret_cast necessary here? You should be able to static_cast no?

MikePopoloski:
Hello, are you planning to continue this PR?

bindings/python/SyntaxBindings.cpp
```diff
@@ -463,7 +547,50 @@ void registerSyntax(py::module_& m) {
              py::arg("newNode"), py::arg("separator") = Token())
         .def("insert_at_back", &PySyntaxRewriter::py_insertAtBack, py::arg("list"),
              py::arg("newNode"), py::arg("separator") = Token())
```
CheeksTheGeek: @MikePopoloski 

Apologies for the delay, but wondering if I should also refactor insert_before, insert_after... to insertBefore, insertAfter and so on...?

MikePopoloski: Ah yes, that would be good. I think the consistency is important.

bindings/python/pyslang.h
CheeksTheGeek: yupp, missed that. Done!

bindings/python/SyntaxBindings.cpp
CheeksTheGeek: did not know that. understood!

scripts/syntax_gen.py
CheeksTheGeek: yeah back when I added it I saw the 390 methods for syntax factory and thought maybe should add a flag for it.

CheeksTheGeek:
Addressed all of them. 
Let me know if the tests cover everything.

Issue #1215, opened 2025-01-20
Python Refactor: Create Python submodules for the various slang namespaces


parker-research:
Migration of archived issue: https://github.com/MikePopoloski/pyslang/issues/404

CheeksTheGeek:
came here from from facing #1620,
the archived issue is no longer accessible but,
was wondering whether this would be the desired hierarchy usage? (check the example imports below)

tldr: (i) do we want to keep backwards compatibility on a root level even if the root level gets super messy? (ii) Is implementing submodules via pybind11's `def_submodule()` in the C++ binding files the right approach? (only alt would be a pure Python wrapper package that imports from a flat `_pyslang` C extension and re-exports types into submodule `__init__.py` but in my perspective that's worse, lmk if there's another approach!)



```python
SV = """
module counter #(parameter WIDTH = 8) (
    input  logic clk,
    input  logic rst,
    output logic [WIDTH-1:0] count
);
    always_ff @(posedge clk or posedge rst) begin
        if (rst)
            count <= '0;
        else
            count <= count + 1;
    end
endmodule
"""


def demo_submodule_imports():
    print("> style #1: import from submodules ('right way' moving forward)")
    from pyslang.ast import Compilation, VariableSymbol, PortSymbol
    from pyslang.syntax import SyntaxTree, SyntaxNode, SyntaxKind, rewrite
    from pyslang.parsing import Token, TokenKind
    from pyslang.numeric import SVInt
    tree = SyntaxTree.fromText(SV, "counter.sv")
    print(f"parsed: {tree.root.kind}")
    print(f"is module: {tree.root.kind == SyntaxKind.ModuleDeclaration}")
    first_token: Token = tree.root.getFirstToken()
    print(f"first token: '{first_token.valueText}' (kind: {first_token.kind == TokenKind.ModuleKeyword})")

    comp = Compilation()
    comp.addSyntaxTree(tree)
    print(f"diag: {len(comp.getAllDiagnostics())}")
    root = comp.getRoot()
    for inst in root.topInstances:
        print(f"module: {inst.name}")
        for member in inst.body:
            if isinstance(member, VariableSymbol):
                print(f"  VariableSymbol: {member.name} : {member.type}")
            elif isinstance(member, PortSymbol):
                print(f"  PortSymbol: {member.name}")

    val = SVInt(32, 42, True)
    print(f"SVInt: {val}")

    def noop(node: SyntaxNode, rewriter):
        pass
[...]

PR #1623, opened 2026-01-02
Add FlowAnalysis and LSPUtilities bindings to pyslang

CheeksTheGeek:
Happy new year!

I really hope it is fine to create a PR without creating an issue, though I can create one if it's needed for clean project maintenance purposes!

Wanted to add some more bindings to expose features to pyslang side:
- FlowAnalysis for allowing custom data flow analysis implementation on py side
- exposing LSPUtilities for handling Longest Static Prefix expressions (getBounds, stringifyLSP, visitLSPs) bindings
- exposing DriverKind enum

Please let me know if some tests or some changes are required :)

### DIFF
--- a/bindings/python/AnalysisBindings.cpp
+++ b/bindings/python/AnalysisBindings.cpp
@@ -253,10 +253,12 @@ void registerAnalysis(py::module_& m) {
         .def(py::init<const Symbol&, AnalysisOptions>(), "symbol"_a,
              "options"_a = AnalysisOptions(),
              "Create a flow analysis for the given symbol (procedural block, subroutine, etc.)")
-        .def("run", [](PyFlowAnalysis& self, const Statement& stmt) { self.run(stmt); }, "stmt"_a,
-             "Run the analysis on a statement")
-        .def("run", [](PyFlowAnalysis& self, const Expression& expr) { self.run(expr); }, "expr"_a,
-             "Run the analysis on an expression")
+        .def(
+            "run", [](PyFlowAnalysis& self, const Statement& stmt) { self.run(stmt); }, "stmt"_a,
+            "Run the analysis on a statement")
+        .def(
+            "run", [](PyFlowAnalysis& self, const Expression& expr) { self.run(expr); }, "expr"_a,
+            "Run the analysis on an expression")
         .def_property("currentState", &PyFlowAnalysis::getCurrentState,
                       &PyFlowAnalysis::setCurrentState, "The current flow state's user data")
         .def_property_readonly("bad", &PyFlowAnalysis::isBad,


bindings/python/AnalysisBindings.cpp
```diff
@@ -82,4 +179,88 @@ void registerAnalysis(py::module_& m) {
...
+                       "Callback to create initial state: () -> state");
+
+    py::class_<LSPUtilities>(
+        m, "LSPUtilities",
+        "Utility methods for working with Longest Static Prefix (LSP) expressions.\n\n"
+        "An LSP expression is the longest static prefix of an expression that can be\n"
+        "used to identify a particular driver of a variable. For example, in the\n"
+        "expression `a.b[3].c.d[2:0]`, if all of the selects are constant, then the entire\n"
+        "expression is the LSP. If instead we had `a.b[i].c.d[2:0]`, then the LSP would be\n"
+        "`a.b[i]`.")
```
MikePopoloski: The LSP would just be `a.b` since the `[i]` part is not static.

bindings/python/AnalysisBindings.cpp
```diff
@@ -82,4 +179,88 @@ void registerAnalysis(py::module_& m) {
...
+        .def_static(
+            "getBounds",
+            [](const Expression& lsp, Compilation& compilation) -> py::object {
+                ASTContext astCtx(compilation.getRoot(), LookupLocation::max);
+                EvalContext evalCtx(astCtx);
+                auto result = LSPUtilities::getBounds(lsp, evalCtx, lsp.type->getCanonicalType());
```
MikePopoloski: This is not correct; the type argument should be the type of the root of the LSP, which is the initial ValueSymbol in the chain.

bindings/python/AnalysisBindings.cpp
```diff
@@ -82,4 +179,88 @@ void registerAnalysis(py::module_& m) {
         .def_property_readonly("semanticLeadingClock", &AnalyzedAssertion::getSemanticLeadingClock)
         .def_property_readonly("root", &AnalyzedAssertion::getRoot)
         .def("getClock", &AnalyzedAssertion::getClock, "expr"_a, byrefint);
+
+    py::class_<PyFlowAnalysis>(
```
MikePopoloski: You should be using the `classh` function for basically all exposed Python classes now.

bindings/python/AnalysisBindings.cpp
```diff
@@ -82,4 +179,88 @@ void registerAnalysis(py::module_& m) {
...
+        m, "FlowAnalysis",
+        "A flow analysis visitor that walks statements with proper control flow handling.\n\n"
+        "Set callback attributes before calling run():\n"
+        "- onAssignment: called for each assignment expression\n"
```
MikePopoloski, line 244: This seems like a strange set of callbacks to expose. Is this sufficient to actually get useful work done with this class?

bindings/python/AnalysisBindings.cpp
```diff
@@ -82,4 +179,88 @@ void registerAnalysis(py::module_& m) {
...
+        "- onConditionalBegin: called when entering an if/else\n"
+        "- onCaseBegin: called when entering a case statement\n"
+        "- onBranchMerge: called when control flow paths merge (state1, state2) -> merged_state\n"
+        "- onStateCopy: called to copy state when forking (state) -> copied_state\n"
+        "- createTopState: called to create initial state () -> state")
+        .def(py::init<const Symbol&, AnalysisOptions>(), "symbol"_a,
+             "options"_a = AnalysisOptions(),
+             "Create a flow analysis for the given symbol (procedural block, subroutine, etc.)")
+        .def("runOnStatement", &PyFlowAnalysis::runOnStatement, "stmt"_a,
```
MikePopoloski: You can expose both `run` overloads using `py::overload_cast` if you want to keep this matching the C++ API.

MikePopoloski:
Hello, are you planning to continue this PR?

bindings/python/AnalysisBindings.cpp
CheeksTheGeek: done!

bindings/python/AnalysisBindings.cpp
CheeksTheGeek: yupp those were the ones I was using, somehow didn't see the other ones, added those too

bindings/python/AnalysisBindings.cpp
CheeksTheGeek: understood

bindings/python/AnalysisBindings.cpp
CheeksTheGeek: fixed

CheeksTheGeek:
added the test for this PR too.

CheeksTheGeek:
I hope this one works on windows

CheeksTheGeek:
@MikePopoloski was wondering if we can merge this one. I'll be making another PR about the Cpp-Python library syncing #1215 as well.

MikePopoloski:
Sorry for the delay, merged.

PR #1644, opened 2026-01-26
Add AllowVirtualIfaceWithOverride compatibility flag

thomasnormal:
## Summary

- Adds a new `AllowVirtualIfaceWithOverride` compilation flag that allows interface instances that are bind/defparam targets to be assigned to virtual interfaces
- This violates IEEE 1800-2017 but matches behavior of some commercial tools like Cadence Xcelium
- The flag can be enabled via `--allow-virtual-iface-with-override`
- Also included in the VCS compatibility flags set (`--compat=vcs`)

## Motivation

This change enables slang to compile UVM-based verification environments that use a common pattern where BFM instances are both:
1. Assigned to virtual interfaces (via `uvm_config_db`)
2. Targets of `bind` directives (for assertions)

Example from a real codebase (`JtagControllerDeviceAgentBfm.sv`):

```systemverilog
// Module instance
JtagControllerDeviceMonitorBfm jtagControllerDeviceMonitorBfm (...);

// Same instance assigned to virtual interface via config_db
uvm_config_db#(virtual JtagControllerDeviceMonitorBfm)::set(null,"*","jtagControllerDeviceMonitorBfm",jtagControllerDeviceMonitorBfm);

// bind directive targeting the same module
bind jtagControllerDeviceMonitorBfm JtagControllerDeviceAssertions TestVectrorTestingAssertions(...);
```

This pattern is accepted by Cadence Xcelium but rejected by strict IEEE 1800-2017 interpretation.

## Test plan

- [x] Added unit test `AllowVirtualIfaceWithOverride flag` in `HierarchyTests.cpp`
- [x] Verified existing `VirtualIfaceDefparam` test still passes (ensures the error is still reported without the flag)
- [x] Built and ran tests locally

### DIFF
--- a/source/driver/Driver.cpp
+++ b/source/driver/Driver.cpp
@@ -50,14 +50,14 @@ using namespace analysis;
     CompilationFlags::AllowRecursiveImplicitCall, \
     CompilationFlags::AllowBareValParamAssignment, \
     CompilationFlags::AllowSelfDeterminedStreamConcat, \
-    CompilationFlags::AllowMergingAnsiPorts, \
-    CompilationFlags::AllowVirtualIfaceWithOverride
+    CompilationFlags::AllowMergingAnsiPorts
 
 static constexpr auto vcsCompFlags = {VCS_COMP_FLAGS};
 static constexpr auto allCompFlags = {
     VCS_COMP_FLAGS,
     CompilationFlags::AllowTopLevelIfacePorts,
-    CompilationFlags::AllowUnnamedGenerate
+    CompilationFlags::AllowUnnamedGenerate,
+    CompilationFlags::AllowVirtualIfaceWithOverride
 };
 
 #define VCS_ANALYSIS_FLAGS \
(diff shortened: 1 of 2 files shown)


source/driver/Driver.cpp
```diff
@@ -50,7 +50,8 @@ using namespace analysis;
     CompilationFlags::AllowRecursiveImplicitCall, \
     CompilationFlags::AllowBareValParamAssignment, \
     CompilationFlags::AllowSelfDeterminedStreamConcat, \
-    CompilationFlags::AllowMergingAnsiPorts
+    CompilationFlags::AllowMergingAnsiPorts, \
+    CompilationFlags::AllowVirtualIfaceWithOverride
```
MikePopoloski: From my testing on edaplayground VCS reports an error for these cases, so we should not include this flag in vcs compat mode.

source/driver/Driver.cpp
```diff
@@ -250,6 +251,10 @@ void Driver::addStandardArgs() {
     addCompFlag(CompilationFlags::AllowUnnamedGenerate, "--allow-genblk-reference",
                 "Allow references to unnamed generate blocks via their external names "
                 "(e.g. genblk1)");
+    addCompFlag(CompilationFlags::AllowVirtualIfaceWithOverride,
+                "--allow-virtual-iface-with-override",
+                "Allow interface instances that are bind/defparam targets to be assigned "
+                "to virtual interfaces (matches Cadence Xcelium behavior)");
```
MikePopoloski: Please remove the parenthetical; it's unnecessarily specific since this flag is more broadly useful.

PR #1646, opened 2026-01-28
BaseClass assert issue

AndrewNolte:
The`baseClass` assert in `getBaseConstructorCall()` can trigger due to the fact that it's possible for `handleExtends()` to call `getBaseConstructorCall()` before finishing and assigning `baseClass`. I'm not sure if this is the correct fix though.


MikePopoloski:
I think this is not complete because it will result in caching the baseConstructorCall as null due to the error base type, but then once we finish the handleExtends call a real base type will have been resolved and then we'd get a different answer when resolving the base constructor. I'm going to land this though and then apply some fixes on top of it.

PR #1650, opened 2026-01-30
Add non-standard string concat operator

AndrewNolte:


scripts/diagnostics.txt
```diff
@@ -1241,7 +1242,7 @@ group extra = { empty-member empty-stmt dup-import pointless-void-cast case-gen-
                 case-wildcard-2state inferred-latch inferred-comb }
 
 group pedantic = { empty-pattern implicit-net-port nonstandard-escape-code nonstandard-generate format-multibit-strength
-                   nonstandard-sys-func nonstandard-foreach nonstandard-dist nonstandard-randomize isunbounded-param-arg udp-coverage }
+                   nonstandard-sys-func nonstandard-foreach nonstandard-dist nonstandard-randomize nonstandard-string-concat isunbounded-param-arg udp-coverage }
```
MikePopoloski: This appears to be an Xcelium-only extension so the warning should be an error by default. This is done by putting it in the default set and upgrading the severity in the Driver.

source/ast/expressions/OperatorExpressions.cpp
```diff
@@ -749,6 +749,19 @@ Expression& BinaryExpression::fromComponents(Expression& lhs, Expression& rhs, B
     bool good = false;
     switch (op) {
         case BinaryOperator::Add:
+            if (bothStrings) {
```
MikePopoloski: I believe this will break the handling for cases where both sides of the + are string literals, which should do integer addition instead. Please fix that and add a test case for it.

Issue #1654, opened 2026-02-04
Issue with forward slash in block comment


por3bski:
**Is your bug report related to the C++ slang project or the pyslang Python bindings?**
C++ slang

**Describe the bug**
The end of a block comment is not detected when there is forward slash just before it
```
test.sv:2:7: error: nested block comments are disallowed by SystemVerilog
    /*/*/
      ^
test.sv:4:1: error: block comment unclosed at end of file
```
**To Reproduce**
Issue can be reproduced with following code:

```verilog
module top();
	/*/*/
endmodule
```

MikePopoloski:
Fixed by e7543071421e18a94157a57365c33c129a4e3f66

PR #1655, opened 2026-02-04
Fix issue with forward slash in block comment

por3bski:
Fix for #1654

### DIFF
--- a/source/parsing/Lexer.cpp
+++ b/source/parsing/Lexer.cpp
@@ -1263,7 +1263,7 @@ void Lexer::scanBlockComment() {
                 advance(2);
                 break;
             }
-            else if (c == '/' && peek(1) == '*') {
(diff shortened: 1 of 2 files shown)


MikePopoloski:
Hmm, unfortunately VCS disagrees with this interpretation of block comment parsing, which means it's possible to construct source snippets that behave differently across simulators, like so:
```
module m;
  initial $display("Hello" /*/*/, "World" /**/);
endmodule
```
VCS prints "Hello", whereas Xcelium prints "Hello World"

Probably nobody is actually relying on this, and I think the Xcelium behavior makes more sense, so I'm inclined to accept this change. However I've been thinking this should probably be downgraded to a warning instead so I may tackle that myself.

MikePopoloski:
Superseded by e7543071421e18a94157a57365c33c129a4e3f66

PR #1665, opened 2026-02-12
tidy: test for unconnected input ports in instantiation

spomata:
New test to catch unconnected input ports in SV instantiations.

Co-authored by Claude Code.


MikePopoloski:
I think this is worthy of being a warning issued by slang proper during AST construction, in which case this wouldn't need to go in slang-tidy.

spomata:
I think slang already issues a `-Wunconnected-port` however it won't make a difference between input and output ports.
In such cases I thought that having a directed test would help.

MikePopoloski:
This is now done in 41df84d257ae308fe178fc825f40296dc0dadfa2 -- there are three new warnings, one for each port direction.

PR #1669, opened 2026-02-14
Allow trailing commas in ANSI port lists

thomasnormal:
## Summary
- Add `AllowTrailingSeparator` option to `parseList()` template methods in `ParserBase.h`
- Enable trailing comma tolerance for ANSI port list parsing in `Parser.cpp`
- Syntax like `module foo(input a, input b,);` is now accepted without diagnostics

## Motivation

Trailing commas in ANSI port lists are a practical necessity for conditional compilation. This is a well-known pain point with a formal standards proposal.

**Accellera/IEEE standards proposal:** [Mantis item 10156](https://www.accellera.org/images/eda/sv-bc/10156.html) from David A. Gates (AMD) proposes allowing trailing commas in ANSI port declarations and named port connection lists. The rationale:

> "If NEED_PORT_A is defined but NEED_PORT_B isn't, then you get syntax errors." [...] "This becomes exponentially worse with more ports."

The proposal notes that trailing commas would enable generating "safe" code across all conditional compilation scenarios without special-casing the first or last element.

**Code generators produce trailing commas:** Tools like icepll have been [observed generating modules with trailing commas](https://github.com/YosysHQ/icestorm/issues/74). While the fix was to remove the trailing comma from icepll's output, the pattern recurs across code generators because managing the last-element comma is error-prone.

**Existing tool behavior:** Yosys accepts trailing commas silently. Verilator issues a suppressible `NULLPORT` warning. Xcelium and Icarus Verilog reject them.

## LRM Reference
Not currently in IEEE 1800-2017. Proposed as Accellera [Mantis 10156](https://www.accellera.org/images/eda/sv-bc/10156.html) for a future revision.

## Test plan
- [x] Existing tests continue to pass (`ctest --output-on-failure -R unittests`)
- [x] New test: module declarations with trailing commas parse without errors
- [x] New test: module declarations without trailing commas are unaffected

### DIFF
--- a/scripts/diagnostics.txt
+++ b/scripts/diagnostics.txt
@@ -116,7 +116,7 @@ error ExpectedIdentifier "expected identifier"
 error ExpectedStringLiteral "expected string literal"
 error ExpectedIntegerLiteral "expected integer literal"
 error ExpectedToken "expected '{}'"
-error MisplacedTrailingSeparator "misplaced trailing '{}'"
+warning misplaced-trailing-separator MisplacedTrailingSeparator "misplaced trailing '{}'"
 error ImplicitNotAllowed "expected data type (implicit type name not allowed)"
 error InvalidAccessDotColon "invalid access token; '{}' should be '{}'"
 error ExpectedMember "expected member"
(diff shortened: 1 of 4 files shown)


MikePopoloski:
No commercial tool allows this. It is desirable to provide compatibility with Yosys, so we can make the error suppressible, but it should remain an error by default otherwise. The Verilator behavior just seems broken; a trailing comma in a non-ANSI port list creates a null port yes, but not in an ANSI port list.

thomasnormal:
Reworked: reverted the `AllowTrailingSeparator` parser option entirely. Instead, made `MisplacedTrailingSeparator` a suppressible warning (promoted to error by default in Driver), following the same pattern as `DuplicateDefinition`/`MixedVarAssigns`/etc.

Issue #1186, opened 2024-12-06
Binding not working for Ariane code


ChangyangZen:
**Describe the bug**
When I use Slang to parse the Ariane MMU SystemVerilog assertion's property module, Slang can not recognize some signals from TLB even if I put the binding file and tlb.sv in the command line.

The command line output is as following shows:
 slang  -Wimplicit-port-type-mismatch -y 'ariane/include/' -y 'ariane/src/riscv-dbg/src' ./ariane/include/riscv_pkg.sv ./ariane/src/riscv-dbg/src/dm_pkg.sv ./ariane/include/ariane_pkg.sv  ./ariane/src/tlb.sv ./ariane/src/ptw.sv ariane/src/mmu.sv mmu_prop.sv mmu_bind.sv
Top level design units:
    mmu

ariane/src/mmu.sv:185:10: warning: implicit named port 'pmpaddr_i' of type 'logic[15:0][53:0]' connects to value of inequivalent type 'logic[15:0][55:0]' [-Wimplicit-port-type-mismatch]
        .pmpaddr_i,
         ^~~~~~~~~
SVA/mmu/sva/mmu_prop.sv.2.save:74:65: error: could not resolve hierarchical path name 'i_itlb'
asgpt__flush_i_tlb_flush: assert property (flush_i |=> (mmu_prop.i_itlb.flushed && mmu_prop.i_dtlb.flushed));
                                                                ^~~~~~~
SVA/mmu/sva/mmu_prop.sv.2.save:74:92: error: could not resolve hierarchical path name 'i_dtlb'
asgpt__flush_i_tlb_flush: assert property (flush_i |=> (mmu_prop.i_itlb.flushed && mmu_prop.i_dtlb.flushed));
                                                                                           ^~~~~~~
**To Reproduce**
1. clone the Ariane repo: 
git clone https://github.com/morenes/cva6.git ariane
cd ariane
git checkout autosva
git submodule update --init --recursive --force src
cd ..
2. Edit mmu_prop.sv and mmu_bind.sv
vi mmu_prop.sv
`module mmu_prop 
  import ariane_pkg::*; #(
    parameter INSTR_TLB_ENTRIES     = 4,
    parameter DATA_TLB_ENTRIES      = 4,
    parameter ASID_WIDTH            = 1,
    parameter ariane_pkg::ariane_cfg_t ArianeCfg = 0//ariane_pkg::ArianeDefaultConfig
) (
    input  logic                            clk_i,
[...]

PR #1672, opened 2026-02-14
Store bind directive scope for port connection name resolution

thomasnormal:
## Summary
- Store the scope where each bind directive was written on `BindDirectiveInfo` and `InstanceSymbol`
- Implement fallback name lookup: port connection names first resolve in the target module scope, then in the bind scope
- Applies to implicit named ports, interface ports, and explicit port expressions
- Lazy scope resolution via `getBindDirectiveScope()` handles cases where scopes aren't fully elaborated at bind time

## Motivation

When a bind directive is written inside a module, verification engineers commonly need to connect signals from the bind scope (where the bind is written) to the bound instance's ports. Currently, this requires fully-qualified hierarchical paths, which are verbose and fragile.

**Pattern this enables:**

```systemverilog
module target(input wire sig);
endmodule

module wrapper;
    logic my_signal;
    target t(.sig(1'b0));
    // 'my_signal' resolves in wrapper's scope (the bind scope)
    bind target target_checker chk(.sig(my_signal));
endmodule

module target_checker(input wire sig);
endmodule
```

Without this fix, slang reports `my_signal` as undeclared because it only looks in `target`'s scope. Users must work around this with hierarchical paths (e.g., `wrapper.my_signal`), which breaks if the hierarchy is refactored.

### Real-world examples

1. **OpenHW Group cv32e40x-dv** ([uvmt_cv32e40x_tb.sv](https://github.com/openhwgroup/cv32e40x-dv/blob/main/tb/uvmt/uvmt_cv32e40x_tb.sv)): This RISC-V processor verification environment has 15+ bind directives. Many require testbench-scope signals and use explicit hierarchical paths as a workaround:

   ```systemverilog
   bind dut_wrap.cv32e40x_wrapper_i.core_i.if_stage_i.mpu_i
     uvmt_cv32e40x_pma_cov #(...) pma_cov_instr_i (
       // These MUST use hierarchical path because bind-scope fallback doesn't exist:
       .pma_status_i (uvmt_cv32e40x_tb.pma_status_instr),
       .pma_status_rvfidata_word0lowbyte_i
         (uvmt_cv32e40x_tb.pma_status_rvfidata_word0lowbyte),
       ...
     );
   ```
   With bind-scope fallback, the verbose `uvmt_cv32e40x_tb.` prefix would be unnecessary.

2. **slang issue [#1186](https://github.com/MikePopoloski/slang/issues/1186)**: A user reported bind not working for CVA6/Ariane RISC-V processor verification. The user expected port connections to resolve signals from the target module's scope (which contains sub-instances), and JasperGold accepted the pattern while slang rejected it. This demonstrates real verification engineers encountering scope resolution issues with bind.

3. **CIRCT project** ([thomasnormal/circt](https://github.com/thomasnormal/circt)): The CIRCT compiler's SystemVerilog frontend uses slang and has 11 bind-related test cases. Three of them rely on bind-scope fallback ([bind-parent-port.sv](https://github.com/thomasnormal/circt/blob/main/test/Conversion/ImportVerilog/bind-parent-port.sv), [bind-interface-port.sv](https://github.com/thomasnormal/circt/blob/main/test/Conversion/ImportVerilog/bind-interface-port.sv), [bind-interface-generate-scope.sv](https://github.com/thomasnormal/circt/blob/main/test/Conversion/ImportVerilog/bind-interface-generate-scope.sv)), while the other 5 bind-interface tests already use hierarchical paths and work with unpatched slang.

### Alternative: hierarchical paths

We are aware that these bind-scope tests could be rewritten using hierarchical paths instead. For example:

```systemverilog
// Current (needs bind-scope fallback):
module Top(input logic clk, ctrl);
  Target t(.clk(clk));
  bind Target Monitor mon(.ctrl(ctrl));      // ctrl resolves in Top
endmodule

// Alternative (works without patch):
module Top(input logic clk, ctrl);
  Target t(.clk(clk));
  bind Target Monitor mon(.ctrl(Top.ctrl));  // explicit hierarchical path
endmodule
```

**Pros of hierarchical paths (status quo, no patch needed):**
- Strictly LRM-compliant — IEEE 1800-2017 §23.11 says bind instantiation resolves "as if placed inside the target scope"
- Matches Xcelium (24.03) behavior, which rejects unqualified bind-scope names
- No slang changes required
- Explicit about where each signal comes from

**Cons of hierarchical paths:**
- Verbose — deeply nested paths like `uvmt_cv32e40x_tb.pma_status_rvfidata_word0lowbyte` are unwieldy
- Fragile — any hierarchy refactoring (renaming a module instance, moving it) breaks all bind port connections that reference it
- Doesn't work well with definition-level binds (`bind TargetModule ...`) which apply to ALL instances — a hierarchical path to the bind scope only makes sense for one specific instantiation context
- Inconsistent with regular instantiation intuition — in `child c(.clk(clk))`, `clk` resolves in the instantiator's scope, but in `bind target child c(.clk(clk))`, it resolves in the target's scope

### Design decision for you

We'd like your guidance on whether this fallback behavior is something you'd consider for slang, even as an opt-in flag. If not, we'll rewrite our CIRCT tests to use hierarchical paths and keep bind-scope as a downstream patch. Either way, the [bundled fixes](#bundled-fixes) below stand on their own — happy to split those into a separate PR if preferred.

## Bundled fixes

Two additional fixes are included because they prevent crashes when processing bind targets:

1. **Skip implicit return variable for void functions** (SubroutineSymbols.cpp, CallExpression.cpp): Per IEEE 1800-2017 Section 13.4.1, only non-void functions have an implicit return variable. Creating one for void functions causes name conflicts when a void function's name matches a parameter name in the bind target.

2. **Allow `virtual` flag on pre/post_randomize overrides** (ClassSymbols.cpp): Per IEEE 1800-2017 Section 18.6.1, `pre_randomize`/`post_randomize` are implicitly virtual, so explicitly marking them `virtual` should not trigger `InvalidRandomizeOverride`.

## LRM Reference
- [IEEE 1800-2017](https://ieeexplore.ieee.org/document/8299595) Section 23.11 (Binding auxiliary code to scopes or instances)
- [IEEE 1800-2017](https://ieeexplore.ieee.org/document/8299595) Section 13.4.1 (Return values and void functions)
- [IEEE 1800-2017](https://ieeexplore.ieee.org/document/8299595) Section 18.6.1 (Randomization methods)

## Test plan
- [x] All existing tests pass (`ctest --output-on-failure -R unittests` — 2052 test cases)
- [x] New test: bind directive with port connections referencing the bind scope compiles without errors
- [x] Serializer test updated to reflect void function return variable removal

### DIFF
--- a/include/slang/ast/symbols/InstanceSymbols.h
+++ b/include/slang/ast/symbols/InstanceSymbols.h
@@ -137,9 +137,7 @@ public:
     void setBindScope(const Scope* scope) const { bindScope_ = scope; }
 
     /// Sets the bind directive syntax for lazy scope resolution.
-    void setBindSyntax(const syntax::BindDirectiveSyntax* syntax) const {
-        bindSyntax_ = syntax;
-    }
+    void setBindSyntax(const syntax::BindDirectiveSyntax* syntax) const { bindSyntax_ = syntax; }
 
     void serializeTo(ASTSerializer& serializer) const;
 
--- a/source/ast/symbols/PortSymbols.cpp
+++ b/source/ast/symbols/PortSymbols.cpp
@@ -787,8 +787,7 @@ public:
         if (!bs)
             return nullptr;
         // Only use bind scope if it's inside an instance body (module scope).
-        if (bs->asSymbol().kind == SymbolKind::InstanceBody ||
-            bs->getContainingInstance())
+        if (bs->asSymbol().kind == SymbolKind::InstanceBody || bs->getContainingInstance())
             return bs;
         return nullptr;
     }
@@ -799,8 +798,7 @@ public:
         // port connection resolution. For bind instances, names that aren't found
         // in the target scope are looked up in the bind scope as a fallback
         // (see implicitNamedPort and getImplicitInterface).
-        scope(*instance.getParentScope()),
-        instance(instance), comp(scope.getCompilation()),
+        scope(*instance.getParentScope()), instance(instance), comp(scope.getCompilation()),
         lookupLocation(LookupLocation::after(instance)),
         context(scope, lookupLocation, ASTFlags::NonProcedural),
         connMap(portConnections, scope, lookupLocation) {
@@ -1819,8 +1817,8 @@ const Expression* PortConnection::getExpression() const {
         if (parentInstance.body.flags.has(InstanceFlags::FromBind)) {
             auto bindScope = parentInstance.getBindScope();
             bool isModuleBindScope = bindScope &&
-                (bindScope->asSymbol().kind == SymbolKind::InstanceBody ||
-                 bindScope->getContainingInstance());
+                                     (bindScope->asSymbol().kind == SymbolKind::InstanceBody ||
+                                      bindScope->getContainingInstance());
             if (isModuleBindScope && bindScope != scope) {
                 if (connectedSymbol) {
                     // For implicit named ports, symbol was already resolved.


MikePopoloski:
All commercial tools I tried do not allow this. RivieraPro, Xcelium, Questa, VCS. Verilator also fails. I think the linked slang issue is not relevant either; the issue there was that JasperGold appears not to error regardless of what you type into the port connection, so it's just busted completely.

Maybe this would be nice for you to have but slang is not the right place to be pioneering new ways of specifying Verilog syntax (and introducing new ways that codebases can be incompatible with other tools). I think if you want this you should be petitioning IEEE to add it to the LRM.

Unrelated fixes in this PR should be split out.

MikePopoloski:
Also your linked examples are just tests that Claude made up and added to your fork, they're not actually in CIRCT upstream right?

thomasnormal:
Yes, my bad. I won't send anymore "pure nice to have" feature requests.

PR #1673, opened 2026-02-17
Add __init__.pyi for type checker support

ilthraim:
Since the refactor to pyslang to reflect the C++ namespaces, static typecheckers cannot resolve submodule references of the form
`import pyslang`
`pyslang.syntax` <- throws missing attribute warning despite working at runtime

Adds `__init__.pyi` to fix false "missing attribute" warnings during type checking

### DIFF
--- a/pyslang/pyslang/__init__.py
+++ b/pyslang/pyslang/__init__.py
@@ -1,6 +1,7 @@
 import sys as _sys
 
 from pyslang.pyslang import *  # noqa: F401,F403
+
 # from pyslang.pyslang import driver  # noqa: F401
 # from pyslang.pyslang import __version__, analysis, ast, parsing, syntax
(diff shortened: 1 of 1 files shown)


CheeksTheGeek:
I started to look into it and I know you closed the PR because you already arrived at the same conclusion as me but just making a note, the stubs are auto-generated (https://github.com/MikePopoloski/slang/blob/0bc125627a1f7f34f1f6a40a6a99235b8b05e19f/bindings/generate_stubs.cmake#L48C1-L66C8), and I couldn't reproduce the warning so I am thinking the library wasn't built (I saw you're using the github repository as a dependancy on one of your repos just like I'm doing on some of my projects 😛, seems like both of us love living on the cutting-edge).

PR #1680, opened 2026-02-22
Use release for fmt and mimalloc when multiple available, remove erroneous mimalloc 2 forcing

timkpaine:
- Libraries may allow multiple available (as with vcpkg) in the form `Release;Debug`, so take the first if this is the case
- The mimalloc min version part accidentally forces mimalloc 2, where 3 is widely available now

xref: https://github.com/microsoft/vcpkg/pull/50116

### DIFF
--- a/external/CMakeLists.txt
+++ b/external/CMakeLists.txt
@@ -45,7 +45,8 @@ endif()
 
 if(fmt_FOUND)
   get_target_property(FMT_CFG fmt::fmt IMPORTED_CONFIGURATIONS)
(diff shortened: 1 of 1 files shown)


MikePopoloski:
mimalloc v3 is not fully production ready yet (the project says the current release is "rc3") which is why I hadn't updated yet. It's not generally safe to just take a different major version the project hasn't tested against, though for the case of mimalloc it's likely that things will build just fine.

timkpaine:
Just sharing for visibility, I'm going to patch for vcpkg and conda anyway

PR #1682, opened 2026-02-23
Allow nested line comments within block comments

mampcs:
This is a specific case where the nested line comment actually starts with //*.


MikePopoloski:
This already works in slang, you can just turn off the warning if you don't want it. This matches the behavior of e.g. Xcelium which warns but continues on. I suppose you could make the warning ignored by default in vcs mode, but I don't think we need a Lexer change for it.

PR #1683, opened 2026-02-23
Allow implicit extern or pure function return in vcs compat mode

mampcs:
VCS allows the same.


MikePopoloski:
Hmm, it seems every tool I can find allows this silently. The grammar does say a return type is explicitly required here but tools have decided otherwise. It's probably fine to just allow this unconditionally. If not, you'll need to add a new driver option to allow this (all changes made by setting --compat must also be achievable via individual flags).

PR #1684, opened 2026-02-23
Allow out of block method qualifiers in vcs compat mode

mampcs:
VCS allows the same.


MikePopoloski:
I would instead change this error to be a warning that gets upgraded to an error by default. Then users can just downgrade it if they want (and you can downgrade it automatically in VCS compat mode), without the need for a new parse option.

PR #1686, opened 2026-02-23
Convert no implementation found to warning in vcs compat mode

mampcs:
VCS considers this a warning instead of an error.

### DIFF
--- a/scripts/diagnostics.txt
+++ b/scripts/diagnostics.txt
@@ -370,7 +370,6 @@ error MaxGenerateStepsExceeded "generate loop hit maximum step limit; possible i
 error MixingSubroutinePortKinds "port declarations not allowed when subroutine defines a formal argument list"
 error UnpackedArrayParamType "unpacked array parameter requires explicit data type"
 error AutomaticNotAllowed "automatic variables can only be declared in procedural contexts"
-error NoMemberImplFound "no implementation found in parent scope for '{}'"
 error MethodKindMismatch "mismatch between 'function' and 'task' declared in prototype"
 error MethodReturnMismatch "return type {} of method '{}' does not match {} declared in its prototype"
 error MethodArgCountMismatch "definition of method '{}' has different number of arguments from its prototype"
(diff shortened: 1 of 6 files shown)


MikePopoloski:
This is fine, but needs to be done by making it a warning in diagnostics.txt (with associated documentation in warnings_docs.txt) and then upgrading it to an error in the Driver if compat mode is *not* VCS. This way the user gets a command line option they can use to toggle it on/off with or without the use of compat mode.

scripts/warning_docs.txt
```diff
@@ -822,6 +822,17 @@ module m;
 endmodule
 ```
 
+-Wno-member-impl-found
+The definition of a class method is not found.
```
MikePopoloski: This documentation seems copied from elsewhere. It should instead describe what causes the warning.

source/driver/Driver.cpp
```diff
@@ -700,6 +700,7 @@ bool Driver::processOptions() {
         diagEngine.setSeverity(diag::NonstandardSysFunc, DiagnosticSeverity::Ignored);
         diagEngine.setSeverity(diag::NonstandardForeach, DiagnosticSeverity::Ignored);
         diagEngine.setSeverity(diag::NonstandardDist, DiagnosticSeverity::Ignored);
+        diagEngine.setSeverity(diag::NoMemberImplFound, DiagnosticSeverity::Warning);
```
MikePopoloski: You don't need this line, it's a warning by default now.

scripts/warning_docs.txt
mampcs: Added a more detailed description.

scripts/warning_docs.txt
```diff
@@ -822,6 +822,19 @@ module m;
 endmodule
 ```
 
+-Wno-member-impl-found
```
MikePopoloski: This should be moved into the Compatibility section.

scripts/diagnostics.txt
```diff
@@ -573,6 +572,7 @@ warning unused-config-instance UnusedConfigInstance "unused config rule -- insta
 warning static-init-order StaticInitOrder "initializer for static variable '{}' refers to another static variable '{}' which will be initialized in an indeterminate order"
 warning static-init-value StaticInitValue "initializer for static variable '{}' refers to '{}' which will not have a value at initialization time"
 warning unnamed-generate UnnamedGenerate "generate block is unnamed"
+warning no-member-impl-found NoMemberImplFound "no implementation found in parent scope for '{}'"
```
MikePopoloski: I just realized this, but we don't start user-visible warning tags with "no" since the CLI syntax for disabling is to prefix with "no-". Could rename this to something like "missing-member-impl"

PR #1689, opened 2026-02-23
Add $deposit system task

mampcs:
This is supported by all major vendors, even thought it's not part of the LRM.

### DIFF
--- a/scripts/systemnames.txt
+++ b/scripts/systemnames.txt
@@ -143,6 +143,7 @@ $assertfailoff AssertFailOff
 $assertnonvacuouson AssertNonvacuousOn
 $assertvacuousoff AssertVacuousOff
 
+$deposit Deposit
 $display Display
 $displayb DisplayB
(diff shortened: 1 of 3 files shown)


MikePopoloski:
Please add a test case to SystemFuncTests.cpp so that there's at least one usage in the unit tests.

MikePopoloski:
Having looked up what $deposit actually does I don't think this is correct. This change only allows single bit inputs to be provided, and furthermore doesn't treat the first argument as an output. It looks like it needs to accept variably sized arguments?

mampcs:
Yes, it can accept variably sized arguments. Is there a type available for that?

MikePopoloski:
No, you need to write out a separate class for it and write the logic for figuring out its arguments dynamically. For example see StringOutputTask in this file. Also AssocArrayDeleteMethod has logic that bases its second argument on the type of the first one, which is what you want for $deposit too.

PR #1695, opened 2026-02-24
Fix the error in parsing nested macro with multiline args in nested define

### LINKED ISSUE #1693: Error parsing nested macro invocation
C++ slang

**Describe the bug**
Consider the following example:

```
`define MACRO_PARENT(code) \
  `define MACRO_FOOBAR 1 \
  code

`define MACRO_CHILD reg r;

module top;
`MACRO_PARENT(
  logic foobar;
  `MACRO_CHILD
);

endmodule
```

Slang throws the following error: 
```
nested_macros.sv:10:3: error: macro operators may only be used within a macro definition
  `MACRO_CHILD
  ^~~~~~~~~~~~
nested_macros.sv:5:21: note: expanded from macro 'MACRO_CHILD'
`define MACRO_CHILD reg r;
```

If on the other hand, the "logic foobar;" line is deleted, the code parses fine.

**To Reproduce**
No additional options are required.

CheeksTheGeek:
Fixes #1693
when a macro body contains a nested `define and is invoked with a multi-line argument, the preprocessor badly kept inDefineDirective set through LineContinuation tokens from the outer macro body, which caused weird LineContinuation tokens to be inserted into the expanded argument, triggering a MacroOpsOutsideDefinition error. should instead reset inDefineDirective when a LineContinuation token is encountered, since it marks a physical line boundary that ends the inner define's scope.

### DIFF
--- a/source/parsing/Preprocessor_macros.cpp
+++ b/source/parsing/Preprocessor_macros.cpp
@@ -510,8 +510,10 @@ bool Preprocessor::expandMacro(MacroDef macro, MacroExpansion& expansion,
     bool inDefineDirective = false;
 
     auto handleToken = [&](Token token) {
-        if (inDefineDirective && !token.isOnSameLine())
+        if (inDefineDirective &&
+            (!token.isOnSameLine() || token.kind == TokenKind::LineContinuation)) {
             inDefineDirective = false;
+        }
 
         if (token.kind != TokenKind::Identifier && !LF::isKeyword(token.kind) &&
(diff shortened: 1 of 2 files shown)


MikePopoloski:
I fixed this differently in 58bffed124e4bc402691d42049510c60b59a8b6a

The real problem was that the line continuation token sucked up the newline immediately after it, such that calling isSameLine on the following token would return true even though it was not on the same line.

PR #1696, opened 2026-02-24
Add Wloop-var-fork

### LINKED ISSUE #1629: New warning for catching bad-usage of loop variables within fork/join_none blocks.
The following code has a bug which is a relatively common occurrence during test-case development.  Because the thread is directly referencing an external loop variable inside a join/join_none block, there will be N instances of the task called each receiving an input value of N.

```Systemverilog
for (int i = 0; i < N; i++) begin
    fork
        do_something_parallel(i);
    join_none
end
```

The code needs to save the iterator-value to a local automatic-variable in order for the task to receive the correct values.

```Systemverilog
for (int i = 0; i < N; i++) begin
    fork
        automatic int inst = i;
        do_something_parallel(inst);
    join_none
end
```

A warning that can detect when external loop variables are referenced inside a fork/join_none block would help catch this common coding-error and save what is often a longer debug cycle to root-cause.

_Originally posted by @jcurtiss8086 in https://github.com/MikePopoloski/slang/discussions/1622_

CheeksTheGeek:
Adds `-Wloop-var-fork` (in `-Wextra`) to warn when a for/foreach loop variable is referenced directly inside a `fork/join_any` or `fork/join_none` block, where all spawned threads will share the loop variable and likely see its final value instead of the per-iteration value.

Fixes #1629

### DIFF
--- a/source/ast/expressions/MiscExpressions.cpp
+++ b/source/ast/expressions/MiscExpressions.cpp
@@ -104,14 +104,12 @@ Expression& ValueExpressionBase::fromSymbol(const ASTContext& context, const Sym
             return badExpr(comp, nullptr);
         }
         else if (flags.has(ASTFlags::ForkJoinVarScope) &&
-                 (symbol.kind == SymbolKind::Variable ||
-                  symbol.kind == SymbolKind::Iterator)) {
+                 (symbol.kind == SymbolKind::Variable || symbol.kind == SymbolKind::Iterator)) {
             auto parentScope = symbol.getParentScope();
             if (parentScope) {
                 auto parentSyntax = parentScope->asSymbol().getSyntax();
-                if (parentSyntax &&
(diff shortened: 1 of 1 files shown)


MikePopoloski:
I think this is not quite right. It's trivially broken by inserting intervening statement blocks that should otherwise not affect the outcome:
```
module m;
    task automatic do_something(int x); endtask
    initial begin
        for (int i = 0; i < 4; i++) begin
            fork
              begin : baz
                do_something(i);
              end
            join_none
        end
    end
endmodule
```

I think implementing this is surprisingly tricky with the current AST since it's hard to know where exactly in a statement tree an expression is being bound. It might require some more tracking of that positioning, not sure.

MikePopoloski:
Superseded by 937feb83d3513ab15cda88804ebe53e50bca563c

PR #1715, opened 2026-03-04
Suppress ShiftCountNegative in untaken branches

AndrewNolte:


source/ast/expressions/OperatorExpressions.cpp
```diff
@@ -1202,8 +1202,9 @@ bool BinaryExpression::propagateType(const ASTContext& context, const Type& newT
             if (auto cv = context.tryEval(*right_); cv && cv.isInteger()) {
```
MikePopoloski: Should probably skip this whole branch if in an unevaluated branch. Wshift-count-overflow should be treated the same.

PR #1721, opened 2026-03-10
Support for macro definitions with trailing spaces after line continu…

mampcs:
…ation

VCS supports trailing spaces after the line continuation in a macro definition, except for the first `define line.

### DIFF
--- a/source/parsing/Preprocessor.cpp
+++ b/source/parsing/Preprocessor.cpp
@@ -762,7 +762,7 @@ Trivia Preprocessor::handleDefineDirective(Token directive) {
                     // Also, the trailing space is not allowed for the first line of the macro
                     // definition.
                     if (!lexerOptions.allowMacroTrailingSpace || numContinuations == 1)
-                         hasContinuation = false;
+                        hasContinuation = false;
                     break;
                 default:
                     hasContinuation = false;


source/driver/Driver.cpp
```diff
@@ -1064,6 +1064,10 @@ void Driver::addParseOptions(Bag& bag) const {
     if (options.maxLexerErrors.has_value())
         loptions.maxErrors = *options.maxLexerErrors;
 
+    if (options.compat == CompatMode::Vcs) {
```
MikePopoloski: This also needs a separate command line flag (with docs) to enable the behavior independently. The `--compat` flag is just a convenience, it should always be possible to configure the individual behaviors without using the compat flag if desired.

tests/unittests/parsing/PreprocessorTests.cpp
```diff
@@ -3308,3 +3308,37 @@ TEST_CASE("Header guard -- directive after endif cancels") {
     preprocess(text);
     CHECK_DIAGNOSTICS_EMPTY;
 }
+
+TEST_CASE("Macro with trailing space after line continuation") {
```
MikePopoloski, line 3312: I'm a little worried that something might strip the whitespace and then suddenly this test provides no coverage without anyone noticing. Can you maybe add a `CHECK(text[whatever] == '\\' && text[whatever+1] == ' ');` to verify we're testing what we think we are?

PR #1722, opened 2026-03-10
Print the filename and line number in the preprocessor output

mampcs:
This is very helpful when dealing with codebases with nested includes.

### DIFF
--- a/include/slang/syntax/SyntaxPrinter.h
+++ b/include/slang/syntax/SyntaxPrinter.h
@@ -64,6 +64,15 @@ public:
     /// @return a reference to this object, to allow chaining additional method calls.
     SyntaxPrinter& print(const SyntaxTree& tree);
 
+    /// If no source manager is provided or if this flag is set, print tokens from any location.
+    /// Otherwise, filter out tokens based on other flags that control inclusion or exclusion of
+    /// preprocesses macros, includes, etc.
+    /// @return a reference to this object, to allow chaining additional method calls.
+    SyntaxPrinter& setIncludeAllLocations(bool include) {
+        includeAllLocations = include;
+        return *this;
+    }
+
     /// Sets whether to include trivia when printing syntax.
     /// @return a reference to this object, to allow chaining additional method calls.
     SyntaxPrinter& setIncludeTrivia(bool include) {
@@ -115,6 +124,13 @@ public:
         return *this;
     }
 
+    /// Sets whether to include source information when printing syntax.
+    /// @return a reference to this object, to allow chaining additional method calls.
+    SyntaxPrinter& setIncludeSource(bool include) {
+        includeSource = include;
+        return *this;
+    }
+
     /// Sets whether to squash adjacent newlines down into one when printing syntax.
     /// @return a reference to this object, to allow chaining additional method calls.
     SyntaxPrinter& setSquashNewlines(bool include) {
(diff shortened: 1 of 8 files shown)


include/slang/syntax/SyntaxPrinter.h
```diff
@@ -64,6 +64,13 @@ class SLANG_EXPORT SyntaxPrinter {
     /// @return a reference to this object, to allow chaining additional method calls.
     SyntaxPrinter& print(const SyntaxTree& tree);
 
+    /// Sets whether to include everything (trivia, expanded macros, expanded includes).
+    /// @return a reference to this object, to allow chaining additional method calls.
+    SyntaxPrinter& setIncludeAll(bool include) {
```
MikePopoloski: This is a very confusing flag since it doesn't actually include all, other flags still have an effect. Maybe something like `includeAllLocations`? Along with some documentation about what that means, like "If no source manager is provided, or if this flag is set, assume that tokens from any location should be included. Otherwise, filter out tokens based on other flags that control inclusion or exclusion of preprocessed macros, includes, etc."

source/syntax/SyntaxPrinter.cpp
```diff
@@ -13,6 +13,7 @@
 #include "slang/syntax/SyntaxTree.h"
 #include "slang/text/SourceLocation.h"
 #include "slang/text/SourceManager.h"
+#include "slang/util/OS.h"
```
MikePopoloski: Include seems unnecessary

tools/driver/slang_main.cpp
```diff
@@ -113,12 +113,15 @@ int driverMain(int argc, TArgs argv) {
...
         std::optional<bool> includeDirectives;
         std::optional<bool> obfuscateIds;
+        std::optional<bool> includeSource;
         driver.cmdLine.add("--comments", includeComments,
                            "Include comments in preprocessed output (with -E)");
         driver.cmdLine.add("--directives", includeDirectives,
                            "Include compiler directives in preprocessed output (with -E)");
         driver.cmdLine.add("--obfuscate-ids", obfuscateIds,
                            "Randomize all identifiers in preprocessed output (with -E)");
+        driver.cmdLine.add("--preprocess-source", includeSource,
```
MikePopoloski, line 123: Please add docs for the new flag.

AndrewNolte:
I don't think this option belongs in the syntax printer- the syntax printer has never printed any metadata about the tree, just subsets of the processed tokens.

It makes much more sense imo to add this to the `--cst-json` output; See `source/syntax/CSTSerializer.cpp`.

mampcs:
Actually this is something that the major vendors support. VCS for example has a way to generate a tokens.v (-Xman=4) which contains all the code in a single file. This file can then actually be parsed/compiled again. Putting the metadata in the json file doesn't give the ability to reparse the output.

MikePopoloski:
I think this is fine to add here, --cst-json is pretty different from running just the preprocessor with -E. It probably also makes sense to have this kind of functionality in --cst-json too.

include/slang/syntax/SyntaxPrinter.h
```diff
@@ -64,6 +64,15 @@ class SLANG_EXPORT SyntaxPrinter {
     /// @return a reference to this object, to allow chaining additional method calls.
     SyntaxPrinter& print(const SyntaxTree& tree);
 
+    /// If no source manager is provided or if this flag is set, print tokens from any location.
+    /// Otherwise, filter out tokens based on other flags that control inclusion or exclusion of
+    /// preprocesses macros, includes, etc.
+    /// @return a reference to this object, to allow chaining additional method calls.
+    SyntaxPrinter& setIncludeAllocations(bool include) {
```
MikePopoloski: "AllLocations" here and elsewhere. "Allocations" is something else.

mampcs: :-)

PR #1723, opened 2026-03-10
Option to dump out the lexer tokens and trivia

mampcs:
This is very helpful for debugging.

### DIFF
--- a/source/parsing/Lexer.cpp
+++ b/source/parsing/Lexer.cpp
@@ -17,6 +17,7 @@
 #include "slang/text/CharInfo.h"
 #include "slang/text/SourceManager.h"
 #include "slang/util/BumpAllocator.h"
+#include "slang/util/OS.h"
 #include "slang/util/ScopeGuard.h"
 #include "slang/util/String.h"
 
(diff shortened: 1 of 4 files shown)


source/parsing/Lexer.cpp
```diff
@@ -1686,8 +1687,17 @@ void Lexer::scanDisabledRegion(std::string_view firstWord, std::string_view seco
 template<typename... Args>
 Token Lexer::create(TokenKind kind, Args&&... args) {
     SourceLocation location(bufferId, size_t(marker - originalBegin));
-    return Token(alloc, kind, triviaBuffer.copy(alloc), lexeme(), location,
-                 std::forward<Args>(args)...);
+    Token token = Token(alloc, kind, triviaBuffer.copy(alloc), lexeme(), location,
+                        std::forward<Args>(args)...);
+    if (options.diagTokenStream) {
```
MikePopoloski, line 1692: I'd like to avoid adding another branch to one of the hottest code paths in the project if at all possible. Can this be done after the fact by printing out the tokens with the SyntaxPrinter?

mampcs: I think whatever we get with cst-json is sufficient. I hadn't tried that before. All the required information is already in there (although it's more verbose since it's json). Going to cancel this pull request.

mampcs: On a side note, I'm seeing that an input of size about 200MB ends up taking nearly 100G memory in slang, which does seem excessive. Is there any memory profiling capabilities in slang?

MikePopoloski: There is not, but you can at least start by narrowing down which stage of compilation is causing the blowup by trying with --preprocess, --parse-only, --disable-analysis, in that order, and see if anything changes.

MikePopoloski: Unless your design is legitimately enormous I expect this to be a bug somewhere, like something is stuck in a loop. If you manage to complete elaboration you can run with --time-trace and look at the flame graph to see which parts of the hierarchy are taking the most time to elaborate.

mampcs: There's still one more enhancement atleast i have to add first (related to protected code) before i hope to get slang through all the stages for the code i'm trying it on. Will then try to figure out what's going on with the memory usage.

Issue #1730, opened 2026-03-13
Bug with chained includes


mampcs:
**Is your bug report related to the C++ slang project or the pyslang Python bindings?**
C++ slang

**Describe the bug**
The nested macro includes are causing the token stream to be jumbled.
Here are the different files:

test.sv:
```
`include "defines1.svh"
`TEST_HEADER(test)
        virtual task test_body();
        endtask
    endclass
endmodule
```

defines1.sv:
```
`ifndef _defines1_svh
`define _defines1_svh

`define TEST_HEADER(TEST_NAME) \
`include "pkg.sv" \
module TEST_NAME; \
timeunit 1ps; timeprecision 1ps; \
    class mytest;

`endif
```

pkg.sv:
```
`ifndef _pkg_sc
`define _pkg_sc

package pkg;
    timeunit 1ps; timeprecision 1ps;

    class c;
        virtual function void f1();
            `include "body.svh"
        endfunction
    endclass

endpackage

`endif
```

body.svh:
```
`ifndef _body_svh
`define _body_svh

`endif
```

Run command: slang --preprocess test.sv

The generated tokens looks like this:
```
package pkg;
    timeunit 1ps; timeprecision 1ps;
    class c;
        virtual function void f1();

module test;
timeunit 1ps; timeprecision 1ps;
    class mytest;
        endfunction
    endclass
endpackage
        virtual task test_body();
        endtask
    endclass
endmodule
```

Notice that the "module test" ended up inside the package instead of the package getting defined first followed by the module.
VCS generates the following:

```
`timescale 1 ps / 1 ps
package pkg;
        class c;
                virtual function void f1;
                endfunction
        endclass
endpackage

`timescale 1 ps / 1 ps
module test;
        class mytest;
                virtual task test_body;
                endtask
        endclass
endmodule
```

Removing the `include "body.svh" from pkg.sv results in the correct code in slang.

**To Reproduce**
Noted in the description above.

MikePopoloski:
Fixed in 1b93683aa7423418f206cbee69bb965d18f549c3

PR #1731, opened 2026-03-14
Fix nested macro bug #1730

mampcs:


mampcs:
The fix for the bug was actually generated by Claude Code using the description of the bug #1730 as part of the prompt.

tests/unittests/parsing/PreprocessorTests.cpp
```diff
@@ -2999,6 +2999,89 @@ endmodule
     NO_COMPILATION_ERRORS;
 }
 
+TEST_CASE("Preprocessor: Include inside macro with nested include in included file") {
+    // Regression test: when a macro body contains `include "X", and X itself
```
MikePopoloski, line 3003: Just a note for the future: prefer to strip out this overly verbose comment novel that LLMs like to put into the tests, it's mostly just annoying.

PR #1732, opened 2026-03-14
Allow function calls in foreach array expression

mampcs:
The foreach loop's array name was parsed using parseName() which only handles identifiers, class scopes, and member selectors. It had no support for function call argument lists '()', so a foreach with an invocation in the array expression (e.g. foreach(c2::self().da2[i])) failed with spurious parse errors.

Introduce parseForeachArrayExpression() which first calls parseName(ForeachName) for the base name, then extends it in a loop to handle '()' invocations, '.' member accesses, and intermediate '[]' selectors, stopping before the final '[loop_variables]' bracket. Change ForeachLoopList.arrayName from Name to Expression in syntax.txt so the syntax node can hold the resulting ExpressionSyntax tree.

### DIFF
--- a/include/slang/parsing/Parser.h
+++ b/include/slang/parsing/Parser.h
@@ -195,6 +195,7 @@ private:
     syntax::SelectorSyntax* parseElementSelector();
     syntax::NameSyntax& parseName(bitmask<NameOptions> options);
     syntax::NameSyntax& parseNamePart(bitmask<NameOptions> options);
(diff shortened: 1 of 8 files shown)


MikePopoloski:
Looks like this is another VCS-specific feature, so you'll need to add an error-by-default warning to cover it.

MikePopoloski:
Heh, I tried to use GitHub's web merge UI and it failed terribly. Please rebase.

mampcs:
Rebase done.

PR #1733, opened 2026-03-14
Allow hierarchical names as cover cross targets

mampcs:
Cover cross items were restricted to simple identifiers, causing a parse failure when a class member was used as a target (e.g. port_cfg.m_lane_pol_inv).


MikePopoloski:
Looks like a duplicate of #1729 , with slightly different choices for picking a name, and that one issues a warning.

MikePopoloski:
I decided #1729 was more complete and so landed that one instead.

PR #1735, opened 2026-03-14
Allow package import declarations inside class bodies in VCS compat mode

mampcs:
IEEE 1800-2023 Clause 26, Syntax 26-2, note 15 explicitly states: "It shall be illegal to have an import statement directly within a class scope."

Some tools (e.g. VCS) support it anyway. Add a new downgradable diagnostic PackageImportInClass (warning in VCS compat mode, error otherwise) so users can control the severity.

Also fix the AST ClassPropertyDeclaration handler in Scope::addMembers, which silently dropped PackageImportDeclaration in its default case since it had only ever seen illegal members there. Now it delegates to the generic PackageImportDeclaration handler so wildcard and explicit imports are resolved from inside class method bodies.


MikePopoloski:
This needs to be rebased.

PR #1736, opened 2026-03-15
Added token rewriting functions to SyntaxRewriter

ilthraim:
Added the methods removeToken and replaceToken to the SyntaxRewriter class for modifying Tokens without having to necessarily touch the parent SyntaxNodes, alongside additional unit tests

The rewriter tracks the target tokens via parent node and index, and optionally takes a bool for preserving trivia or not. On replacement, the new node takes its trivia from the old node, and on removal the next token processed gets the trivia from the removed node. Please let me know if you disagree with this way of handling the trivia.

### DIFF
--- a/include/slang/ast/symbols/CoverSymbols.h
+++ b/include/slang/ast/symbols/CoverSymbols.h
@@ -193,8 +193,7 @@ public:
     CoverpointSymbol(Compilation& compilation, std::string_view name, SourceLocation loc);
 
     static CoverpointSymbol& fromSyntax(const Scope& scope, const syntax::CoverpointSyntax& syntax);
-    static CoverpointSymbol& fromImplicit(const Scope& scope,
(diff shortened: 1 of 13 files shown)


ilthraim:
Converting to a draft, it looks like there is an issue with the tests I didn't see when running locally. Investigating and will update PR when resolved

ilthraim:
Thanks to @arkanemoose for the help with this one

ilthraim:
I noticed the wasi-sdk build is failing, it looks like [upstream](https://github.com/WebAssembly/wasi-sdk) may have rolled back to v31.0 from v32.0 (which is what slang is trying to build against). Hopefully they put it back up at some point?

MikePopoloski:
Looks like they republished it.

MikePopoloski:
Looks pretty good, thanks for the PR.

PR #1737, opened 2026-03-15
Add ability to print all the macros groups by file

mampcs:


source/driver/Driver.cpp
```diff
@@ -753,6 +753,24 @@ void Driver::reportMacros() {
         printer.print(macro->body);
 
         OS::print(fmt::format("{}\n", printer.str()));
+    };
+
+    if (groupByFile) {
+        std::map<std::string_view, std::vector<const syntax::DefineDirectiveSyntax*>> byFile;
+        for (auto macro : preprocessor.getDefinedMacros()) {
+            auto fileName = sourceManager.getFileName(macro->directive.location());
```
MikePopoloski: I think you will want to do `sourceManager.getFullyOriginalLocation()` on the directive location, in case it happens to be defined inside another macro.

tests/unittests/DriverTests.cpp
```diff
@@ -236,6 +236,39 @@ __slang__ 1
...
+    driver.addStandardArgs();
+
+    auto filePath = findTestDir() + "test.sv";
+    const char* argv[] = {"testfoo", filePath.c_str()};
+    CHECK(driver.parseCommandLine(2, argv));
+    CHECK(driver.processOptions());
+    driver.reportMacros(/* groupByFile */ true);
+
+    // Each source file should appear as a "//<filename>:" header.
+    CHECK(stdoutContains("test.sv:"));
```
MikePopoloski: Rather than all this partial checking you can probably just check against one hardcoded string containing the result, there aren't that many macros or files involved here and the output should be stable across platforms.

tools/driver/slang_main.cpp
```diff
@@ -103,6 +103,9 @@ int driverMain(int argc, TArgs argv) {
         driver.cmdLine.add("-E,--preprocess", onlyPreprocess,
                            "Only run the preprocessor (and print preprocessed files to stdout)");
         driver.cmdLine.add("--macros-only", onlyMacros, "Print a list of found macros and exit");
+        std::optional<bool> groupMacrosByFile;
```
MikePopoloski: nit: Prefer this option declared up by the other variables.

MikePopoloski:
Sorry, I missed that there's no documentation for the new flag. Please add.

PR #1740, opened 2026-03-16
Fix header guard re-include after undef of guard macro

mampcs:
When a file with a header guard (ifndef/define/endif idiom) was included, slang recorded it in includeOnceHeaders so subsequent includes would be skipped. However, `undef-ing the guard macro had no effect on that set, causing the file to remain permanently skipped even after the user explicitly removed the guard.

Fix this by maintaining a reverse map (headerGuardMacros) from guard macro name to buffer pointer. When handleUndefDirective removes a macro, it now also removes the corresponding entry from includeOnceHeaders, allowing the file to be re-included as expected.

### DIFF
--- a/include/slang/parsing/Preprocessor.h
+++ b/include/slang/parsing/Preprocessor.h
@@ -487,9 +487,12 @@ private:
     // A buffer used to hold tokens while we're busy consuming them for directives.
     SmallVector<Token> scratchTokenBuffer;
 
-    // A set of files (identified by a pointer to the start of their text buffer) that
-    // have been marked pragma once so that we avoid trying to include them more than once.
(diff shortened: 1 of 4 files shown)


mampcs:
VCS uses a single namespace for the defines and the header guards. If this is not the expected default behavior, I can add this to compat mode.

MikePopoloski:
Oops yeah, overlooked this. The proposed fix pressimizes the case where the macro is undefined and then later redefined again prior to the include, which seems unlikely but might as well handle it. Rather than having the separate map, the includeOnce set can instead contain the name of the macro it's predicated on (if any), and then at include time we can check that said macro isDefined() at point of include.

mampcs:
Pushed suggested changes.

PR #1741, opened 2026-03-16
Add file kind tracking and debug output when parsing files

mampcs:
Adds a FileKind enum (Design, LibraryFile, LibraryMap) to SourceManager so that callers can determine how a buffer was originally loaded. SourceLoader sets the kind when reading buffers: LibraryMap for files loaded via --libmap, and LibraryFile for files loaded via -v/--libfile.

Adds a --show-parsed-files option to the preprocessor that prints the name and kind of each file as it is pushed onto or popped from the lexer stack, making it easier to trace which files are parsed and in what order. The first file processed is printed before the first pop so that it is not silently skipped.

### DIFF
--- a/include/slang/driver/Driver.h
+++ b/include/slang/driver/Driver.h
@@ -17,6 +17,8 @@
 #include "slang/text/SourceManager.h"
 #include "slang/util/Bag.h"
 #include "slang/util/CommandLine.h"
+#include "slang/util/ConcurrentMap.h"
+#include "slang/util/Function.h"
 #include "slang/util/LanguageVersion.h"
 #include "slang/util/OS.h"
 #include "slang/util/Util.h"
@@ -175,6 +177,10 @@ public:
         /// If true, the preprocessor will allow trailing spaces after the continuation character
         /// in macro definitions.
         std::optional<bool> allowMacroTrailingSpace;
+
+        /// If true, the preprocessor will print the name and kind of each file
+        /// as it is parsed.
+        std::optional<bool> showParsedFiles;
 
         /// @}
         /// @name Parsing
@@ -467,7 +473,8 @@ private:
     bool parseUnitListing(std::string_view text);
     std::string parseMapKeywordVersion(std::string_view value);
     void addLibraryFiles(std::string_view pattern);
-    void addParseOptions(Bag& bag) const;
+    void addParseOptions(Bag& bag,
+                         function_ref<void(BufferID, bool, bool)> bufferChangeCB = nullptr) const;
     void addCompilationOptions(Bag& bag) const;
     bool reportLoadErrors();
 
@@ -476,6 +483,9 @@ private:
     std::vector<std::tuple<std::string_view, std::string_view, std::string_view>>
         translateOffFormats;
     std::unique_ptr<JsonWriter> jsonWriter;
+
+    /// For storing the list of file parsing strings, keyed by thread index.
+    concurrent_map<size_t, std::vector<std::string>> parsedFiles;
 };
 
 } // namespace slang::driver
--- a/include/slang/parsing/Lexer.h
+++ b/include/slang/parsing/Lexer.h
@@ -118,6 +118,9 @@ public:
     /// Returns the library with which the lexer's source buffer is associated.
     const SourceLibrary* getLibrary() const { return library; }
 
+    /// Returns the BufferID of the source buffer being lexed.
+    BufferID getBufferId() const { return bufferId; }
+
     /// Concatenates two tokens together. This may result in more than one output token
     /// if the right hand token being concatenated ends up splitting and being re-lexed.
     /// Returns true if the concatenation succeeded and false otherwise.
--- a/include/slang/text/SourceManager.h
+++ b/include/slang/text/SourceManager.h
@@ -93,6 +93,24 @@ public:
     /// or nullptr if it's not explicitly part of any library.
     const SourceLibrary* getLibraryFor(BufferID buffer) const;
 
+    /// Describes the origin of a source buffer.
+    enum class BufferKind {
+        Unknown,
+        DesignFile,
+        LibraryFile,
+        LibraryMap,
+        IncludeFile,
+        Macro,
+        MacroArg
+    };
+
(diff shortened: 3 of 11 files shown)


include/slang/parsing/Lexer.h
```diff
@@ -118,6 +118,8 @@ class SLANG_EXPORT Lexer {
     /// Returns the library with which the lexer's source buffer is associated.
     const SourceLibrary* getLibrary() const { return library; }
 
+    BufferID getBufferId() const { return bufferId; }
```
MikePopoloski, line 122: Add a simple doc comment show it shows up in the generated API reference.

include/slang/text/SourceManager.h
```diff
@@ -93,6 +93,16 @@ class SLANG_EXPORT SourceManager {
...
     const SourceLibrary* getLibraryFor(BufferID buffer) const;
 
+    /// Describes the origin of a source buffer.
+    enum class FileKind { UnknownFile, DesignFile, LibraryFile, LibraryMap, IncludeFile };
+
+    /// Gets the file kind for the given buffer.
+    FileKind getFileKind(BufferID buffer) const;
+
+    /// Marks the given buffer as the specified file kind.
+    /// @note Not thread-safe; must be called before concurrent reads.
```
MikePopoloski: This note seems wrong? The method does the proper locking.

include/slang/text/SourceManager.h
```diff
@@ -238,6 +248,11 @@ class SLANG_EXPORT SourceManager {
     /// source manager.
     std::vector<BufferID> getAllBuffers() const;
 
+    /// Print out a message of the file being parsed and it's type.
+    /// If isBack is true, then we are back to parsing the specified file.
+    /// If isSkip is true, then we are skipping an already loaded header file.
+    void printParsing(BufferID buffer, bool isBack = false, bool isSkip = false) const;
```
MikePopoloski: This is a pretty weird method to live on SourceManager, which otherwise doesn't do any kind of OS output. How about instead of `bool showParsedFiles` in PreprocessorOptions we use a `function_ref<> bufferChangeCB` callback function. Then the Driver can set a callback and handle the printing itself. It would also let you remove the caveat about disabling parallel parsing, since the Driver can collect all the output per-thread and display it in a coherent manner at the end.

include/slang/text/SourceManager.h
```diff
@@ -93,6 +93,16 @@ class SLANG_EXPORT SourceManager {
     /// or nullptr if it's not explicitly part of any library.
     const SourceLibrary* getLibraryFor(BufferID buffer) const;
 
+    /// Describes the origin of a source buffer.
+    enum class FileKind { UnknownFile, DesignFile, LibraryFile, LibraryMap, IncludeFile };
```
MikePopoloski: It would be nice if this was actually `BufferKind` and included Macro and MacroArg.

include/slang/text/SourceManager.h
```diff
@@ -275,6 +290,8 @@ class SLANG_EXPORT SourceManager {
         std::vector<LineDirectiveInfo> lineDirectives;
 
         FileInfo() {}
+        FileKind fileKind = FileKind::UnknownFile;
```
MikePopoloski: Rather than UnknownFile, we can probably just default to DesignFile (or IncludeFile, which the SourceManager can determine itself) unless explicitly overridden.

Also nit: put this up with the other variables.

source/parsing/Preprocessor.cpp
```diff
@@ -126,6 +127,9 @@ void Preprocessor::pushSource(SourceBuffer buffer) {
         }
     }
 
+    if (options.showParsedFiles && includeDepth > 0) {
```
MikePopoloski: If you don't restrict by includeDepth here then won't need special casing in SourceLoader to print for the first entry.

source/driver/SourceLoader.cpp
```diff
@@ -139,6 +139,7 @@ void SourceLoader::addLibraryMapsInternal(std::string_view pattern, const fs::pa
             continue;
         }
 
+        sourceManager.setFileKind(buffer->id, SourceManager::FileKind::LibraryMap);
         auto tree = SyntaxTree::fromLibraryMapBuffer(*buffer, sourceManager, optionBag);
```
MikePopoloski, line 142: Probably fromLibraryMapBuffer should set the kind so it works as expected even if you aren't using the SourceLoader.

source/parsing/Preprocessor.cpp
mampcs: From what i observed, we first push all the source files. We then start lexing, which starts with the first file and once it is done, moves on to the next one in the stack. If we print the parsing message here always, we end up with saying parsing each file before we have actually started parsing it and the order is messed up as well. That's why i only handle the include files here. The first file is then special cased in SourceLoader and all subsequent files from the filelist are handled in popsource.

MikePopoloski: Hmm, if you are using `--single-unit` that is true. I often forget legacy designs like to jam all their files together.

There should be a way to get the ordering correct without involving the SourceLoader though, since there's otherwise no reason this functionality shouldn't work with anyone using the SyntaxTree API without using the SourceLoader.

mampcs: Yes i'm using --single-unit (because that's how VCS handles all our files). Let me see if this can be done without SourceLoader.

mampcs: Removed the changes in SourceLoader and added it just once in SyntaxTree.

include/slang/driver/Driver.h
```diff
@@ -471,6 +479,12 @@ class SLANG_EXPORT Driver {
     std::vector<std::tuple<std::string_view, std::string_view, std::string_view>>
         translateOffFormats;
     std::unique_ptr<JsonWriter> jsonWriter;
+
+    /// For storing the list of file parsing strings.
+    std::vector<std::string> parsedFiles;
```
MikePopoloski: Rather than mixing all of the output across threads, it would be nice if it was collected and shown per-thread at the end. This is straightforward to do, just switch this to a `concurrent_map<size_t, std::vector<std::string>>` and use `BS::this_thread::get_index()` as the key.

PR #1743, opened 2026-03-17
Allow reserved keywords to be used as identifiers

mampcs:
Add a --allow-keyword-as-identifier <kw> CLI option (comma-separated or repeatable) that causes the lexer to emit Identifier tokens instead of keyword tokens for the named keywords, enabling their use as variable or port names in legacy code.

A single keywordsAsIdentifiers vector in LexerOptions drives the check for all keywords. 'context' and 'config' are the first supported values.

DPI import parsing and SubroutineSymbol creation retain their existing identifier-fallback handling so that 'context' continues to work as a DPI property qualifier even when it is lexed as an identifier.

The unit test verifies that 'config' used as an identifier coexists with a real config block in the same compilation, confirming that config block parsing is not broken by the option.

The config-block pre-check in parseMemberImpl is guarded so it only fires when keywordsAsIdentifiers actually contains "config"; this avoids misparsing "config inst;" as a config declaration when 'config' is merely not a keyword in the current keyword-version mapping (e.g. 1364-2001-noconfig).

### DIFF
--- a/docs/command-line-ref.dox
+++ b/docs/command-line-ref.dox
@@ -187,6 +187,12 @@ If true, the preprocessor will allow trailing whitespaces after the continuation
 in a macro definition. The trailing whitespaces are only allowed after the first line of the
 macro definition.
 
+`--allow-keyword-as-identifier`
+
+If true, treat the named keyword as a plain identifier for compatibility with legacy code.
+May be specified multiple times or as a comma-separated list (e.g. 'context,config').
+Only supports context and config currently.
+
 `--translate-off-format <common>,<start>,<end>`
 
 Specify the format of a "translate off" directive in a comment that denotes a span
--- a/include/slang/driver/Driver.h
+++ b/include/slang/driver/Driver.h
@@ -171,6 +171,9 @@ public:
         /// in macro definitions.
         std::optional<bool> allowMacroTrailingSpace;
 
+        /// A list of keywords to treat as plain identifiers for compatibility with legacy code.
+        std::vector<std::string> allowKeywordsAsIdentifiers;
+
         /// @}
         /// @name Parsing
         /// @{
--- a/include/slang/parsing/Lexer.h
+++ b/include/slang/parsing/Lexer.h
@@ -81,6 +81,10 @@ struct SLANG_EXPORT LexerOptions {
     /// If true, the preprocessor will allow trailing spaces after the continuation character
     /// in macro definitions.
     bool allowMacroTrailingSpace = false;
+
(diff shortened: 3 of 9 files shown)


MikePopoloski:
Is this not achievable with the existing `--map-keyword-version` functionality?

mampcs:
Unfortunately that does not work because we have some IP files that are using config/context but are also using 1800-2005 constructs. This is mostly because VCS did not fully support all features in each of these extensions for a while and thus we ended up with users unknowingly using a some of the unimplemented keywords as identifiers.

MikePopoloski:
That's unfortunate. Does VCS maintain backward compatibility for this somehow or are you just never able to upgrade?

I'm still unsure why anything downstream from the Preprocessor needs to know about this change. Surely if you lex something as not a keyword then you don't need to *also* support it as a keyword?

mampcs:
VCS goes through a depreciation process usually. But for some of these, it will unfortunately continue to support it as both a keyword and identifier. Without the downstream change, this won't support a case where we want config for example to be both a keyword in the case where it's being used as a keyword (in let's say file1) and as an identifier in file2. That's exactly what we have with out RTL/VIP sources right now. We have RTL that is using config as a keyword but we have some VIPs that are using it as a keyword (and that unfortunately we cannot change the source for).

MikePopoloski:
Sure, but if you correctly lex it as a keyword in file1 and an identifier in file2, why does the parser need to know the difference? You've lexed it correctly in both cases. You then only need to do get the mapping correct per-file, like we already do with other keyword controlling options.

mampcs:
The problem is determining upfront whether to lex it as a keyword or identifier based on the file. They all have the same extension, so that's not going to work. Doing it based on specific file names would be a maintenance issue since we'll have to keep updating this mapping for each IP drop we get.

MikePopoloski:
Are your IP files where this would apply not in a common folder? The keyword mapping logic already supports arbitrary glob patterns, so it's straightforward to do something like `.../vendor_ip/.../*.v` to grab them all. Then this new keyword demotion logic could be applied in the preprocessor like the existing --map-keyword-version feature without infecting everything downstream of it.

mampcs:
I will try that out and see. I think we might be able to make that work.

MikePopoloski:
Did you get anywhere on this?

mampcs:
Sorry haven't had a chance to revisit this. Focused on the memory issue right now (filed an issue for that with some details on my investigation so far).

MikePopoloski:
Closing it out since there's been no movement and it seems like it should be handled by the existing mechanisms.

PR #1744, opened 2026-03-17
Handle missing scope end token when include file has protected code

mampcs:
When an included file contains an encrypted/protected region that hides the endmodule/endclass/endprogram/endpackage keyword, slang would previously report a parse error for the missing end token.

Add --allow-missing-protected-scope-end to suppress this error:
- The preprocessor tracks whether parsing is inside an outer scope element (enterOuterScope/exitOuterScope) and whether the current included file has encountered a protected region (setHasProtectedCode).
- When an included file's EOF is reached with both flags set and the option enabled, the EOF is surfaced to the parser instead of silently consumed.
- parseMemberList detects this via shouldIgnoreMissingEndToken() and synthesizes a missing end token rather than issuing a diagnostic.

### DIFF
--- a/tests/unittests/parsing/PreprocessorTests.cpp
+++ b/tests/unittests/parsing/PreprocessorTests.cpp
@@ -3450,12 +3450,11 @@ endmodule
 TEST_CASE("allowMissingProtectedScopeEnd - module") {
     // Include file where the module's endmodule is inside a protected region.
     SyntaxTree::getDefaultSourceManager().assignText(
-        "missing_endmodule.sv",
-        "module m;\n"
-        "`pragma protect begin_protected\n"
-        "`pragma protect encoding=(enctype=\"raw\"), data_block\n"
-        "encrypted_payload_containing_endmodule\n"
-        "`pragma protect end_protected\n");
+        "missing_endmodule.sv", "module m;\n"
+                                "`pragma protect begin_protected\n"
+                                "`pragma protect encoding=(enctype=\"raw\"), data_block\n"
+                                "encrypted_payload_containing_endmodule\n"
+                                "`pragma protect end_protected\n");
 
     PreprocessorOptions ppOpts;
     ppOpts.allowMissingProtectedScopeEnd = true;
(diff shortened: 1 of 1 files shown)


MikePopoloski:
Oof, this is pretty gnarly and intrusive. What's the story here, some vendor shipped you incorrectly protected IP and you have no way to get it fixed, and also can't exclude it from the files you pass to slang because it contains other definitions that are needed elsewhere?

mampcs:
Yes it's a mess. We got a couple of VIPs from 2 major vendors where the file has the module header and then ends with encrypted code. The endmodule is actually inside the encrypted code (because the code compiles fine with VCS which is able to parse the encrypted code). Modifying the VIPs is messy because we would have to do that each time we get a new drop. And yes, can't really exclude it from slang because they get included via nested includes from VIP sources itself. 
While we could do something to exclude these somehow, making that as a general solution is going to just break because we want to use slang extensively for analyzing sources on an ongoing basis.

I agree the change is very gnarly and intrusive. In the worst case we would have to maintain our own fork of slang and keep these kind of changes in there (which i would like to avoid if possible).

MikePopoloski:
Some thoughts on a more targeted approach here:
* Add a new callback to the Preprocessor options that gets invoked when we're about to return a new token after finishing an include file, right at the end of Preprocessor::nextRaw
* If this new option is set, the Parser hooks up that callback. It can look at the trivia of the incoming token and see that there is a protected directive in there.
* parseModule() should set a class-level member variable to the expected endKind, if parsing a top-level module definition.
* When the callback gets invoked, the Parser can see that there is an expected endKind, and fabricate the correct token and insert it into the ParserBase::Window, so that the loop in parseMemberList will get the correct token and naturally fall out of the loop. It may take some finessing to get the token in the right spot here given the flow of the callback but it seems doable conceptually.

This way there is only one extra branch in the preprocessor, on what is not really a hot path (falling out of the end of an include file), and no extra branches in the parser. The complexity is contained to the callback implementation.

source/parsing/Preprocessor.cpp
```diff
@@ -561,9 +565,14 @@ Token Preprocessor::nextRaw() {
     };
 
     auto token = getNext();
+
     if (token.kind != TokenKind::EndOfFile)
         return token;
 
+    // Capture whether the include file being popped had protected code, before
+    // popSource() restores the previous file's hasProtectedCode state.
+    bool includeHadProtectedCode = hasProtectedCode && (includeDepth > 0);
```
MikePopoloski: Do we need to know whether the include had protected regions *anywhere* in it, or only at the end? It seems like having it at the end of the file is sufficient to trigger the mitigation, in which case we don't need to do any tracking here. The callback can examine the token's trivia and find the protected region in there and act appropriately.

mampcs:
Simplified the code to only handle protected code with missing end keyword and not other includes. Using the EOF token to figure out if the file has protected code doesn't work because the in many cases there's a `endif after the protected code and before the EOF. However, since we only handle the simple case of protected code with no other includes, the stack is no longer required. Also, the changes are concentrated in the Preprocessor for the most part with no changes in the Parser now.

PR #1747, opened 2026-03-17
Stream AST JSON output to file instead of building in memory

mampcs:
Apply the same streaming approach used for the CST printer to printASTJson.  Add JsonWriter::flushTo(std::ostream&) which writes the accumulated content up to (but not including) the trailing separator to the given stream, then discards that content while retaining the separator in the buffer so subsequent writes remain well-formed.

A single ASTSerializer (and its underlying JsonWriter) is kept alive for the entire output so that shared state -- notably the set of already-printed enum types -- stays consistent across the design root and all definition serializations.  flushTo() is called after serializing the design root and after each individual definition, keeping peak memory proportional to the largest single serialized object rather than the sum of all objects.

The output format is byte-for-byte identical to the previous implementation.

### DIFF
--- a/tools/driver/slang_main.cpp
+++ b/tools/driver/slang_main.cpp
@@ -28,6 +28,25 @@ using namespace slang::driver;
 void printASTJson(Compilation& compilation, const std::string& fileName,
                   const std::vector<std::string>& scopes, bool includeSourceInfo,
                   bool detailedTypes) {
+    // Stream output directly to the destination.  A single JsonWriter (and
+    // ASTSerializer) is kept alive for the entire run so that shared state
+    // such as the set of already-printed enum types remains consistent across
+    // the design root and all definition serializations.  After each complete
+    // top-level value the buffer is flushed and cleared via flushTo(), keeping
+    // peak memory proportional to the largest single serialized object.
+
+    std::ofstream fileStream;
+    std::ostream* out;
+
+    if (fileName == "-") {
+        out = &std::cout;
+    }
+    else {
+        fileStream.open(fileName);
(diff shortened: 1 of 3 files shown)


MikePopoloski:
JSON is really not the right output format for large designs, it's very verbose. slang should probably have some way of round-trip serializing ASTs and CSTs to some binary format, with tools for consuming it from other languages / projects.

That said, that shouldn't prevent this improvement. `<iostreams>` are kind of garbage which is why I avoid them (see also https://llvm.org/docs/CodingStandards.html#include-iostream-is-forbidden) but I don't have the equivalent llvm:raw_ostream utility in slang yet so maybe I should just pull this in for now and mark that as a TODO.

tools/driver/slang_main.cpp
```diff
@@ -40,23 +59,29 @@ void printASTJson(Compilation& compilation, const std::string& fileName,
...
             auto sym = compilation.getRoot().lookupName(scopeName);
-            if (sym)
+            if (sym) {
                 serializer.serialize(*sym);
+                writer.flushTo(*out);
+            }
         }
     }
 
-    writer.writeNewLine();
```
MikePopoloski, line 58: It seems like this loses the trailing newline?

tools/driver/slang_main.cpp
```diff
@@ -40,23 +59,29 @@ void printASTJson(Compilation& compilation, const std::string& fileName,
...
-    OS::writeFile(fileName, writer.view());
+    // Write whatever remains in the buffer (closing brackets, final newline).
+    out->write(writer.view().data(), (std::streamsize)writer.view().size());
```
MikePopoloski: Should this just be a final call to flushTo() instead?

PR #1748, opened 2026-03-17
Stream CST JSON output to file instead of building in memory

mampcs:
Previously printCSTJson accumulated the entire JSON for all syntax trees in a single JsonWriter buffer before calling OS::writeFile, causing peak memory usage proportional to the full output size.

Rework the function to open the output destination once and serialize each syntax tree into a short-lived JsonWriter that is written to the file and destroyed immediately.  Peak memory is now proportional to the largest single tree rather than the sum of all trees.

Add JsonWriter::setInitialIndent so that per-tree writers can be started at the correct nesting depth (4 spaces), keeping the output format byte-for-byte identical to the previous implementation.  The outer envelope ("{\n  \"syntaxTrees\": [" ... "]\n}\n") is written directly
as literal strings, and per-tree separators ("\n    " / ",\n    ") are
written between consecutive tree objects.

### DIFF
--- a/tools/driver/slang_main.cpp
+++ b/tools/driver/slang_main.cpp
@@ -61,23 +61,46 @@ void printASTJson(Compilation& compilation, const std::string& fileName,
 
 void printCSTJson(Driver& driver, const std::string& fileName,
                   CSTJsonMode mode = CSTJsonMode::Full) {
-    JsonWriter writer;
-    writer.setPrettyPrint(true);
+    // Stream output directly to the destination instead of accumulating the
+    // entire JSON tree in memory first.  Each syntax tree is serialized into
+    // its own temporary JsonWriter buffer, written out immediately, and then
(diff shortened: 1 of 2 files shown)


tools/driver/slang_main.cpp
```diff
@@ -61,23 +61,46 @@ void printASTJson(Compilation& compilation, const std::string& fileName,
...
+    // discarded, keeping memory proportional to the largest single tree.
 
-    CSTSerializer converter(writer, mode);
+    std::ofstream fileStream;
+    std::function<void(std::string_view)> write;
 
-    writer.startObject();
-    writer.writeProperty("syntaxTrees");
-    writer.startArray();
+    if (fileName == "-") {
```
MikePopoloski, line 72: You could use std::cout as the output stream like you did in the other PR instead of the rigamarole with the std::function

Issue #1755, opened 2026-03-26
[slang] Invalid AST node for unresolved module instantiations


likeamahoney:
Hi @MikePopoloski 

While using a synthesis pipeline based on `yosys-slang`, I encountered an issue related to instantiating unknown (unresolved) modules.

Consider the following example:

```verilog
module top();
  hard unk(.i());
endmodule
```

When slang is invoked with the `--ignore-unknown-modules` option, I would expect the instantiation to be preserved in some form, since such constructs are generally allowed by the SystemVerilog standard (e.g. for black-box / hard macro integration).

However, slang currently generates an `Invalid` AST node for this case instead of keeping a valid representation of the instance.

This behavior seems questionable, as unresolved module instantiations are a common and valid use case in synthesis flows (e.g. for hard blocks or external IPs).

As a workaround, I can patch this in `yosys-slang`, but it feels more appropriate to handle this directly in `slang` and avoid producing `Invalid` nodes in this scenario (we can delete this [node creation]( https://github.com/MikePopoloski/slang/blob/b92cb5b014321cc917a5a39efe77c4c3f24cda8e/source/ast/symbols/InstanceSymbols.cpp#L1368)).

Is this behavior intentional? If not, I’d be happy to work on a fix.

MikePopoloski:
Your question is a bit confusing because we pretty clearly preserve the instantiation in the form of the UninstantiatedModule symbol. Are you referring specifically to empty port connections? You need some way to represent the empty connection, it doesn't have to be the Invalid node I suppose.

likeamahoney:
> Your question is a bit confusing because we pretty clearly preserve the instantiation in the form of the UninstantiatedModule symbol. Are you referring specifically to empty port connections? You need some way to represent the empty connection, it doesn't have to be the Invalid node I suppose.

Yes, my concern is specifically about empty port connections.

I understand that the instantiation itself is preserved via `UninstantiatedModule`, which is fine. The issue is more about how empty connections like `.i()` are represented in the AST.

Ideally, I would prefer one of the following:

either such empty connections are not materialized in the AST at all, or
they are represented with a dedicated node (instead of Invalid)

MikePopoloski:
[...]

PR #1756, opened 2026-03-26
Fix UninstantiatedDefSymbol invalid connections

likeamahoney:
Closes https://github.com/MikePopoloski/slang/issues/1755


MikePopoloski:
Looks like there's no test coverage of this path; please add a simple test to detect future breakages.

likeamahoney:
> Looks like there's no test coverage of this path; please add a simple test to detect future breakages.

Added

PR #1757, opened 2026-03-26
Add bits-of-integer-constant lint

AndrewNolte:
$bits() on an integer parameter are almost always a mistake.

### DIFF
--- a/scripts/diagnostics.txt
+++ b/scripts/diagnostics.txt
@@ -1076,6 +1076,7 @@ warning format-multibit-strength FormatMultibitStrength "formatting strength val
 warning nonstandard-sys-func NonstandardSysFunc "system function '{}' is non-standard"
 warning isunbounded-param-arg IsUnboundedParamArg "argument to $isunbounded must be the name of a parameter"
 warning dynamic-cast-const DynamicCastConst "$cast from {} to {} will always {}"
+warning bits-of-integer-constant BitsOfIntegerConstant "$bits of expression with type {} always returns '{}'; did you mean to use the constant value directly, or use $clog2?"
 
 subsystem ConstEval
 note NoteInCallTo "in call to '{}'"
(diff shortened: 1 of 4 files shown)


source/ast/builtins/QueryFuncs.cpp
```diff
@@ -47,6 +47,35 @@ class BitsFunction : public SystemSubroutine {
...
+            };
+
+            // Only warn for constant expressions (parameters, localparams,
+            // literals, etc.) -- not regular variables.
+            auto isConstExpr = [&] { return static_cast<bool>(context.tryEval(*args[0])); };
+
+            if (!isFromTypeParam() && isConstExpr()) {
+                TypePrinter printer;
+                printer.append(type);
+                context.addDiag(diag::BitsOfIntegerConstant, args[0]->sourceRange)
```
MikePopoloski: You don't need to TypePrinter this here, you can pass a Type as an argument to diagnostics. It will auto-add the single quotes around it, so you should remove them from the message.

source/ast/builtins/QueryFuncs.cpp
```diff
@@ -47,6 +47,15 @@ class BitsFunction : public SystemSubroutine {
                 diag.addNote(diag::NoteDeclarationHere, type.location);
             return comp.getErrorType();
         }
+
+        // const evalable non data type
+        if (args[0]->kind != ExpressionKind::DataType && context.tryEval(*args[0])) {
+            const Type& ct = type.getCanonicalType();
```
MikePopoloski: No need to getCanonicalType() here, all of the accessor methods on Type like isPredefinedInteger do that automatically.

scripts/diagnostics.txt
```diff
@@ -1076,6 +1076,7 @@ warning format-multibit-strength FormatMultibitStrength "formatting strength val
 warning nonstandard-sys-func NonstandardSysFunc "system function '{}' is non-standard"
 warning isunbounded-param-arg IsUnboundedParamArg "argument to $isunbounded must be the name of a parameter"
 warning dynamic-cast-const DynamicCastConst "$cast from {} to {} will always {}"
+warning bits-of-integer-constant BitsOfIntegerConstant "$bits of expression with type '{}' always returns '{}'; did you mean to use the constant value directly, or use $clog2?"
```
MikePopoloski: You need to remove the single quotes around the type argument, as the TypePrinter formatting is going to add them itself. I typically like to look at the messages produced by the expected-fail unit tests (just throw in a temporary NO_COMPILATION_ERRORS to see them) to make sure they look good.

PR #1761, opened 2026-03-29
Fix /* in libmap file paths being treated as block comments

mampcs:
Paths in library map declarations can contain glob wildcards that produce /* sequences (e.g. $IP_ROOT/*/stub/*.v). The SV lexer was consuming /* as a block comment start, swallowing the rest of the path and producing an 'is a directory' error when only the env var prefix survived parsing.

Add a filePathMode flag to the Lexer that prevents '/' from being processed as a comment start in trivia scanning. The Parser enables this mode around file path list parsing in library declarations and include statements.

### DIFF
--- a/include/slang/parsing/Lexer.h
+++ b/include/slang/parsing/Lexer.h
@@ -121,6 +121,10 @@ public:
     /// Returns the BufferID of the source buffer being lexed.
     BufferID getBufferId() const { return bufferId; }
 
+    /// When set to true, '/' is not treated as the start of a comment during trivia
+    /// scanning. Used when parsing library map file paths that may contain /* glob patterns.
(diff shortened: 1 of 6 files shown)


include/slang/parsing/Lexer.h
```diff
@@ -235,6 +239,10 @@ class SLANG_EXPORT Lexer {
 
     const SourceLibrary* library = nullptr;
     SourceManager& sourceManager;
+
+    // When true, '/' is not treated as the start of a comment in trivia scanning.
+    // Used when parsing library map file paths that may contain /* glob patterns.
+    bool filePathMode = false;
```
MikePopoloski: Move this up after `errorCount` for better packing, and more likely to be in cache when hit the branch condition.

PR #1763, opened 2026-03-29
Add public read-only accessor for collected DPI exports

### LINKED ISSUE #1762: Expose collected DPI export directives via a public `Compilation` accessor
C++ slang

Yes. `Compilation` already collects DPI export directives during elaboration, but there doesn't seem to be a public way to read them afterward.

Right now they're stored in `Compilation::dpiExports` as `(const syntax::DPIExportSyntax*, const Scope*)` pairs and recorded through `noteDPIExportDirective()`, but that list is private.

This makes DPI exports harder to work with because they don't become normal member symbols, so they don't show up in normal member iteration. Walking syntax can find `DPIExportSyntax` nodes, but not the paired `Scope*` that slang recorded during elaboration.

A small public read-only accessor on `ast::Compilation` for the collected DPI export entries.

Something like `getDPIExports()` returning the existing collected list would be enough.

Walking syntax trees to find `DPIExportSyntax` nodes.

That works for finding the syntax, but it doesn't give the same `(syntax, Scope)` information as the collected list.

**Additional context**
slang already has this information internally and already uses it for DPI export validation, so this seems like a small API gap.

hankhsu1996:
## Summary
- Add `Compilation::getDPIExports()` — a public const accessor returning `std::span<const DPIExportEntry>` over the existing private `dpiExports` collection.
- Add `Compilation::DPIExportEntry` type alias for `std::pair<const syntax::DPIExportSyntax*, const Scope*>`.
- No behavior, storage, or elaboration logic changes.

Closes #1762

## Test plan
- [x] New unit test "Compilation collects DPI exports" verifies count, SV names, explicit C identifier, and associated scope kind
- [x] All existing tests pass

### DIFF
--- a/source/ast/Compilation.cpp
+++ b/source/ast/Compilation.cpp
@@ -2005,7 +2005,7 @@ void Compilation::checkDPIMethods(std::span<const SubroutineSymbol* const> dpiIm
             }
 
             if (shouldRecordResolved)
-                dpiExports.emplace_back(&sub, std::string(cId));
+                dpiExports.push_back(DPIExport{&sub, std::string(cId), syntax});
         }
     }
 }
--- a/bindings/python/CompBindings.cpp
+++ b/bindings/python/CompBindings.cpp
@@ -135,8 +135,8 @@ void registerCompilation(py::module_& m, py::module_& ast, py::module_& driver) 
              [](const Compilation& self) {
                  py::list result;
                  auto parent = py::cast(&self);
-                 for (auto& [sub, cId] : self.getDPIExports())
-                     result.append(py::make_tuple(py::cast(sub, byrefint, parent), cId));
+                 for (auto& entry : self.getDPIExports())
+                     result.append(py::cast(&entry, byrefint, parent));
                  return result;
              })
         .def("getAttributes",
@@ -179,6 +179,11 @@ void registerCompilation(py::module_& m, py::module_& ast, py::module_& driver) 
         .def_readwrite("definition", &Compilation::DefinitionLookupResult::definition)
         .def_readwrite("configRoot", &Compilation::DefinitionLookupResult::configRoot)
         .def_readwrite("configRule", &Compilation::DefinitionLookupResult::configRule);
+
+    py::classh<Compilation::DPIExport>(comp, "DPIExport")
+        .def_readonly("subroutine", &Compilation::DPIExport::subroutine)
+        .def_readonly("cIdentifier", &Compilation::DPIExport::cIdentifier)
+        .def_readonly("syntax", &Compilation::DPIExport::syntax);
 
     py::classh<ScriptSession>(ast, "ScriptSession")
         .def(py::init<>())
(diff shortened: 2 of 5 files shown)


MikePopoloski:
I think the compilation should instead store the resolved DPI exports and expose those instead? Otherwise you will have to re-resolve them to the subroutine symbols. Like add a new member and expose that via accessor:
```
// Resolved DPI exports: (subroutine, C identifier) pairs.
std::vector<std::pair<const SubroutineSymbol*, std::string>> resolvedDPIExports;
```

hankhsu1996:
I think that makes sense. If the API can provide the resolved subroutine symbol together with the C identifier, that would probably be even better and more directly useful. I will update it.

source/ast/Compilation.cpp
```diff
@@ -2002,6 +2002,8 @@ void Compilation::checkDPIMethods(std::span<const SubroutineSymbol* const> dpiIm
                     diag.addNote(diag::NotePreviousDefinition, it->second->name.location());
                 }
             }
+
+            resolvedDPIExports.emplace_back(&sub, std::string(cId));
```
MikePopoloski: This should only be added if it's not a duplicate (in either of the map checks above).

hankhsu1996:
* Added the Python binding for `getResolvedDPIExports`, returning `(SubroutineSymbol, cId)` tuples with the symbol tied back to the owning `Compilation` for correct lifetime handling, and added a pyslang test for it.
* Changed resolved export recording so entries are only added if both duplicate checks pass.

hankhsu1996:
One API naming question: since this is the only public DPI export accessor, should this maybe just be `getDPIExports()`? It seems more consistent with the rest of `Compilation`, where the public getters expose the semantic result directly. If the raw directive collection remains internal, maybe that could be named something like `dpiExportDirectives` instead.

MikePopoloski:
I agree, it seems better to just name this getDPIExports

hankhsu1996:
- Renamed `ResolvedDPIExport` to `DPIExport`, `getResolvedDPIExports()` to `getDPIExports()` to match the existing
  Compilation getter style                                                                                  
  - Renamed internal raw collection from `dpiExports` to `dpiExportDirectives`; resolved collection takes the
  dpiExports name

hankhsu1996:
One follow-up API question: I agree the semantic form is the better public shape, but with `getDPIExports()` returning only `(subroutine, cIdentifier)`, the declaration-side information is lost from the public API. I've already run into a downstream case where I need to point diagnostics back to the original export declaration site.

Would it make sense for a `DPIExport` entry to preserve some link back to the export declaration as well, so the public API keeps both the semantic result and the declaration provenance?

MikePopoloski:
Yes that's fine, although you have to decide what to do about duplicate declarations. Like pick just the first one maybe?

hankhsu1996:
- Updated `DPIExport` from a pair to a named entry carrying `subroutine`, `cIdentifier`, and `syntax`.
- For duplicate declarations, I use the first surviving declaration as the canonical provenance, matching the current duplicate-handling behavior.
- Also updated the Python bindings and tests to match the new API shape.

PR #1773, opened 2026-04-02
Add --allow-lib-module-redef option for VCS-compatible module redefin…

mampcs:
…ition

When multiple library files (-v / --libfile) define a module, interface, program, or primitive with the same name in the same library, slang normally errors. VCS resolves such conflicts silently by keeping the first definition.

Add CompilationFlags::AllowLibModuleRedefinition and the --allow-lib-module-redef CLI option to match this behavior: when the conflicting definition comes from a library file, the first definition wins and the duplicate is silently discarded, regardless of whether the kinds match (e.g. module vs primitive). Non-library (regular source) redefinitions are still rejected as errors even when the flag is set.

The option is automatically enabled when --compat=vcs is used. Four unit tests and documentation are included.

### DIFF
--- a/source/ast/Compilation.cpp
+++ b/source/ast/Compilation.cpp
@@ -22,6 +22,7 @@
 #include "slang/syntax/AllSyntax.h"
 #include "slang/syntax/SyntaxTree.h"
 #include "slang/text/CharInfo.h"
+#include "slang/text/SourceManager.h"
 #include "slang/util/TimeTrace.h"
 
 using namespace slang::parsing;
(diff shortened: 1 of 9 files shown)


MikePopoloski:
Why is `Wno-duplicate-definition` not good enough here? Is the problem that primitives and other definitions still aren't allowed to duplicate? Like you have some library that defines both a primitive and a module with the same name?

mampcs:
The issue is that while Wno-duplicate-definition will allow a duplicate of a module or of a primitive, it won't allow them to be different i.e, module and primitive with the same name (as you mentioned above). Also, that would be a heavy hammer since it will allow it even for non libraries.

source/ast/Compilation.cpp
```diff
@@ -865,9 +866,27 @@ void Compilation::insertDefinition(Symbol& symbol, const Scope& scope) {
...
-                    // Duplicate in the same library. If they are both the same kind
-                    // then we report a warning and take the first one, otherwise
-                    // we give a hard error.
+                    // Duplicate in the same library.
+                    if (hasFlag(CompilationFlags::AllowLibModuleRedefinition)) {
+                        // VCS-like behavior: silently keep the first definition and
+                        // discard all subsequent ones, but only when the incoming
+                        // duplicate comes from a library file (-v / --libfile).
+                        bool isLibrary = false;
+                        if (symbol.kind == SymbolKind::Definition) {
```
MikePopoloski, line 875: I think you just want to check that the source library (vLib / symLib) is not `isDefault`? For example you can specify a library via library maps as well.

mampcs: I tried the unit test and in this case, both vLib and symLib were default. Checking for !isDefault would be false, so we would end up issuing an error.

MikePopoloski: Ah yeah, I misunderstood. -v will put files into the default library but still mark them as a library units.

PR #1774, opened 2026-04-02
Add CST memory usage statistics (--cst-memory-stats)

mampcs:
Introduces a new --cst-memory-stats option that reports detailed memory usage for the concrete syntax tree: driver components (source manager, syntax tree allocators), and a syntax node data breakdown by kind, token, trivia, and scope.

Supporting infrastructure added:
- BumpAllocator::getTotalAllocatedBytes() via a per-segment allocatedSize field
- Token::getReferencedSize() / getTriviaArraySize() for heap beyond sizeof(Token)
- SourceManager::getMemoryUsage() for raw source text + line-offset cache bytes
- syntax_gen.py: generate getSize() on each SyntaxNode covering all non-child members (token info blocks, list span arrays, separator tokens)
- MemStatsUtil: shared helpers for RSS query and human-readable byte formatting

### DIFF
--- a/docs/command-line-ref.dox
+++ b/docs/command-line-ref.dox
@@ -340,6 +340,22 @@ Controls the level of detail included in the CST JSON output. The available mode
 
 The default mode is `full` if this option is not specified.
 
+`--cst-memory-stats <file>`
+
+Dump memory usage statistics for the concrete syntax tree to the specified file, or '-' for
+stdout. May be invoked as soon as parsing is complete, even when elaboration is skipped.
+
+The report is organized into three sections:
+
+- **Driver Components** — bytes consumed by the source manager (raw file content and cached
+  line-offset tables) and by the bump allocators backing each parsed syntax tree and library
+  map tree.
+- **Syntax Node Data** — total bytes attributed to syntax nodes, with a per-kind breakdown
+  sorted by memory consumption (count, total bytes, and average bytes per node), followed by
+  token and trivia heap totals.
+- **Module/Package/Scope breakdown** — node memory grouped by the enclosing
+  module/interface/program/package scope, sorted by memory consumption.
+
 @section depfiles Dependency Files
 
 These options output various dependency files as filelists in addition to other actions.
--- a/include/slang/parsing/Token.h
+++ b/include/slang/parsing/Token.h
@@ -185,6 +185,14 @@ public:
     bool valid() const { return info != nullptr; }
     explicit operator bool() const { return valid(); }
 
+    /// Returns the size in bytes of all heap-allocated data referenced by this token beyond
+    /// sizeof(Token): the Info block (including any extra type-specific data) plus the trivia
+    /// array.
+    size_t getReferencedSize() const;
+
+    /// Returns the size in bytes of the trivia array separately allocated for this token.
+    size_t getTriviaArraySize() const;
+
     bool operator==(const Token& other) const { return kind == other.kind && info == other.info; }
 
     /// Modification methods to make it easier to deal with immutable tokens.
--- a/include/slang/syntax/CSTMemoryStats.h
+++ b/include/slang/syntax/CSTMemoryStats.h
@@ -1,0 +1,21 @@
+//------------------------------------------------------------------------------
+// CSTMemoryStats.h
+// Memory usage statistics for the CST (Concrete Syntax Tree).
+//
+// SPDX-FileCopyrightText: Michael Popoloski
+// SPDX-License-Identifier: MIT
+//------------------------------------------------------------------------------
+#pragma once
+
+#include <string>
+
+#include "slang/util/Util.h"
+
+namespace slang::driver {
+class Driver;
+}
+
+/// Prints memory usage statistics for all parsed CST syntax trees held by @a driver.
+/// May be called as soon as parsing is complete, even when elaboration is skipped.
+/// Output is written to @a fileName; use "-" for stdout.
+SLANG_EXPORT void printCSTMemoryStats(slang::driver::Driver& driver, const std::string& fileName);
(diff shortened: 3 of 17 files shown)


docs/command-line-ref.dox
```diff
@@ -340,6 +340,22 @@ Controls the level of detail included in the CST JSON output. The available mode
 
 The default mode is `full` if this option is not specified.
 
+`--cst-memory-stats <file>`
```
MikePopoloski, line 343: This should probably just be --memory-stats, i.e. one option for all memory stats output, once you add more. The user can always run with --parse-only if they only want CST stats. Also please put this down in the Profiling section

include/slang/parsing/Token.h
```diff
@@ -185,6 +185,14 @@ class SLANG_EXPORT Token {
     bool valid() const { return info != nullptr; }
     explicit operator bool() const { return valid(); }
 
+    /// Returns the size in bytes of all heap-allocated data referenced by this token beyond
+    /// sizeof(Token): the Info block (including any extra type-specific data) plus the trivia
+    /// array.
+    size_t getReferencedSize() const;
```
MikePopoloski, line 191: How about getHeapSize() instead? ReferencedSize is confusing to me.

include/slang/parsing/Token.h
```diff
@@ -185,6 +185,14 @@ class SLANG_EXPORT Token {
...
+
+    /// Returns the size in bytes of the trivia array separately allocated for this token.
+    size_t getTriviaArraySize() const;
```
MikePopoloski, line 194: This method seems unnecessary; it's easy enough to just look at trivia() and figure out the size from there.

include/slang/syntax/CSTMemoryStats.h
```diff
@@ -0,0 +1,21 @@
...
+#include "slang/util/Util.h"
+
+namespace slang::driver {
+class Driver;
+}
+
+/// Prints memory usage statistics for all parsed CST syntax trees held by @a driver.
+/// May be called as soon as parsing is complete, even when elaboration is skipped.
+/// Output is written to @a fileName; use "-" for stdout.
+SLANG_EXPORT void printCSTMemoryStats(slang::driver::Driver& driver, const std::string& fileName);
```
MikePopoloski, line 21: The syntax / parsing namespace shouldn't call back up into the Driver. Instead this function should just take a list of syntax trees to operate on. Also this should probably take the ostream as an argument so that e.g. tests can skip having to write to file and read it back again.

MikePopoloski: Also this can probably be made a static method on SyntaxTree once the above stuff is done.

include/slang/util/MemStatsUtil.h
```diff
@@ -0,0 +1,47 @@
...
+        return fmt::format("{:.2f} KB", d / 1024.0);
+    else if (bytes < 1024 * 1024 * 1024)
+        return fmt::format("{:.2f} MB", d / (1024.0 * 1024.0));
+    else
+        return fmt::format("{:.2f} GB", d / (1024.0 * 1024.0 * 1024.0));
+}
+
+// Returns the current RSS (resident set size) of the process in bytes.
+// Reads /proc/self/status on Linux; returns 0 on unsupported platforms.
+static inline size_t memStatsGetProcessRSS() {
```
MikePopoloski, line 32: This should be moved to the OS class, and an implementation added for Windows and macos. Here's how it looks on windows: https://learn.microsoft.com/en-us/windows/win32/psapi/collecting-memory-usage-information-for-a-process

Actually this SO thread is pretty good: https://stackoverflow.com/questions/63166/how-to-determine-cpu-and-memory-consumption-from-inside-a-process

include/slang/util/MemStatsUtil.h
```diff
@@ -0,0 +1,47 @@
...
+
+#include <cstddef>
+#include <cstdio>
+#include <fmt/format.h>
+#include <fstream>
+#include <string>
+
+namespace slang {
+
+static inline std::string memStatsFmtBytes(size_t bytes) {
```
MikePopoloski, line 18: You can put this in String.h to avoid needing a new header with just one function.

scripts/syntax_gen.py
```diff
@@ -382,6 +382,58 @@ def generateSyntax(builddir, alltypes, kindmap):
...
+            #   SyntaxList<U>  - sizeof in sizeof(*this); add the pointer span array
+            #                    (size*sizeof(U*)); do NOT recurse into U nodes —
+            #                    the visitor counts those separately.
+            #   SepSyntaxList  - sizeof in sizeof(*this); add the TokenOrSyntax span
+            #                    array and getReferencedSize() for separator Tokens;
+            #                    do NOT recurse into U nodes.
+            #   U* / not_null  - pointer already in sizeof(*this); nothing extra.
+            outf.write("    size_t getSize() const {\n")
+            outf.write("        size_t total = sizeof(*this);\n")
+            for m in currtype.combinedMembers:
```
MikePopoloski, line 403: It seems like all of this should be doable with the existing visitor / getChild methods on SyntaxNode, so we shouldn't need to generate new methods for every derived syntax type.

include/slang/util/MemStatsUtil.h
MikePopoloski: Added getPeakMemoryBytes() here: https://github.com/MikePopoloski/slang/commit/b3b324ec35eb473182f29d7cacf7c49f1c5f3d80#diff-1ce7b2ab736ec93103f8933a8e7ef364083acb5a41539828346f2d6942fb0c1eR78

MikePopoloski:
Superseded by 58af4f0037563a18a232d0cb40ef0bdde89385b9

PR #1775, opened 2026-04-02
Free initialClone as soon as defparam/bind discovery is complete

mampcs:
Wrap the initialClone Compilation object in a nested scope so it is destroyed immediately after the defparam/bind discovery phase, before the subsequent fixpoint iteration that allocates fresh Compilation objects.

### DIFF
--- a/source/ast/Compilation.cpp
+++ b/source/ast/Compilation.cpp
@@ -2444,9 +2444,6 @@ void Compilation::resolveDefParamsAndBinds() {
         // constantly mucking with parameter values in ways that can change the actual
         // hierarchy that gets instantiated. Cloning lets us do that in an isolated context
         // and throw that work away once we know the final parameter values.
-        Compilation initialClone({}, defaultLibPtr);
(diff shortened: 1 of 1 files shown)


source/ast/Compilation.cpp
```diff
@@ -2460,37 +2457,47 @@ void Compilation::resolveDefParamsAndBinds() {
...
+        } // initialClone freed here
 
-        // If we have found more binds, do another visit to let them be applied
-        // and potentially add blocks and defparams to our set for this level.
-        if (initialClone.bindDirectives.size() > numBindsSeen) {
-            // Reset the number of defparams seen to ensure that
-            // we re-resolve everything after the next iteration.
-            numDefParamsSeen = 0;
-            numBindsSeen = initialClone.bindDirectives.size();
+        if (moreBinds)
```
MikePopoloski: This moreBinds variable is unnecessary right? You can just `continue` like the old code used to.

PR #1791, opened 2026-04-16
Downgrade multiple errors to warnings in VCS compat mode

mampcs:


mampcs:
Sorry had to submit multiple commits in a single pull request. Splitting them up was a pain since they all conflict with each other as they touch the same pieces of code.

docs/command-line-ref.dox
```diff
@@ -38,6 +38,13 @@ By default all files are considered to be separate compilation units and orderin
 does not matter. When this option is provided, all files are concatenated together, in order, to
 produce a single compilation unit. See @ref compilation-units for more discussion.
 
+`--libs-in-own-unit`
```
MikePopoloski, line 41: Curious why this is necessary. Is there some conflict in names that gets hidden when they're in separate units? If so, I wonder if we can just make this the default behavior and not need yet another flag.

scripts/warning_docs.txt
```diff
@@ -1945,6 +1958,166 @@ module m;
...
+    end
+endmodule
+```
+
+-Wparam-class-covariance
+The LRM (§6.22.3) does not allow implicit assignment between different specializations
+of the same parameterized class. Some tools allow this covariant handle assignment
+anyway, so the error can be downgraded for compatibility.
+```
+class Base; endclass
```
MikePopoloski, line 2014: This definitely does not seem kosher; it's trivially easy to construct cases where you'd have expressions thinking they're accessing a variable of a particular type but the class member has a different type entirely. I tried this in VCS via edaplayground and it properly reports an error, so we need to see what your actual use case is.

scripts/warning_docs.txt
```diff
@@ -1945,6 +1958,166 @@ module m;
...
+        cp_cross : cross cp_a, cp_b;
+        cp_nested : cross cp_cross, cp_b {
+            bins c = binsof(cp_cross);
+        }
+    endgroup
+    cg cg_inst = new();
+endmodule
+```
+
+-Wempty-arg-not-allowed
```
MikePopoloski, line 1995: This seems too broadly permissive. It seems like you just want to allow this for sformatf, since these tools definitely don't allow empty arguments for arbitrary system functions.

scripts/warning_docs.txt
```diff
@@ -1945,6 +1958,166 @@ module m;
...
+endclass
+```
+
+-Wunpacked-in-assert-expr
+LRM §16.6 requires concurrent assertion expression types to be cast compatible
+with an integral type. Unpacked arrays do not meet this requirement. Some tools
+accept unpacked array expressions in concurrent assertions with only a warning,
+so the error can be downgraded for compatibility.
+```
+module m(input logic clk);
```
MikePopoloski, line 2115: I get an error when I try this with VCS.

What would the behavior of this even be?

source/ast/expressions/MiscExpressions.cpp
```diff
@@ -99,9 +99,9 @@ Expression& ValueExpressionBase::fromSymbol(const ASTContext& context, const Sym
         else if (flags.has(ASTFlags::ForkJoinAnyNone) && !var.flags.has(VariableFlags::RefStatic) &&
                  symbol.kind == SymbolKind::FormalArgument &&
                  symbol.as<FormalArgumentSymbol>().direction == ArgumentDirection::Ref) {
-            // Can't refer to ref args in fork-join_any/none
+            // Can't refer to ref args in fork-join_any/none per LRM.
+            // Non-standard: VCS allows this, so the diagnostic can be downgraded.
```
MikePopoloski, line 103: Let's avoid over-specializing comments in the code to a specific tool (i.e. VCS). Suffice it to say that some commercial tools allow it.

scripts/warning_docs.txt
```diff
@@ -1945,6 +1958,166 @@ module m;
...
+```
+module m;
+  bit flag;
+  int i = 0;
+  initial if (flag[i]) $display("set");
+endmodule
+```
+
+-Wundefined-param-override
+LRM §23.10 does not define behavior when a named parameter override refers to a
```
MikePopoloski, line 2065: I've been trying to avoid specific references to LRM sections, since every time they publish a new version they move things around and invalidate all the section numbers.

source/ast/expressions/SelectExpressions.cpp
```diff
@@ -55,9 +55,11 @@ static const Type& getIndexedType(TTypeProvider& typeProvider, const ASTContext&
         return typeProvider.getErrorType();
     }
     else if (ct.isScalar()) {
+        // Non-standard: VCS allows bit-select on a scalar; the result is the scalar's element
```
MikePopoloski, line 58: You probably have to update constant evaluation as well.

source/ast/symbols/CoverSymbols.cpp
```diff
@@ -1087,6 +1087,12 @@ BinsSelectExpr& ConditionBinsSelectExpr::fromSyntax(const BinsSelectConditionExp
         return badExpr(comp, nullptr);
 
     auto sym = nameExpr.getSymbolReference();
+    if (sym && sym->kind == SymbolKind::CoverCross) {
+        auto& diag = context.addDiag(diag::CrossIdentInBinsof, syntax.name->sourceRange());
+        diag << sym->name;
+        diag.addNote(diag::NoteDeclarationHere, sym->location);
+        return badExpr(comp, nullptr);
```
MikePopoloski, line 1094: You shouldn't return badExpr here if you want this to be a suppressible warning.

scripts/warning_docs.txt
mampcs: Here's the example that I pulled from our TB that VCS happily parses.

```
class Base;
endclass

class Derived extends Base;
endclass

class Callback #(type T = Base);
  T obj;
endclass

class EventHandler #(type T = Base);
  // Queue of callbacks parameterized with T
  Callback #(T) cbs[$];

  // Queue of callbacks parameterized with Base (the "object pool")
  Callback #(Base) base_cbs[$];

  // VCS allows assigning Callback#(Base) to Callback#(T) and vice versa
  // Slang rejects this as a type mismatch
  function void add_callback(Callback #(T) cb);
    Callback #(Base) b;
    b = cb;            // error: value of type 'Callback#(T)' cannot be assigned to type 'Callback#(Base)'
    base_cbs.push_back(b);
  endfunction

  function Callback #(T) get_callback();
    Callback #(Base) b = base_cbs[0];
    Callback #(T) result;
    result = b;        // error: value of type 'Callback#(Base)' cannot be assigned to type 'Callback#(T)'
    return result;
  endfunction

  function bit compare_callback(Callback #(T) cb);
    Callback #(Base) b = base_cbs[0];
    return (cb == b);  // error: invalid operands to binary expression
  endfunction

endclass
```

scripts/warning_docs.txt
mampcs: Here's the example:

```
module unpacked_array_in_property;

  logic clk;
  logic [11:0] ptr_in;
  logic [3:0]  src_in;
  logic [11:0] ptr_out;

  // Property with unpacked array local variables
  property p_unpacked_array_vars;
    bit [11:0] ptr[2];   // unpacked array local var
    bit [3:0]  src[2];   // unpacked array local var
    int        tag[2];   // unpacked array local var
    @(posedge clk)
    (1, ptr[0] = ptr_in, src[0] = src_in, tag[0] = 0)
    |=>
    (ptr_out == ptr[0]);
  endproperty

  assert property (p_unpacked_array_vars);

endmodule
```

scripts/warning_docs.txt
MikePopoloski: Ok, this just seems like a bug in slang when the type parameter is unresolved for e.g. a class that is never instantiated anywhere. VCS will not allow you to actually instantiate like this: `EventHandler #(Derived)`, it only works when `T` is actually `Base`.

scripts/warning_docs.txt
MikePopoloski: Yeah, it seems like all tools allow a local var to be unpacked, but not actually referenced directly in an assertion expression, you need to select from it.

docs/command-line-ref.dox
mampcs: The primary reason to keep them in separate syntax trees is that tools like VCS ignore any errors in library files code that is not used. To allow this capability in slang, we need to determine what library the code came from. And the library name is attached to the syntax tree. Will make this the default behavior.

scripts/diagnostics.txt
```diff
@@ -1017,6 +1020,7 @@ error HierarchicalFromPackage "cannot use hierarchical reference from within a p
 error CompilationUnitFromPackage "cannot reference compilation unit item from within a package"
 error InvalidHierarchicalIfacePortConn "cannot connect interface port hierarchically through a generate block or instance array"
 error UnnamedGenerateReference "cannot reference unnamed generate block via external name '{}' (use --allow-genblk-reference to allow this)"
+warning port-arg-redeclared PortArgRedeclared "local variable '{}' redeclares a port argument with the same name"
```
MikePopoloski, line 1023: You've already made redefinition suppressible so I don't see why you also need this new warning for the same thing.

source/ast/types/Type.cpp
```diff
@@ -736,6 +736,25 @@ bool Type::implements(const Type& ifaceClass) const {
     return false;
 }
 
+bool Type::isSameGenericClass(const Type& other) const {
```
MikePopoloski, line 739: Rather than a new function we have to remember to call everywhere you should just make isAssignmentCompatible return true for these cases.

tests/unittests/DriverTests.cpp
```diff
@@ -263,3 +267,732 @@ TEST_CASE("Driver customize default lib name") {
...
+        if (tree->isLibraryUnit)
+            ++libraryUnits;
+        else
+            ++defaultUnits;
+    }
+    CHECK(defaultUnits == 1);
+    CHECK(libraryUnits == 1);
+}
+
+TEST_CASE("Driver ClockVarTargetAssign -- error in strict mode") {
```
MikePopoloski, line 300: Most of these new driver tests that are just checking flag behavior should use the new slang-lit regression testing setup, less boilerplate involved.

source/ast/expressions/CallExpression.cpp
```diff
@@ -613,6 +613,13 @@ Expression& CallExpression::createSystemCall(
                             buffer.push_back(compilation.emplace<EmptyArgumentExpression>(
                                 compilation.getVoidType(), actualArgs[i]->sourceRange()));
                         }
+                        else if (subroutine.allowEmptyArgumentInCompatMode(index) &&
```
MikePopoloski, line 616: This whole empty arg thing seems convoluted. We don't need a whole separate virtual method that can only ever return true for sformatf (which is baked into the new compat flag name). The sformat method can check the compat flag and return true from the existing allowEmptyArgument method.

Alternatively, since format arg handling already has a dedicated error for empty arguments (which this PR deletes from FmtHelpers) you can instead have allowEmptyArgument always return true for sformat, make FormatEmptyArg a suppressible warning, and then not need a new compilation flag at all.

scripts/diagnostics.txt
```diff
@@ -896,6 +898,7 @@ error ConcurrentAssertNotInProc "concurrent assertion statements may only appear
 error ImmediateAssertNotInProc "immediate assertion statements may only appear in procedural contexts"
 error CheckerNotInProc "procedural checkers may only appear in always and initial blocks"
 error CheckerInForkJoin "checkers cannot be instantiated inside fork-join blocks"
+warning string-in-constraint StringConstraintExpr "'string' type is not valid in a constraint expression; string equality in constraint if-conditions is not defined by LRM §18.5"
```
MikePopoloski, line 901: I think "string' type is not valid in a constraint expression" is sufficient here, the second clause doesn't add enough information to be worth it.

scripts/diagnostics.txt
```diff
@@ -315,6 +315,7 @@ warning foreach-call-expr ForeachCallExpr "foreach array expression is not a nam
 warning lifetime-prototype LifetimeForPrototype "lifetime specifier is not allowed on prototype declaration"
 warning nonstandard-dist NonstandardDist "use of parentheses around 'dist' expression is non-standard"
 warning nonstandard-inside NonstandardInside "use of 'inside' without braces is non-standard"
+warning nonstandard-constraint-block NonstandardConstraintBlock "nested constraint block used as a constraint item is not permitted by the LRM"
```
MikePopoloski, line 318: Use the "is non-standard" instead of "is not permitted by the LRM" wording for consistency

MikePopoloski:
I separated these out into independent commits, cleaned them up, and landed them all to master.

PR #1796, opened 2026-04-22
Allow streaming operators as branches of conditional (ternary) expressions

likeamahoney:
Slang incorrectly rejected streaming operators (`{<<N{...}}, {>>N{...}}`) when used as branches of a conditional expression in an assignment context (commercial tools like VCS accept this case):

```verilog
wire c;
assign c = c ? {<<8{c & c}} : c;  // error: streaming operator can only be used in an assignment or bit-stream cast argument
```

Also, this type of RTL design can be found in open-source projects such as CVA6 (https://github.com/openhwgroup/cva6/blob/8e694e6460d017fb8a175336cd7dcb91d8eff682/vendor/pulp-platform/axi_riscv_atomics/src/axi_riscv_amos.sv#L918).

The LRM permits this - streaming operators are valid wherever a bitstream source expression is allowed, including as branches of a ternary that is itself in an assignment context.

Two issues in `ConditionalExpression::fromSyntax`:

1. Missing flag propagation: `ASTFlags::StreamingAllowed` was not forwarded to the true/false branch contexts, so the streaming concatenation parser rejected the expression before type-checking even began.
2. Void-type unification failure: `StreamingConcatenationExpression` intentionally carries a void type (its real type is only resolved when `convertAssignment` wraps it). The existing type-unification logic had no handling for void branches, causing a "bad conditional expression" error even after the flag was propagated.

Fix in `source/ast/expressions/OperatorExpressions.cpp`:

1. Propagate `ASTFlags::StreamingAllowed` to both branch contexts (mirrors the existing `AllowUnboundedLiteral` pattern).
2. Before type unification, detect streaming branches (void-typed `ExpressionKind::Streaming`) and substitute the assignment target type (or the other branch's type) as a stand-in. This lets the existing numeric/class type-unification logic compute a correct `resultType`.
3. After `resultType` is determined but before constructing `ConditionalExpression`, call `convertAssignment` on any streaming branches to produce the proper `ConversionExpression(StreamingConcat)` wrappers — exactly as a direct assignment would.

### DIFF
--- a/source/ast/expressions/ConversionExpression.cpp
+++ b/source/ast/expressions/ConversionExpression.cpp
@@ -178,11 +178,11 @@ Expression& Expression::convertAssignment(const ASTContext& context, const Type&
         }
 
         if (expr.kind == ExpressionKind::Streaming) {
+            // This path is reached only when streaming is used in a self-determined context
+            // (AllowSelfDeterminedStreamConcat mode) where no assignmentTarget was available
+            // to pre-convert in fromSyntax. Validate and wrap as a streaming conversion.
             if (Bitstream::canBeSource(type, expr.as<StreamingConcatenationExpression>(),
                                        assignmentRange, context)) {
-                // Add an implicit bit-stream casting otherwise types are not assignment compatible.
-                // The size rule is not identical to explicit bit-stream casting so a different
-                // ConversionKind is used.
                 result = comp.emplace<ConversionExpression>(type, ConversionKind::StreamingConcat,
                                                             *result, result->sourceRange);
                 selfDetermined(context, result);
@@ -296,7 +296,8 @@ Expression& ConversionExpression::fromSyntax(Compilation& comp, const CastExpres
             return badExpr(comp, nullptr);
         }
 
-        operand = &create(comp, *syntax.right, context, ASTFlags::StreamingAllowed, type);
+        operand = &create(comp, *syntax.right, context,
+                          ASTFlags::StreamingAllowed | ASTFlags::BitstreamCast, type);
         if (operand->bad())
             return badExpr(comp, nullptr);
     }
--- a/source/ast/expressions/ConversionExpression.cpp
+++ b/source/ast/expressions/ConversionExpression.cpp
@@ -287,8 +287,8 @@ Expression& ConversionExpression::fromSyntax(Compilation& comp, const CastExpres
         // For other operands (e.g. assignment patterns), assignmentTarget is needed for type
         // deduction. The cast syntax wraps the operand in ParenthesizedExpressionSyntax, so
         // check the inner expression kind.
-        const bool isStreamingOperand =
-            syntax.right->expression->kind == SyntaxKind::StreamingConcatenationExpression;
+        const bool isStreamingOperand = syntax.right->expression->kind ==
+                                        SyntaxKind::StreamingConcatenationExpression;
         operand = &create(comp, *syntax.right, context, ASTFlags::StreamingAllowed,
                           isStreamingOperand ? nullptr : type);
         if (operand->bad())
--- a/source/ast/expressions/ConversionExpression.cpp
+++ b/source/ast/expressions/ConversionExpression.cpp
@@ -177,20 +177,6 @@ Expression& Expression::convertAssignment(const ASTContext& context, const Type&
                                                       *result, nullptr, assignmentRange);
         }
 
-        if (expr.kind == ExpressionKind::Streaming) {
-            // This path is reached only when streaming is used in a self-determined context
-            // (AllowSelfDeterminedStreamConcat mode) where no assignmentTarget was available
-            // to pre-convert in fromSyntax. Validate and wrap as a streaming conversion.
-            if (Bitstream::canBeSource(type, expr.as<StreamingConcatenationExpression>(),
-                                       assignmentRange, context)) {
-                result = comp.emplace<ConversionExpression>(type, ConversionKind::StreamingConcat,
-                                                            *result, result->sourceRange);
-                selfDetermined(context, result);
-                return *result;
-            }
-            return badExpr(comp, &expr);
-        }
-
         if (expr.kind == ExpressionKind::ValueRange) {
             // Convert each side of the range and return that as a new range.
             auto convert = [&](Expression& expr) -> Expression& {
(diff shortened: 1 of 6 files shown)


MikePopoloski:
I'm pretty sure the LRM does *not* allow this. It says:
> A streaming_concatenation (as specified in Syntax 11-4) shall be used either as the target of an assignment,
or as the source of an assignment, or as the operand of a bit-stream cast, or as a stream_expression in another
streaming_concatenation.

and
> It shall be an error to use a streaming_concatenation as an operand in an expression without first casting it to
a bit-stream type.

That being said, it does seem that most tools (though not all) silently allow this anyway, so I suppose we should add support.

I don't love all the extra special casing this is building up. I wonder if we can reduce most of it by not making rvalue streaming concats have a void type, but instead pass through the `assignmentTarget` and let that be the target type. Then we don't need to have that logic in convertAssignment anymore.

likeamahoney:
> I'm pretty sure the LRM does _not_ allow this. It says:
> 
> > A streaming_concatenation (as specified in Syntax 11-4) shall be used either as the target of an assignment,
> > or as the source of an assignment, or as the operand of a bit-stream cast, or as a stream_expression in another
> > streaming_concatenation.
> 
> and
> 
> > It shall be an error to use a streaming_concatenation as an operand in an expression without first casting it to
> > a bit-stream type.
> 
> That being said, it does seem that most tools (though not all) silently allow this anyway, so I suppose we should add support.
> 
> I don't love all the extra special casing this is building up. I wonder if we can reduce most of it by not making rvalue streaming concats have a void type, but instead pass through the `assignmentTarget` and let that be the target type. Then we don't need to have that logic in convertAssignment anymore.

Thanks for the suggestion - I've reworked it along those lines.

What changed:

`StreamingConcatenationExpression::fromSyntax` now accepts an `assignmentTarget` parameter (forwarded via the existing create/assignmentTarget mechanism already used by `ConditionalExpression` and `ConcatenationExpression`). When the target is known at bind time, the streaming concat validates widths with `canBeSource` and wraps itself in a `ConversionExpression(StreamingConcat)` carrying the target type directly. This means streaming branches of a ternary naturally get the right type without any post-hoc fixup.

`ConditionalExpression::fromSyntax` no longer needs the leftIsStreaming/rightIsStreaming detection, the type substitution trick, or the `convertAssignment` calls on the branches - all of that is gone. The branches already arrive with the correct type.

One edge case: explicit bitstream casts. In that context `ConversionExpression::fromSyntax` passes type as assignmentTarget so that assignment patterns inside the cast can deduce their type. But if the streaming concat pre-converts using `canBeSource`, the outer explicit cast then sees an already-typed expression and emits a spurious UselessCast (and the semantics also differ - `isBitstreamCast` requires exact width match, canBeSource only requires target > source). Fixed by adding `ASTFlags::BitstreamCast` (set by `ConversionExpression::fromSyntax`) as a guard in `fromSyntax`: in cast context we skip pre-conversion and let the existing `isBitstreamCast` code path in `ConversionExpression::fromSyntax` handle it unchanged.

`convertAssignment` streaming block is kept but is now only reachable via the VCS compat path `AllowSelfDeterminedStreamConcat`) where no `assignmentTarget` was available.

source/ast/expressions/OperatorExpressions.cpp
```diff
@@ -1515,9 +1524,18 @@ Expression& ConditionalExpression::fromSyntax(Compilation& comp,
         diag << *lt << *rt;
         diag << left.sourceRange;
         diag << right.sourceRange;
-        return badExpr(comp, result);
+        return badExpr(comp,
```
MikePopoloski: You should revert all these changes which appear to be no-ops but just repeat the expression creation multiple times.

source/ast/expressions/OperatorExpressions.cpp
```diff
@@ -2391,24 +2409,35 @@ Expression& StreamingConcatenationExpression::fromSyntax(
...
 
     if (!context.flags.has(ASTFlags::StreamingAllowed)) {
-        // Cap the width so we don't overflow. The conversion will error for us
-        // since the target width will be less than the source width.
+        // VCS compat mode: the error about requiring an assignment context can be silenced,
+        // so produce a real result type by converting to a packed bit vector of bitstream width.
+        // Cap the width so we don't overflow; the conversion will error for width mismatch.
         auto width = std::min(bitstreamWidth, (uint64_t)SVInt::MAX_BITS);
         auto& type = comp.getType(bitwidth_t(width), IntegralFlags::FourState);
         return convertAssignment(context, type, result, result.sourceRange);
```
MikePopoloski: Maybe we can just set this new type as the `assignmentTarget` and fall through to the handling below? Then we can remove the special case in convertAssignment.

source/ast/expressions/ConversionExpression.cpp
```diff
@@ -296,7 +296,8 @@ Expression& ConversionExpression::fromSyntax(Compilation& comp, const CastExpres
             return badExpr(comp, nullptr);
         }
 
-        operand = &create(comp, *syntax.right, context, ASTFlags::StreamingAllowed, type);
+        operand = &create(comp, *syntax.right, context,
+                          ASTFlags::StreamingAllowed | ASTFlags::BitstreamCast, type);
```
MikePopoloski: It's a little weird to call this ASTFlags::BitstreamCast because it isn't necessarily a bitstream cast. All we know for sure at this point is that it's an explicit cast. Anyway, you mentioned in the notes that this is to avoid spurious UselessCast warnings. If that's true we can probably just special case the check for useless casts and avoid needing this new flag altogether. Actually, since the streaming concat already produces an appropriate ConversionExpression we can just check for that and return it directly.

source/ast/expressions/OperatorExpressions.cpp
likeamahoney: sorry for this - it was reverted

source/ast/expressions/OperatorExpressions.cpp
likeamahoney: yeah, good idea - done in last commit

source/ast/expressions/ConversionExpression.cpp
likeamahoney: flag was removed and logic was fixed according to your suggestion

PR #1801, opened 2026-04-23
Preserve bound control expressions and branch kind on generate symbols

hankhsu1996:
## Summary

After elaboration of if-generate, case-generate, and loop-generate constructs, slang binds the control expressions and resolves the associated genvar but then discards them. Downstream tools that want to reason about the original generate structure have to rebind from syntax and re-resolve the genvar. This PR retains that information directly on `GenerateBlockSymbol` and `GenerateBlockArraySymbol` so consumers can reuse values that were already computed during elaboration. Refs #1800.

## Design

The change is purely additive provenance — no behavior change, no new abstraction, no extra allocations. Every new field is populated from a value that `BlockSymbols.cpp` already constructs during elaboration and that is otherwise dropped on return.

### GenerateBlockSymbol

- `branchKind` — new enum (`IfTrue`, `IfFalse`, `CaseItem`, `CaseDefault`, `LoopIteration`, `IllegalUnconditional`) identifying which arm produced the block.
- `generateConstructSyntax` — direct pointer to the originating `IfGenerateSyntax`, `CaseGenerateSyntax`, or `LoopGenerateSyntax`. Kept as a raw `const SyntaxNode*` to match how every other held-syntax pointer in the AST (`Expression::syntax`, `Statement::syntax`, `ParameterSymbol::defaultValSyntax`, etc.) is expressed; the invariant is documented on the field and enforced by a `SLANG_ASSERT` at the assignment site. A typed wrapper/tagged union was considered and rejected as an idiom slang doesn't use elsewhere.
- `conditionExpression` — bound condition of the enclosing if- or case-generate.
- `caseItemExpressions` — bound case-item label expressions.

### GenerateBlockArraySymbol

- `initialExpression`, `stopExpression`, `iterExpression` — bound loop-control expressions.
- `genvar` — the canonical `GenvarSymbol`: for an inline genvar this is the fabricated symbol held as a member of the array; for an external genvar it is the resolved declaration.
- Intentionally no `generateConstructSyntax` on the array: `getSyntax()` already returns the `LoopGenerateSyntax` directly, so a duplicate field would only be surface symmetry. Per-iteration `GenerateBlockSymbol`s do carry the field, because their `getSyntax()` is the block body rather than the construct.

### Directly-nested conditional generate

Per LRM 27.5, inner conditional generates without `begin`/`end` continue to flatten into sibling blocks at the outer scope. Each flattened block now carries its own immediate `generateConstructSyntax`; outer constructs remain reachable via `SyntaxNode::parent` on that pointer, so no nesting information is lost.

### Serialization

`serializeTo` emits the new fields. `generateConstructSyntax` is written as a syntax-kind label (not as a serializable identity reference) — documented at the call site.

## Testing

C++ tests in `HierarchyTests.cpp` cover the flat case plus three nested scenarios: `if`-in-`if` with `begin`/`end`, directly-nested `if`-in-`if`, and `case`-in-`if`. The directly-nested test walks up through `SyntaxNode::parent` to verify that outer structure stays recoverable. Parallel pyslang tests exercise the same paths plus inline-genvar to validate the Python binding surface.

## AI disclosure

### DIFF
--- a/include/slang/ast/symbols/BlockSymbols.h
+++ b/include/slang/ast/symbols/BlockSymbols.h
@@ -16,6 +16,7 @@
 namespace slang::ast {
 
 class Expression;
+class VariableSymbol;
 
 /// Indicates which branch of a conditional or loop generate construct
 /// produced a given GenerateBlockSymbol.
@@ -195,19 +196,17 @@ public:
     bool valid = false;
     bool isUnnamed = false;
 
-    /// Bound initializer expression of the enclosing loop-generate.
+    /// Initial value assigned to loopVariable.
     const Expression* initialExpression = nullptr;
 
-    /// Bound stop expression of the enclosing loop-generate.
+    /// Bound stop expression of the loop-generate.
     const Expression* stopExpression = nullptr;
 
-    /// Bound iteration expression of the enclosing loop-generate.
+    /// Bound iteration expression of the loop-generate.
     const Expression* iterExpression = nullptr;
 
-    /// Canonical genvar symbol for this loop. For an inline genvar this is a
-    /// fabricated symbol held as a member of this array; for an external
-    /// genvar it resolves to the declaration in an enclosing scope.
-    const Symbol* genvar = nullptr;
+    /// Loop variable used by the bound stop and iteration expressions.
+    const VariableSymbol* loopVariable = nullptr;
 
     GenerateBlockArraySymbol(Compilation& compilation, std::string_view name, SourceLocation loc,
                              uint32_t constructIndex) :
--- a/include/slang/ast/symbols/BlockSymbols.h
+++ b/include/slang/ast/symbols/BlockSymbols.h
@@ -15,6 +15,15 @@
 
 namespace slang::ast {
 
+class Expression;
+
+/// Indicates which branch of a conditional or loop generate construct
+/// produced a given GenerateBlockSymbol.
+#define GENERATE_BRANCH_KIND(x) \
+    x(IfTrue) x(IfFalse) x(CaseItem) x(CaseDefault) x(LoopIteration) x(IllegalUnconditional)
+SLANG_ENUM(GenerateBranchKind, GENERATE_BRANCH_KIND)
+#undef GENERATE_BRANCH_KIND
+
 /// Represents a statement block, either sequential or parallel.
 class SLANG_EXPORT StatementBlockSymbol final : public Symbol, public Scope {
 public:
@@ -122,6 +131,20 @@ public:
     bool isUnnamed = false;
     const SVInt* arrayIndex = nullptr;
 
+    /// Indicates which branch of the originating generate construct produced
+    /// this block.
+    GenerateBranchKind branchKind = GenerateBranchKind::IllegalUnconditional;
+
+    /// Originating generate construct; always an IfGenerateSyntax,
+    /// CaseGenerateSyntax, or LoopGenerateSyntax when non-null.
+    const syntax::SyntaxNode* generateConstructSyntax = nullptr;
+
+    /// Bound condition of the enclosing if- or case-generate, if any.
+    const Expression* conditionExpression = nullptr;
+
+    /// Bound case-item label expressions for case-generate blocks.
+    std::span<const Expression* const> caseItemExpressions;
+
     GenerateBlockSymbol(Compilation& compilation, std::string_view name, SourceLocation loc,
                         uint32_t constructIndex, bool isUninstantiated) :
         Symbol(SymbolKind::GenerateBlock, name, loc), Scope(compilation, this),
@@ -154,6 +177,20 @@ public:
     uint32_t constructIndex;
     bool valid = false;
     bool isUnnamed = false;
+
+    /// Bound initializer expression of the enclosing loop-generate.
+    const Expression* initialExpression = nullptr;
+
+    /// Bound stop expression of the enclosing loop-generate.
+    const Expression* stopExpression = nullptr;
+
+    /// Bound iteration expression of the enclosing loop-generate.
+    const Expression* iterExpression = nullptr;
+
+    /// Canonical genvar symbol for this loop. For an inline genvar this is a
+    /// fabricated symbol held as a member of this array; for an external
+    /// genvar it resolves to the declaration in an enclosing scope.
+    const Symbol* genvar = nullptr;
 
     GenerateBlockArraySymbol(Compilation& compilation, std::string_view name, SourceLocation loc,
                              uint32_t constructIndex) :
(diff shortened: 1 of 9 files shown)


include/slang/ast/symbols/BlockSymbols.h
```diff
@@ -122,6 +131,20 @@ class SLANG_EXPORT GenerateBlockSymbol final : public Symbol, public Scope {
     bool isUnnamed = false;
     const SVInt* arrayIndex = nullptr;
 
+    /// Indicates which branch of the originating generate construct produced
+    /// this block.
+    GenerateBranchKind branchKind = GenerateBranchKind::IllegalUnconditional;
```
MikePopoloski, line 133: Move this up above arrayIndex to take advantage of the padding there. This can be a uint8_t enum so that all of the non-pointer members fit together into 8 bytes.

include/slang/ast/symbols/BlockSymbols.h
```diff
@@ -122,6 +131,20 @@ class SLANG_EXPORT GenerateBlockSymbol final : public Symbol, public Scope {
...
+
+    /// Originating generate construct; always an IfGenerateSyntax,
+    /// CaseGenerateSyntax, or LoopGenerateSyntax when non-null.
+    const syntax::SyntaxNode* generateConstructSyntax = nullptr;
```
MikePopoloski: I think you should be able to get this by calling getSyntax() on this symbol and then walking up the parent chain of the syntax nodes?

source/ast/symbols/BlockSymbols.cpp
```diff
@@ -683,6 +714,18 @@ std::string GenerateBlockSymbol::getExternalName() const {
 void GenerateBlockSymbol::serializeTo(ASTSerializer& serializer) const {
     serializer.write("constructIndex", constructIndex);
     serializer.write("isUninstantiated", isUninstantiated);
+    serializer.write("branchKind", toString(branchKind));
+    // Emitted as a syntax-kind label only; not a serializable identity reference.
+    if (generateConstructSyntax)
+        serializer.write("generateConstructSyntax", toString(generateConstructSyntax->kind));
```
MikePopoloski: Why do you need this if you have the branchKind already?

include/slang/ast/symbols/BlockSymbols.h
```diff
@@ -122,6 +131,20 @@ class SLANG_EXPORT GenerateBlockSymbol final : public Symbol, public Scope {
...
+
+    /// Bound condition of the enclosing if- or case-generate, if any.
+    const Expression* conditionExpression = nullptr;
```
MikePopoloski: This can be unioned with the arrayIndex member, with an accessor method to gate access based on the branchKind

MikePopoloski (commented):
Looks ok to me, but I'm generally worried about AST memory consumption and there can be quite a lot of generate blocks in a large design so I have some suggestions to reduce the impact.

include/slang/ast/symbols/BlockSymbols.h
hankhsu1996: Makes sense for `conditionExpression` vs `arrayIndex`, I will go ahead and do that.

One thing I was thinking about though is the case-generate side. Each block also carries a caseItemExpressions span, which is already shared backing storage but still costs a pointer plus size per block. So even if I fold `conditionExpression` into the union, the per-block footprint for case branches does not really change much.

The only way I can see to reduce that further would be to move the conditional or case payload out to a shared object at the generate-construct or per-case-item level, and have each block just reference it. But that would introduce a new abstraction and make this change quite a bit less local.

Happy to keep this simple and just follow your suggestion with the union, just wanted to sanity check whether you would prefer going further in that direction or not.

MikePopoloski: Yeah, I thought about it as well and concluded that it's easier to just eat the extra two pointers for now. We want to reduce memory usage but not at the expense of a huge amount of complexity.

hankhsu1996:
One more thing I ran into for loop-generate:

`GenerateBlockArraySymbol::genvar` gives the semantic genvar, but inside `stopExpression` / `iterExpression` the `i` actually resolves to a compiler-generated `VariableSymbol`, not the `GenvarSymbol`.

So from the consumer side these look like two unrelated symbols, even though they represent the same source-level variable.

Would it make sense to also expose that expression-binding symbol (e.g. `genvarExpressionSymbol`) on `GenerateBlockArraySymbol`, so consumers don’t have to rely on name/flag heuristics to connect the two?

MikePopoloski:
I guess it depends what the purpose of linking to the genvar is. genvars don't really do anything, they're kind of syntactic noise, so I'm not sure what you need the link for.

The rest of the diff looks ok to me though.

hankhsu1996:
I think what I care about here is more about preserving the source-level identity than the exact behavior of genvar.

From a consumer's point of view, the `i` inside expressions like `i < 4` ends up being a compiler-generated variable, which feels more like an slang internal artifact than something from the SV language itself. The original genvar is still there, but the connection between the two is not really visible, so the SV source-level intent gets a bit lost.

A more practical issue: If I wanted to do something like LSP go-to-definition on the `i` in the loop condition or iteration expression, I would expect it to point back / jump to the genvar declaration. Right now there is not a straightforward way to do that, since there is no explicit link between the expression-side symbol and the genvar.

MikePopoloski:
The simple fix might be to set the SyntaxNode pointer on the generated variable to point to the genvar syntax, like we already do for the per-block created localparam. A more involved fix is to make GenvarSymbols derive from VariableSymbol and use them directly in the expression evaluation, but I'm not sure how useful that additional work would be.

hankhsu1996:
I think your syntax fix is useful, but the thing I still need is a stable symbol identity for the loop variable.

What I am building downstream is trying to keep the generate loop around instead of only looking at the fully elaborated blocks. For that to work cleanly, I need the loop variable to exist somewhere on the generate-loop symbol, so I can treat it as the declaration and then have the control expressions refer back to the same thing.

So maybe instead of exposing the `GenvarSymbol`, `GenerateBlockArraySymbol` could expose the generated `VariableSymbol` that `initialExpression` / `stopExpression` / `iterExpression` already refer to, maybe as something like `loopVariable`.

That would give downstream tools one symbol to treat as the loop variable declaration, and the expressions already point at that same symbol. If that symbol's syntax can also point back to the source genvar declaration, that would cover the navigation side too.

Does that sound like a reasonable direction?

MikePopoloski:
Yes, that seems ok to me.

hankhsu1996:
Updated the PR based on the discussion:

* Replaced `genvar` with `loopVariable`, so the loop variable exposed by `GenerateBlockArraySymbol` is the same symbol referenced by the bound stop / iteration expressions.
* Set the generated loop variable's syntax from the source genvar syntax, so source provenance / navigation still points back to the genvar.
* Updated serialization, pyslang bindings, and tests accordingly.

hankhsu1996:
Pushed a small follow-up for the latest `SyntaxNode::parent` change on master:

* Updated the affected C++ test code to use explicit `const SyntaxNode*`.
* Updated the pyslang `parent` binding to keep exposing `SyntaxNode | None` instead of the internal `ParentRef` wrapper.

PR #1802, opened 2026-04-25
Fix modport reference tracking for unused analysis

### LINKED ISSUE #1798: Unused output / inout modport can suppress `-Wundriven-net` on interface signals
C++ slang

**Describe the bug**
It looks like declaring an unused `output` / `inout` modport on an interface signal can incorrectly suppress `-Wundriven-net`.

In the example below, the signal is only ever read and never driven anywhere, so I would expect an undriven warning. That works as expected when only a reader modport exists. However, if I add a second modport with `output` direction (and never actually use it), the warning disappears.

From a quick look into the code, it seems like the underlying signal is being marked as “written” just because of the modport declaration, even though there’s no actual write in the design.

**To Reproduce**

Works as expected (warning is reported):

```systemverilog
interface status_if;
    wire status;
    modport reader (input status);
endinterface

module status_consumer(
    status_if.reader intf,
    output wire observed_status
);
    assign observed_status = intf.status;
endmodule

module top;
    status_if intf();
    wire observed_status;

    status_consumer u_consumer(
        .intf(intf),
        .observed_status(observed_status)
    );

    // Expected and actual:
    // warning: status is never driven (-Wundriven-net)
endmodule
```

Adding an unused writer modport suppresses the warning:

```systemverilog
interface status_if;
    wire status;
    modport reader (input status);
    modport writer (output status); // never used
endinterface

module status_consumer(
    status_if.reader intf,
    output wire observed_status
);
    assign observed_status = intf.status;
endmodule

module top;
    status_if intf();
    wire observed_status;

    status_consumer u_consumer(
        .intf(intf),
        .observed_status(observed_status)
    );

    // Expected:
    // warning: status is never driven (-Wundriven-net)
    //
    // Actual:
    // no warning
endmodule
```

**Additional context**
I did a bit of local investigation and it seems related to how modport ports are tracked as references. Specifically, `output` / `inout` modports appear to mark the underlying signal as being written (lvalue-referenced) at declaration time, even if they’re never actually used.

That would explain why adding an unused writer modport flips the signal into a “written somewhere” state and suppresses the undriven check.

Happy to dig further or try a fix if this direction sounds right.

hankhsu1996:
## Summary

Fixes #1798. Declaring an `output` or `inout` modport currently silences `-Wundriven-net` / `-Wunused-net` on the underlying interface signal, even when the modport is never instantiated. Minimal repro:

```sv
interface status_if;
    wire status;
    modport reader(input status);
    modport writer(output status);
endinterface

module status_consumer(status_if.reader intf, output wire observed_status);
    assign observed_status = intf.status;
endmodule
```

`writer` is never used, so `status` is never driven — but no diagnostic was issued.

## Design

`ModportPortSymbol::fromSyntax` bound the connection expression under an `LValue`-tagged context, so a single call did two unrelated jobs: direction/lvalue validation *and* recording a read/write on the underlying signal. The mere existence of a modport declaration was effectively counted as a use of every signal it named.

Modport ports are aliases for interface members; references should be recorded only at actual use sites and forwarded to the connected signal(s). The change separates the two responsibilities.

### Declaration-time binding

Both `ModportPortSymbol::fromSyntax` paths (named-port and explicit `.alias()`) now `resetFlags(NonProcedural | NoReference)`. Validation (direction checks, lvalue checks, type linking) still runs against the same checking context; only the side effect of marking the signal as referenced is suppressed. This mirrors the established `NoReference` pattern at `PortSymbols.cpp:1349,1386` for module-port internals.

### Use-site forwarding

`ASTContext::noteReference` now treats `ModportPortSymbol` as an alias and forwards the recorded reference to its underlying value symbol(s):

- `internalSymbol` — followed directly for the named-port form.
- `explicitConnection` — walked with `visitSymbolReferences`, restricted to value symbols, for the `.alias(expr)` form. Mirrors the pattern at `PortSymbols.cpp:1410`.

The shared `LAndRValue` re-mark is factored into a single local `mark` lambda so the forwarded references and the original facade reference share one source of truth for read/write/inout semantics.

### Behavior change

Code that previously relied on declaration-time marking will now correctly emit `-Wunused-net` / `-Wundriven-net`. This is intentional and is the original bug.

## Testing

- New regression test in `UnusedTests.cpp` covering #1798: an unused `writer` modport no longer suppresses `-Wundriven-net` on the underlying net.
- New coverage for the explicit form: `output .alias(a)` no longer masks `-Wunused-variable`.
- Two existing tests (`Inout ports are treated as readers and writers`, `Unused nets and vars false positives regress`) updated to exercise their modports through real users; they previously relied on declaration-time marking.

Full test suite passes (2281 / 2281).

## AI disclosure

### DIFF
--- a/source/ast/ASTContext.cpp
+++ b/source/ast/ASTContext.cpp
@@ -15,6 +15,7 @@
 #include "slang/ast/symbols/BlockSymbols.h"
 #include "slang/ast/symbols/CheckerSymbols.h"
 #include "slang/ast/symbols/InstanceSymbols.h"
+#include "slang/ast/symbols/MemberSymbols.h"
 #include "slang/ast/symbols/PortSymbols.h"
 #include "slang/ast/symbols/VariableSymbols.h"
 #include "slang/ast/types/Type.h"
@@ -417,10 +418,31 @@ void ASTContext::noteReference(const ValueSymbol& symbol, bool isDottedAccess) c
         }
 
         auto& comp = getCompilation();
-        comp.noteReference(*syntax, isLValue);
+        auto mark = [&](const syntax::SyntaxNode& node) {
+            comp.noteReference(node, isLValue);
+            if (isLValue && flags.has(ASTFlags::LAndRValue))
+                comp.noteReference(node, /* isLValue */ false);
+        };
 
-        if (isLValue && flags.has(ASTFlags::LAndRValue))
-            comp.noteReference(*syntax, /* isLValue */ false);
+        mark(*syntax);
+
+        // Modport ports are aliases for interface members; references to the
+        // facade should also count as references to the connected value(s).
+        if (auto mpp = symbol.as_if<ModportPortSymbol>()) {
+            if (mpp->internalSymbol) {
+                if (auto underlying = mpp->internalSymbol->getSyntax())
+                    mark(*underlying);
+            }
+            else if (mpp->explicitConnection) {
+                mpp->explicitConnection->visitSymbolReferences(
+                    [&](const Expression&, const Symbol& s) {
+                        if (!s.isValue())
+                            return;
+                        if (auto underlying = s.getSyntax())
+                            mark(*underlying);
+                    });
+            }
+        }
     }
 }
 
--- a/source/ast/ASTContext.cpp
+++ b/source/ast/ASTContext.cpp
@@ -418,28 +418,23 @@ void ASTContext::noteReference(const ValueSymbol& symbol, bool isDottedAccess) c
         }
 
         auto& comp = getCompilation();
-        auto mark = [&](const syntax::SyntaxNode& node) {
-            comp.noteReference(node, isLValue);
-            if (isLValue && flags.has(ASTFlags::LAndRValue))
-                comp.noteReference(node, /* isLValue */ false);
-        };
+        comp.noteReference(*syntax, isLValue);
 
-        mark(*syntax);
+        if (isLValue && flags.has(ASTFlags::LAndRValue))
+            comp.noteReference(*syntax, /* isLValue */ false);
 
         // Modport ports are aliases for interface members; references to the
-        // facade should also count as references to the connected value(s).
+        // facade also count as references to the connected value(s).
         if (auto mpp = symbol.as_if<ModportPortSymbol>()) {
-            if (mpp->internalSymbol) {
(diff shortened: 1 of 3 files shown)


tests/unittests/analysis/UnusedTests.cpp
```diff
@@ -27,22 +27,18 @@ Diagnostics analyze(const std::string& text, Compilation& compilation) {
 TEST_CASE("Inout ports are treated as readers and writers") {
     auto& text = R"(
 interface I;
-    wire integer i;
-    modport m(inout i);
+    wire integer a;
+    modport m(inout a);
 endinterface
 
-module m(inout wire a);
```
MikePopoloski, line 34: Why did you delete this part of the test?

source/ast/ASTContext.cpp
```diff
@@ -417,10 +418,31 @@ void ASTContext::noteReference(const ValueSymbol& symbol, bool isDottedAccess) c
...
+            if (mpp->internalSymbol) {
+                if (auto underlying = mpp->internalSymbol->getSyntax())
+                    mark(*underlying);
+            }
+            else if (mpp->explicitConnection) {
+                mpp->explicitConnection->visitSymbolReferences(
+                    [&](const Expression&, const Symbol& s) {
+                        if (!s.isValue())
+                            return;
+                        if (auto underlying = s.getSyntax())
```
MikePopoloski: You should probably recursively call noteReference instead of using this mark() lambda, in case you have a modport port that refers to another modport port (perhaps in a parent scope).

tests/unittests/analysis/UnusedTests.cpp
hankhsu1996: Good catch. The original test did need a small update because this fix makes the unused interface/modport declaration start warning correctly, but I initially changed too much.

I restored the original `module m(inout wire a)` / `tranif1` part of the test and only removed the dead interface/modport block that depended on the old declaration-time marking behavior.

hankhsu1996:
* Restored the original inout-module-port test and only removed the unused interface/modport part that was relying on the previous behavior.
* Switched the forwarding logic to recurse through `noteReference` as suggested.
* Added focused tests for:
  * the original undriven case with an unused writer modport
  * explicit modport outputs not masking unused variables
  * the happy-path where reads/writes through a modport correctly propagate

PR #1806, opened 2026-04-28
Allow upcasts through generic class specializations

thomasnormal:
This fixes class-handle assignment compatibility when a base class type is reached through a parameterized class specialization.

Example:

```systemverilog
class cb_base #(type T = int);
endclass

class ev_cls #(type T = int);
    typedef cb_base#(T) cb_type;
    function void add_callback(cb_type cb);
    endfunction
    function void add_direct(cb_base#(T) cb);
    endfunction
endclass

class veto_cb extends cb_base#(int);
endclass

module top;
    initial begin
        ev_cls#(int) ev;
        veto_cb cb;
        ev = new;
        cb = new;
        ev.add_callback(cb);
        ev.add_direct(cb);
    end
endmodule
```

With both Slang 10.0.269 and 10.0.35 I get
```
  test.sv:21:25: error: value of type 'veto_cb' cannot be assigned to type 'cb_type' (aka 'cb_base#(T (aka 'int'))')
          ev.add_callback(cb);
  test.sv:22:23: error: value of type 'veto_cb' cannot be assigned to type 'cb_base#(T (aka 'int'))'
          ev.add_direct(cb);
```

After specialization `ev_cls#(int)`, the formal type is `cb_base#(int)`, and `veto_cb` extends `cb_base#(int)`, so by normal SV class polymorphism (LRM 8.14) the derived-to-base assignment should be allowed.
With Xcelium 24.03 this compiles and simulates cleanly.

Currently this is rejected because `Type::isDerivedFrom` walks to a base class specialization and compares it to the destination base type by pointer identity. In this case both types represent the same `cb_base#(int)` specialization, but can be distinct `ClassType` objects.

The fix reuses the existing `ParameterSymbolBase::allMatching` logic when both class types come from the same generic class definition.

Validation:

* Added an AST unit test covering both the typedef and direct `base#(T)` formal forms.
* Verified the reproducer is accepted by Xcelium 24.03.
* In the CIRCT integration checkout, applying this change to the vendored slang source and rebuilding `circt-verilog` makes the reduced repro and CIRCT regression pass.

AI disclosure:
I reviewed every line and am fully accountable for the change.


source/ast/types/Type.cpp
```diff
@@ -703,8 +703,18 @@ bool Type::isDerivedFrom(const Type& base) const {
...
+        if (!d)
+            break;
+
+        d = &d->getCanonicalType();
         if (d == b)
             return true;
+
+        if (d->isClass() && b->isClass() && d->as<ClassType>().genericClass &&
+            d->as<ClassType>().genericClass == b->as<ClassType>().genericClass &&
+            ParameterSymbolBase::allMatching(d->as<ClassType>().genericParameters,
```
MikePopoloski, line 715: It should be the case that all specializations with the same set of parameter values is already the same symbol instance, since the generic class maintains a hash table keyed by parameter values. The actual problem here is that this mismatches for the "uninstantiated" specialization, where all param values are given an unset / error value. I think the fix can be more targeted. It's also related to a portion of this other open PR: https://github.com/MikePopoloski/slang/pull/1791/changes#diff-f830e3f0413a6a6b563b0baee320db9db9ff12e5ee605ea73fcd7b9a6ba18298R751

thomasnormal: Should I try to generalize it, or is it covered by the other PR?

MikePopoloski: I included a fix for this in b33486b2b79e36a3ec409e0115699122c0817599

PR #1812, opened 2026-05-05
ARM64 _umul128 fallback via __umulh

toddstrader:
Needed for Windows + arm build, which mason wants for slang-server:
https://github.com/mason-org/mason-registry

Let me know if you want me to add win + arm to the slang build matrix.


MikePopoloski:
Yeah, if we want it to keep working we need a build for it.

PR #1820, opened 2026-05-10
Add CompilationFlags::IgnoreUninstantiatedModules

ayanbanrj:
This adds a new compilation flag that suppresses the automatic instantiation of unreferenced modules during compilation.

EDA tools generally completely skip compilation of unreachable modules when a top module is specified by the user. This flag provides an opt-out for tools that need such behavior.

### DIFF
--- a/include/slang/ast/Compilation.h
+++ b/include/slang/ast/Compilation.h
@@ -127,9 +127,12 @@ enum class SLANG_EXPORT CompilationFlags {
     /// Allow multiple definitions of the same module, interface, program, or primitive at
     /// the root scope within the same library, keeping the first and silently discarding
     /// subsequent ones, but only when the conflicting definition comes from a library file.
-    AllowLibModuleRedefinition = 1 << 16
+    AllowLibModuleRedefinition = 1 << 16,
+
+    /// Don't instantiate unreferenced modules to perform semantic checking on them.
+    IgnoreUninstantiatedModules = 1 << 17
 };
-SLANG_BITMASK(CompilationFlags, AllowLibModuleRedefinition)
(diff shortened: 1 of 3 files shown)


MikePopoloski:
What is the use case for this? All commercial tools I've tried will report an error for undeclared identifiers, even in uninstantiated modules. It's hard for me to imagine why you want to provide invalid code to the tool and have it ignore it.

ayanbanrj:
RTL static analysis tools such as JasperGold and AscentLint have this behavior, and some users expect tools not to perform elaboration on any module that is not reachable from the specified top module, even if the module is included in the file list.

MikePopoloski:
slang is not doing elaboration on these modules, it's just doing some basic correctness checking. I guess the thing I don't understand is why you have invalid code at all? It can't do anything useful in any tool, why not fix it? And if you don't want to fix it, why are you providing it as input to the tool instead of just... not including it?

But anyway, you probably want this hooked up to a Driver command line flag too? Or are you just going to use it via API? Also note that you can already achieve this same behavior by marking these files as library files (e.g. with `-v` or with library maps).

ayanbanrj:
> Or are you just going to use it via API?

Yes I am using it via the API, so it's not hooked to the Driver.

> It can't do anything useful in any tool, why not fix it?

The intent is to not break users' existing flows, which might have large filelists that include invalid but unused modules.

I agree that the proper way is to fix the input filelists, and understand if you'd prefer not to merge this.

MikePopoloski:
This is fine, it's not going to be hard to maintain going forward. Just want to make sure we aren't adding flags for no reason.

PR #1822, opened 2026-05-12
Feat/add waiver file support

sjalloq:
This PR adds waiver file support as originally mentioned in #1586.  Since that discussion I've re-written the waiver syntax which was overly verbose and it seems to work well now.  I've been using this day to day for the last 6 months so it's pretty well tested.  [**NOTE**: YAML -> TOML]

I wanted to open this PR to restart the discussion and review the code.  There are likely changes you'd want to make either to code style or UX design.

The waiver syntax has reduced to a file glob, a list of diagnostic messages to ignore, and an optional regex pattern to further limit the waiver:

```toml
# Waive width truncation in a specific instance
[[waivers]]
hier = "top/u_subsys/u_conv"
diagnostic = "width-trunc"

# Same rule using dot separators
[[waivers]]
hier = "top.u_subsys.u_conv"
diagnostic = "width-trunc"

# Waive in a hierarchy subtree using globs
[[waivers]]
hier = "**/u_debug*"
diagnostic = "unused-variable"

# Combine with regex to match specific line content
[[waivers]]
hier = "**/u_debug*"
diagnostic = "unused-variable"
regex = '\bdbg_status\b'
```


MikePopoloski:
Thanks for the PR. The general idea seems good to me. How tied are you to YAML? It's such a complicated format for something that is a pretty simple use; when thinking about this in the past I was leaning towards TOML instead.

sjalloq:
TOML is fine too.  Any particular library you'd want to use?

I had a quick look and `toml++` looks like it will work with `-fno-exceptions` so I'll start there unless you have a preference.

MikePopoloski:
Yeah, toml++ seems decent.

sjalloq:
Made the changes and will test it out tomorrow.  I'll port my existing waivers to TOML and check everything runs again.

If you could check that the no-exception handling fits with the rest of the project that would be good.

sjalloq:
Seems to be working fine with TOML waivers so over to you for review.

docs/waivers.md
```diff
@@ -0,0 +1,156 @@
+# Diagnostic Waivers
```
MikePopoloski: This should be written in the .dox format so it can be included in the documentation on the website.

docs/waivers.md
```diff
@@ -0,0 +1,156 @@
...
+regex = '\bdbg_status\b'
+```
+
+**Limitation: not every diagnostic carries a hierarchical path.** Hier-scope rules only match diagnostics that are tied to an AST symbol with a parent scope. Preprocessor warnings (e.g. `include-not-found`), parser-level warnings, and some structural warnings have no symbol and are silently passed through by hier rules — use `file` scope for those.
+
+If a hier rule names a diagnostic that was issued but had no symbol, the unused-waivers summary flags it specifically (`diagnostic 'X' has no symbol; hier scope cannot match — use 'file' scope`) so you'll see it instead of the generic "no hier paths matched" message.
+
+## Glob Patterns
+
+Both `file` and `hier` fields support glob patterns:
```
MikePopoloski: This should probably also support `...` for consistency with other globs used in slang. Note that `...` is the form specified in the LRM, which is why slang uses it. I'm open to also allowing `**` interchangeably everywhere for consistency, since that's the form most people expect to use.

docs/waivers.md
```diff
@@ -0,0 +1,156 @@
...
+```
+
+Common diagnostic names:
+- `unused-variable`
+- `unused-implicit-net`
+- `width-trunc`
+- `empty-member`
+- `implicit-net-port`
+
+Use `slang --help-diagnostics` to see a list of available diagnostic names.
```
MikePopoloski: This command line flag does not exist, looks like the LLM made it up.

docs/waivers.md
```diff
@@ -0,0 +1,156 @@
...
+file = "**/rtl/**"
+diagnostic = "unused-variable"
+
+# Multiple diagnostics
+[[waivers]]
+file = "**/rtl/**"
+diagnostic = ["unused-variable", "width-trunc"]
+```
+
+Common diagnostic names:
```
MikePopoloski: You can just link to the warning docs here instead of picking this set.

source/CMakeLists.txt
```diff
@@ -136,6 +137,21 @@ set_target_properties(
 target_compile_options(slang_slang PRIVATE ${SLANG_WARN_FLAGS})
 target_compile_definitions(slang_slang PUBLIC $<$<CONFIG:Debug>:SLANG_DEBUG>)
 
+# Build toml++ in result-style mode so WaiverManager.cpp compiles in both
+# default and -fno-exceptions presets. Scoped to the only TU that includes
+# toml.h so the rest of slang_slang doesn't pay the redefine cost.
+set_source_files_properties(diagnostics/WaiverManager.cpp
+                            PROPERTIES COMPILE_DEFINITIONS TOML_EXCEPTIONS=0)
```
MikePopoloski: I would just define this in WaiverManager.cpp before the include so it's not hidden in the cmake.

source/CMakeLists.txt
```diff
@@ -136,6 +137,21 @@ set_target_properties(
...
+
+# GCC's libstdc++ <regex> trips -Walloc-zero in _Executor::_M_lookahead (new
+# bool[__n]() with __n possibly zero). The diagnostic fires in a system header
+# during template instantiation, so source-level pragmas do not suppress it;
+# disable the warning for the only TU that uses std::regex_search.
+if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+  set_source_files_properties(diagnostics/WaiverManager.cpp APPEND
```
MikePopoloski: Same with this; see other uses of `pragma GCC` in the library.

include/slang/diagnostics/DiagnosticEngine.h
```diff
@@ -172,6 +173,13 @@ class SLANG_EXPORT DiagnosticEngine {
     /// to an error.
     std::error_code addIgnoreMacroPaths(std::string_view pattern);
 
+    /// Sets the waiver manager to use for suppressing diagnostics based on external
+    /// waiver file rules. The waiver manager should be loaded from a TOML file before
```
MikePopoloski: I think this qualifier is not true; this just stores the pointer, so the loading from file can happen after this call.

source/driver/Driver.cpp
```diff
@@ -1316,6 +1337,13 @@ bool Driver::reportDiagnostics(bool quiet) {
                               diagEngine.getNumWarnings() == 1 ? "" : "s"));
     }
 
+    if (auto waiverManager = diagEngine.getWaiverManager()) {
```
MikePopoloski, line 1340: Should probably suppress the summary if --quiet is passed and print-unused isn't provided?

source/diagnostics/WaiverManager.cpp
```diff
@@ -0,0 +1,493 @@
...
+            OS::printE(fmt::format("[waiver] applied rule {} (scope={})\n", idx,
+                                   rule.hierScope ? "hier" : "file"));
+        }
+        return true;
+    }
+
+    return false;
+}
+
+std::string WaiverManager::getLineText(SourceLocation location,
```
MikePopoloski: This looks like a duplicate of the same function in DiagnosticClient::getSourceLine, should find a way to reuse that instead.

sjalloq: Do you want to move getSourceLine to a helper or just make it non-protected?

docs/waivers.md
sjalloq: Fixed.

docs/waivers.md
sjalloq: Added support.

docs/waivers.md
sjalloq: Good catch. Removed.

docs/waivers.md
sjalloq: Fixed.

source/CMakeLists.txt
sjalloq: Fixed.

source/CMakeLists.txt
sjalloq: Fixed.

include/slang/diagnostics/DiagnosticEngine.h
sjalloq: Fixed.

source/driver/Driver.cpp
sjalloq: Fixed.

source/CMakeLists.txt
sjalloq: Looks like pragmas don't work for this when optimisations are turned on.  Any thoughts on how to fix this?

sjalloq: I went round the loop here and did a bit more reading into the issue.  The bug is an optimisation pass issue so pragmas around function calls don't work.  I had to drop the optimisation level for WaiverManger.cpp to get pragmas to work.  If that's not acceptable then we might have to revert to the original set_source_files_properties.

source/diagnostics/WaiverManager.cpp
sjalloq: I've made a stab at tidying this up by moving getSourceLine into SourceManager.  We can iterate again if that's not what you had in mind.

source/CMakeLists.txt
MikePopoloski: Hmm ok, the original usage is fine then. I'll probably look at pulling in a better regex library anyway, since std::regex is known to be terrible.

sjalloq: OK, will revert that and let you pick a better library that you're happy with.

sjalloq:
I think this is ready to be squashed and merged unless my changes to `SourceManager` raise concerns.  Let me know if there's anything else to fix and thanks for the review.

Issue #1819, opened 2026-05-08
Python callback to report on parsing process


tomverbeure:
**Is your feature request related to the C++ slang project or the pyslang Python bindings?**

Pyslang Python bindings

**Is your feature request related to a problem? Please describe.**

I'm trying to load a huge design with a filelist with driver.parseCommandLine, driver.processOptions and driver.parseAllSources. The filelist itself is ~700 lines, a large part of which are -y links to standard cell libraries and implementation files with millions of instances.
After calling parseAllSources, pyslang goes to work, but there is no feedback about what's actually going on. I have no clue how it will take to parse the whole design. 

**Describe the solution you'd like**

Some kind of callback function that can be used to report parsing progress. 
E.g. when a new module or file has been started.
I'm not familiar with the C++ library, but it looks like there's alread something like slang::parsing::PreprocessorOptions::bufferChangeCB. I don't know if that's the ticket. If it is, can it be exposed to the Python library?

**Describe alternatives you've considered**

Write everything in C++, which I really don't want to do...

MikePopoloski:
Yes, we can expose that callback, which probably does what you want. You can also just run the main slang C++ binary on your codebase to get an idea for how long the parsing will take.

tomverbeure:
I tried to vibe code a temporary solution for personal use. It works well, the code looks clean and contained, but I don’t understand all the details. If you’re not allergic to these kind of submissions, I can create a PR, but I understand if you don’t want to go anywhere near it.

MikePopoloski:
It should be like two lines added to the bindings and then maybe a test Python script showing it works? If so then sure, it's straightforward enough for an LLM to do.

tomverbeure:
The solution is more complicated that you described it. I think it's to still allow concurrency?

MikePopoloski:
Added in 308708ec15d9f715b145869283380261c12167bc

PR #1823, opened 2026-05-13
Python buffer change cb

tomverbeure:
Fix for issue #1819.


MikePopoloski:
Hmm yeah, I can see why it did this but it's not right to duplicate all the logic into the bindings. Instead we should just refactor the parseAllSources method a bit to allow providing the callback directly. I can do it myself soonish.

tomverbeure:
I figured. :-) It’s a nice to have, but there’s no rush.

MikePopoloski:
Superseded by 308708ec15d9f715b145869283380261c12167bc

PR #1827, opened 2026-05-13
Classes: expose paramDecls

AndrewNolte:
Stacked PRs:
 * #1831
 * #1829
 * #1839
 * #1838
 * #1837
 * #1836
 * #1835
 * #1834
 * #1833
 * #1832
 * #1830
 * #1828
 * __->__#1827

--- --- ---

[View individual changes](https://github.com/AndrewNolte/slang/commit/d9cde94c7f548359582cf26b3397862df171cded)

### DIFF
--- a/include/slang/ast/symbols/ClassSymbols.h
+++ b/include/slang/ast/symbols/ClassSymbols.h
@@ -278,6 +278,11 @@ public:
     /// Behavior is undefined if this generic class has already been instantiated and used.
     void addParameterDecl(const DefinitionSymbol::ParameterDecl& decl);
 
+    /// Gets the list of parameter declarations for this class.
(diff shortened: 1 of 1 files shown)


include/slang/ast/symbols/ClassSymbols.h
```diff
@@ -239,6 +239,8 @@ class SLANG_EXPORT GenericClassDefSymbol final : public Symbol {
     /// Set to true if the generic class is an interface class.
     bool isInterface = false;
 
+    SmallVector<DefinitionSymbol::ParameterDecl, 8> paramDecls;
```
MikePopoloski: Does this need to be the mutable vector or can you add a getter method to return it via span, which would be preferred?

PR #1829, opened 2026-05-13
Expressions: Better bad expressions in ConversionExpression::fromSyntax

AndrewNolte:
Stacked PRs:
 * #1831
 * #1837
 * #1836
 * __->__#1829
 * #1839

--- --- ---

[View individual changes](https://github.com/AndrewNolte/slang/commit/a05d487a2d075376651d4d8fecddf31cfb6fe7e3)

### DIFF
--- a/source/ast/expressions/ConversionExpression.cpp
+++ b/source/ast/expressions/ConversionExpression.cpp
@@ -268,18 +268,28 @@ static bool actuallyNeededCast(const Type& type, const Expression& operand) {
 Expression& ConversionExpression::fromSyntax(Compilation& comp, const CastExpressionSyntax& syntax,
                                              const ASTContext& context,
                                              const Type* assignmentTarget) {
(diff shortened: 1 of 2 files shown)


source/ast/expressions/ConversionExpression.cpp
```diff
@@ -268,18 +268,31 @@ static bool actuallyNeededCast(const Type& type, const Expression& operand) {
...
     auto& targetExpr = bind(*syntax.left, context, ASTFlags::AllowDataType);
-    if (targetExpr.bad())
-        return badExpr(comp, nullptr);
+    if (targetExpr.bad()) {
+        return badExpr(comp, result(bindOperand()));
+    }
 
-    auto type = &comp.getErrorType();
     Expression* operand;
+    auto conversion = [&](ConversionKind cast = ConversionKind::Explicit) {
```
MikePopoloski: This helper seems like a bug waiting to happen, since it captures a local that may or may not be valid when it's called. I'd just make everyone call result() with the right argument.

PR #1833, opened 2026-05-13
Type Printer: flexible quote char

AndrewNolte:
Stacked PRs:
 * #1831
 * #1829
 * #1839
 * #1838
 * #1837
 * #1836
 * #1835
 * #1834
 * __->__#1833
 * #1832
 * #1830

--- --- ---

[View individual changes](https://github.com/AndrewNolte/slang/commit/d5ce1700414ef93d6bbf2db7e7c170a3917beebe)

PR #1830, opened 2026-05-13
Driver: add checkFiles flag to processOptions

AndrewNolte:
Stacked PRs:
 * #1831
 * #1829
 * #1839
 * #1838
 * #1837
 * #1836
 * #1835
 * #1834
 * #1833
 * #1832
 * __->__#1830

--- --- ---

[View individual changes](https://github.com/AndrewNolte/slang/commit/0148f7fc99b5a4e0bbd866aaa968204334433275)

PR #1834, opened 2026-05-13
Type Printer: Add printIntegralRange option

AndrewNolte:
Stacked PRs:
 * #1831
 * #1829
 * #1839
 * #1838
 * #1837
 * #1836
 * #1835
 * __->__#1834
 * #1833
 * #1832
 * #1830

--- --- ---

[View individual changes](https://github.com/AndrewNolte/slang/commit/bb85e0b3dbc6e67026308c1deb5fd7d3074ad09d)

### DIFF
--- a/bindings/python/TypeBindings.cpp
+++ b/bindings/python/TypeBindings.cpp
@@ -127,6 +127,7 @@ void registerTypes(py::module_& m) {
         .def_readwrite("anonymousTypeStyle", &TypePrintingOptions::anonymousTypeStyle)
         .def_readwrite("skipScopedTypeNames", &TypePrintingOptions::skipScopedTypeNames)
         .def_readwrite("fullEnumType", &TypePrintingOptions::fullEnumType)
+        .def_readwrite("printIntegralRange", &TypePrintingOptions::printIntegralRange)
         .def_readwrite("skipTypeDefs", &TypePrintingOptions::skipTypeDefs);
 
     py::native_enum<TypePrintingOptions::AnonymousTypeStyle>(typePrintingOptions,
--- a/source/ast/types/TypePrinter.cpp
+++ b/source/ast/types/TypePrinter.cpp
@@ -56,6 +56,26 @@ void TypePrinter::append(const Type& type) {
 
     if (options.printAKA && type.kind == SymbolKind::TypeAlias)
         printAKA(type);
+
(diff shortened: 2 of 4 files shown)


source/ast/types/TypePrinter.cpp
```diff
@@ -48,11 +52,31 @@ void TypePrinter::append(const Type& type) {
...
+    if (options.printIntegralRange) {
+        auto& canon = type.getCanonicalType();
+        switch (canon.kind) {
+            case SymbolKind::EnumType:
+            case SymbolKind::PackedStructType:
+            case SymbolKind::PackedUnionType: {
+                auto& intType = canon.as<IntegralType>();
+                auto range = intType.getBitVectorRange();
+                auto arrKind = intType.isFourState ? "logic" : "bit";
+                type.isSimpleType();
```
MikePopoloski: Random call to isSimpleType here?

source/ast/types/TypePrinter.cpp
```diff
@@ -48,11 +52,31 @@ void TypePrinter::append(const Type& type) {
...
+                buffer->append(" (");
+                maybeAddQuote();
```
MikePopoloski, line 70: It's hard to imagine you'll want the quote chars here?

bindings/python/TypeBindings.cpp
```diff
@@ -121,8 +121,7 @@ void registerTypes(py::module_& m) {
         .finalize();
 
     py::classh<TypePrintingOptions> typePrintingOptions(m, "TypePrintingOptions");
-    typePrintingOptions.def_readwrite("addSingleQuotes", &TypePrintingOptions::addSingleQuotes)
-        .def_readwrite("elideScopeNames", &TypePrintingOptions::elideScopeNames)
+    typePrintingOptions.def_readwrite("elideScopeNames", &TypePrintingOptions::elideScopeNames)
```
MikePopoloski: Add the new option here.

source/ast/types/TypePrinter.cpp
AndrewNolte: The use case here is for adding backticks around anything code-like. The usage here is also similar to the (aka ...), which does have quotes. Although in that case it serves the purpose of separating the aka from the code, whereas this does look worse if using single quotes.

PR #1836, opened 2026-05-13
Preproc: Add MacroUsageMetadata to preprocessor and syntax tree

AndrewNolte:
Stacked PRs:
 * __->__#1836
 * #1831
 * #1837

--- --- ---

[View individual changes](https://github.com/AndrewNolte/slang/commit/d1c91e8e9a594dffca63609a9e041a14a204d4ac)

### DIFF
--- a/include/slang/parsing/PreprocessorMetadata.h
+++ b/include/slang/parsing/PreprocessorMetadata.h
@@ -1,0 +1,50 @@
+//------------------------------------------------------------------------------
+//! @file PreprocessorMetadata.h
+//! @brief Metadata collected during preprocessing
+//
+// SPDX-FileCopyrightText: Michael Popoloski
+// SPDX-License-Identifier: MIT
+//------------------------------------------------------------------------------
+#pragma once
+
+#include <string_view>
+#include <vector>
+
+#include "slang/syntax/SyntaxFwd.h"
+#include "slang/text/SourceLocation.h"
+
+namespace slang::parsing {
+
+/// Metadata about an include directive that was invoked.
+struct SLANG_EXPORT IncludeMetadata {
+    const syntax::IncludeDirectiveSyntax* syntax;
+    std::string_view path;
+    SourceBuffer buffer;
+    bool isSystem;
+};
+
+/// Metadata about a macro reference (usage or undef) and its active definition.
+struct SLANG_EXPORT MacroRefMetadata {
+    /// The syntax node referencing the macro (MacroUsage or UndefDirective)
+    const syntax::SyntaxNode* syntax;
+    /// The macro definition that was used for the expansion
+    const syntax::DefineDirectiveSyntax* definition;
+};
+
+/// Various bits of metadata collected during preprocessing.
+struct SLANG_EXPORT PreprocessorMetadata {
+    /// The macros that were defined at the end of preprocessing.
+    std::vector<const syntax::DefineDirectiveSyntax*> definedMacros;
+
+    /// The include directives that were encountered while preprocessing.
+    std::vector<IncludeMetadata> includeDirectives;
+
+    /// The macro references that were expanded while preprocessing.
+    std::vector<MacroRefMetadata> macroRefs;
+
+    /// The source buffer IDs that were pushed into the preprocessor, in order.
+    /// This includes top-level and included source files, but not macro expansion buffers.
+    std::vector<BufferID> sourceBufferIds;
+};
+
+} // namespace slang::parsing
--- a/source/parsing/Preprocessor.cpp
+++ b/source/parsing/Preprocessor.cpp
@@ -103,6 +103,7 @@ void Preprocessor::pushSource(std::string_view source, std::string_view name) {
 
 void Preprocessor::pushSource(SourceBuffer buffer) {
     SLANG_ASSERT(buffer.id);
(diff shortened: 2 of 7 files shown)


source/parsing/Preprocessor.cpp
```diff
@@ -858,12 +862,24 @@ Trivia Preprocessor::handleDefineDirective(Token directive) {
 }
 
 std::pair<Trivia, Trivia> Preprocessor::handleMacroUsage(Token directive) {
+    // Look up the definition before expansion (it may be undef'd later)
+    auto macro = findMacro(directive);
+    auto* definition = macro.valid() && !macro.isIntrinsic() ? macro.syntax : nullptr;
```
MikePopoloski: This is only going to register top-level macro uses, not ones inside other macro bodies or arguments, which is probably not what you want? I also don't want to duplicate the findMacro call for no good reason.

source/syntax/SyntaxTree.cpp
MikePopoloski, line 118: The argument list is getting kind of ridiculous. Perhaps it should be factored out into an argument struct?

source/parsing/Preprocessor.cpp
AndrewNolte: For the slang-server use case I think we just want top level. I'll change it to pass the findMacro result down

source/parsing/Preprocessor.cpp
```diff
@@ -858,13 +867,25 @@ Trivia Preprocessor::handleDefineDirective(Token directive) {
...
     inMacroBody = true;
-    auto [actualArgs, extraTrivia] = handleTopLevelMacro(directive);
+    auto [actualArgs, extraTrivia] = handleTopLevelMacro(directive, macroDef);
     inMacroBody = false;
 
-    auto syntax = alloc.emplace<MacroUsageSyntax>(directive, actualArgs);
-    return std::make_pair(Trivia(TriviaKind::Directive, syntax), extraTrivia);
+    auto* usageSyntax = alloc.emplace<MacroUsageSyntax>(directive, actualArgs);
+
+    // auto* definition = macro ? macro.syntax : nullptr;
```
MikePopoloski: Remove commented out code.

source/syntax/SyntaxTree.cpp
```diff
@@ -23,16 +23,16 @@ SyntaxTree::SyntaxTree(SyntaxNode* root, SourceManager& sourceManager, BumpAlloc
     rootNode(root), library(library), sourceMan(sourceManager), alloc(std::move(alloc)) {
     metadata = std::make_unique<ParserMetadata>(ParserMetadata::fromSyntax(*root));
     if (parent) { // copy parent's info
-        for (size_t i = 0; i < parent->macros.size(); ++i) {
-            auto macro = parent->macros[i];
+        const auto& parentMeta = parent->getMetadata();
```
MikePopoloski: Should copy include directives as well?

source/parsing/PreprocessorMetadata.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+namespace slang::parsing {
+
+using namespace syntax;
+
+PreprocessorMetadata PreprocessorMetadata::fromPreTransform(
+    const PreprocessorMetadata& preTransform, BumpAllocator& alloc) {
+    PreprocessorMetadata result;
+
+    SmallVector<std::pair<const DefineDirectiveSyntax*, const DefineDirectiveSyntax*>> macroMap;
```
MikePopoloski: Why not use an actual map?

include/slang/parsing/PreprocessorMetadata.h
```diff
@@ -0,0 +1,59 @@
...
+#include "slang/text/SourceLocation.h"
+
+namespace slang {
+class BumpAllocator;
+}
+
+namespace slang::parsing {
+
+/// Metadata about an include directive that was invoked.
+struct IncludeMetadata {
```
MikePopoloski: These need the SLANG_EXPORT tag as well.

AndrewNolte:
I don't think it really makes sense for transformed trees to have PreprocessorMetadata actually, since to faithfully reconstruct you'd have to run a preprocessor. I don't even think the ParserMetadata should be collected by default for transforms, and should instead be lazy loaded in getMetadata(). The primary function for SyntaxRewriter is just to re-print the tree after the transform right?

MikePopoloski:
The ParserMetadata definitely needs to be correct in case they pass the syntax tree along to a Compilation, which makes use of that. I suppose the PreprocessorMetadata isn't needed for that so you could just discard it?

ParserMetadata can be reconstructed by visiting the syntax tree. Actually, the preprocessor metadata *also* could be reconstructed that way. Perhaps the Preprocessor shouldn't be tracking this stuff at all, and instead you can have a utility method to walk a syntax tree and find the preprocessor trivia you care about?

AndrewNolte:
I don't think the ParserMetadata should be empty, but that it can be lazily constructed for transformed trees since most use cases just print the tree and don't use the metadata.

Yeah for macro refs I think reconstruction should be possible, but it's not as feasible for things like include buffers and source id collection. Resolved include paths depend on active include search paths and other data that's lost.

MikePopoloski:
Ok yeah, that sounds right.

PR #1837, opened 2026-05-13
Driver: track define sources and command files used

AndrewNolte:
Stacked PRs:
 * #1836
 * #1831
 * __->__#1837

--- --- ---

[View individual changes](https://github.com/AndrewNolte/slang/commit/6c7e0031d825b6ad855eb3bb0e1253bfea3f2add)

### DIFF
--- a/include/slang/driver/Driver.h
+++ b/include/slang/driver/Driver.h
@@ -21,6 +21,7 @@
 #include "slang/util/Function.h"
 #include "slang/util/LanguageVersion.h"
 #include "slang/util/OS.h"
+#include "slang/util/ScopeGuard.h"
 #include "slang/util/Util.h"
 
 namespace slang {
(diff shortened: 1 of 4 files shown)


source/driver/Driver.cpp
```diff
@@ -565,10 +574,13 @@ bool Driver::processCommandFiles(std::string_view pattern, bool makeRelative, bo
             result = parseCommandLine(buffer->data, parseOpts);
         }
 
+        currentParseSource = savedSource;
+
         if (makeRelative)
             fs::current_path(currPath, ec);
 
         activeCommandFiles.erase(path);
+        defineSources.try_emplace(path);
```
MikePopoloski: Why add this here? I realize the comment says there will be an entry for every command file even if there are no defines but why?

include/slang/driver/Driver.h
```diff
@@ -500,7 +506,15 @@ class SLANG_EXPORT Driver {
     bool reportLoadErrors();
 
     bool anyFailedLoads = false;
+
+public:
+    /// The file currently being parsed; set by processCommandFiles and can be
+    /// set externally before parseCommandLine to attribute -D defines.
+    std::filesystem::path currentParseSource;
```
MikePopoloski: This is a weird name for this since it implies source file parsing. Why not currentDefineSource? Also add a setter for this so you don't need to make the variable public.

include/slang/driver/Driver.h
```diff
@@ -490,6 +490,12 @@ class SLANG_EXPORT Driver {
     /// Sets whether terminal output should use color.
     void setTerminalColorsEnabled(bool enable);
 
+    /// Gets the map from source file to the raw -D defines it contributed.
+    /// Every processed command file has an entry, even if it contributed no defines.
+    const flat_hash_map<std::filesystem::path, std::vector<std::string>>& getDefineSources() const {
```
MikePopoloski: Add Python bindings for new methods.

source/driver/Driver.cpp
AndrewNolte: I use it to also track which command files were processed, so the file can be watched for changes. I'll update the naming here to make more sense

include/slang/driver/Driver.h
```diff
@@ -490,6 +491,37 @@ class SLANG_EXPORT Driver {
...
+    /// Gets the map from command file to metadata collected while processing it.
+    /// Every processed command file has an entry, even if no metadata was collected.
+    const flat_hash_map<std::filesystem::path, CommandFileMetadata>& getCommandFileMetadata()
+        const {
+        return commandFileMetadata;
+    }
+
+    /// Temporarily sets the file used to attribute parsed options.
+    /// The returned guard will restore the previous command file when destroyed.
+    [[nodiscard]] auto setCurrentCommandFile(std::filesystem::path path) {
```
MikePopoloski, line 509: Is this something you need to modify downstream? If not we can just make it private if it's only used internally to the Driver.

include/slang/driver/Driver.h
```diff
@@ -490,6 +491,37 @@ class SLANG_EXPORT Driver {
...
+        auto canonicalPath = std::filesystem::weakly_canonical(path, ec);
+        if (!ec)
+            path = std::move(canonicalPath);
+
+        auto savedCommandFile = std::move(currentCommandFile);
+        currentCommandFile = std::move(path);
+        if (!currentCommandFile.empty())
+            commandFileMetadata.try_emplace(currentCommandFile);
+
+        return ScopeGuard([this, savedCommandFile = std::move(savedCommandFile)]() mutable {
```
MikePopoloski: Could just be `savedCommandFile = std::exchange(currentCommandFile, std::move(path))`

MikePopoloski (commented):
Yeah, this looks a lot better. Just some minor comments now.

include/slang/driver/Driver.h
AndrewNolte: Yeah, to set the source to one of the config files like `.slang/server.json`

PR #1839, opened 2026-05-13
Lookups: Add instance arrays and gen arrays to lookup paths

AndrewNolte:
Stacked PRs:
 * #1831
 * #1837
 * #1836
 * #1829
 * __->__#1839

--- --- ---

[View individual changes](https://github.com/AndrewNolte/slang/commit/8ebc61a15d049e522cc68cd0a062c0004d51b0ad)

### DIFF
--- a/include/slang/ast/Lookup.h
+++ b/include/slang/ast/Lookup.h
@@ -247,14 +247,14 @@ public:
                      bitmask<LookupFlags> flags, LookupResult& result);
 
     /// Performs an unqualified lookup in this scope, then recursively up the parent
-    /// chain until we reach root or the symbol is found. No errors are reported if
-    /// no symbol can be found.
+    /// chain until we reach root or the symbol is found.
+    /// Does not report an error if the symbol is not found.
     static const Symbol* unqualified(const Scope& scope, std::string_view name,
                                      bitmask<LookupFlags> flags = LookupFlags::None);
 
     /// Performs an unqualified lookup in this scope, then recursively up the parent
-    /// chain until we reach root or the symbol is found. Reports an error if
-    /// the symbol is not found.
+    /// chain until we reach root or the symbol is found.
+    /// Reports an error if the symbol is not found.
     static const Symbol* unqualifiedAt(const Scope& scope, std::string_view name,
                                        LookupLocation location, SourceRange sourceRange,
                                        bitmask<LookupFlags> flags = LookupFlags::None);
--- a/source/ast/Lookup.cpp
+++ b/source/ast/Lookup.cpp
@@ -1211,6 +1211,11 @@ void Lookup::name(const NameSyntax& syntax, const ASTContext& context, bitmask<L
     }
 
     unwrapResult(scope, result);
+    if (!name.selectors.empty() && result.found &&
+        (result.found->kind == SymbolKind::InstanceArray ||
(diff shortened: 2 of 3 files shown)


source/ast/Lookup.cpp
```diff
@@ -2026,6 +2026,15 @@ void Lookup::unqualifiedImpl(const Scope& scope, std::string_view name, LookupLo
                         result.nameRange = savedRange;
                     }
                 }
+                else {
```
MikePopoloski: This is a weird spot to put this, the select doesn't happen here and this function can be called from a variety of places. Probably it should happen in applySelectors(), or right before it, since that method is shared with the unqualified case and the downward lookup case. It's a little strange to expect a path at all though, since this is arguable not a hierarchical lookup and so it doesn't really have a hierarchical path.

include/slang/ast/Lookup.h
```diff
@@ -247,14 +247,14 @@ class SLANG_EXPORT Lookup {
                      bitmask<LookupFlags> flags, LookupResult& result);
 
     /// Performs an unqualified lookup in this scope, then recursively up the parent
-    /// chain until we reach root or the symbol is found. No errors are reported if
-    /// no symbol can be found.
+    /// chain until we reach root or the symbol is found.
+    /// Does not report an error if the symbol is not found.
```
MikePopoloski, line 251: These comment changes are basically saying the same thing, any reason for the churn?

AndrewNolte: I was trying to make the difference more apparent, since the two doc comments say mostly the same thing. I wonder if it makes sense to combine the methods.

source/ast/Lookup.cpp
AndrewNolte: Should it go in selectors then? It's not there either atm.

MikePopoloski: The selectors are only for select expressions (that apply to a typed value). Indexing into a gen array or instance array is a compile-time constant operation so you just get the specific symbol of the array, no runtime expression evaluation needed.

include/slang/ast/Lookup.h
MikePopoloski: It would be annoying to have to provide the lookup location for the case where no error is needed.

source/ast/Lookup.cpp
AndrewNolte: I guess it can't really go there though, since then you can have a selector before a hierarchical path in the generate case.

include/slang/ast/Lookup.h
AndrewNolte: Yeah I meant like have the location and range optional

source/ast/Lookup.cpp
MikePopoloski: What are you trying to use this for? The idea behind the API was that `path` is not valid if `LookupResultFlags::IsHierarchical` isn't set.

AndrewNolte: For proper token to symbol mapping, which is used for hovers and completions. https://github.com/hudson-trading/slang-server/pull/344/files#diff-c48621925170381653f52eabe1ffdf182dc8bb91f67aaee7c805f6181a35402e and previously https://github.com/hudson-trading/slang-server/pull/193/files#diff-c48621925170381653f52eabe1ffdf182dc8bb91f67aaee7c805f6181a35402e

PR #1840, opened 2026-05-14
Fix attribute segfault

AndrewNolte:
This was also hit in less contrived examples because of instance caching, but that example was difficult to reproduce reliably.

### DIFF
--- a/source/ast/ElabVisitors.h
+++ b/source/ast/ElabVisitors.h
@@ -503,7 +503,10 @@ struct DiagnosticVisitor : public ASTVisitor<DiagnosticVisitor> {
         }
 
         // Visit all attributes and force their values to resolve.
-        for (auto& [_, attrList] : compilation.attributeMap) {
+        // Map is copied in case binding the expression causes map mutation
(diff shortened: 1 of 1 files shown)


source/ast/ElabVisitors.h
```diff
@@ -502,11 +502,29 @@ struct DiagnosticVisitor : public ASTVisitor<DiagnosticVisitor> {
...
         }
 
-        // Visit all attributes and force their values to resolve.
-        for (auto& [_, attrList] : compilation.attributeMap) {
-            for (auto attr : attrList)
+        // Visit all attributes and force their values to resolve. Resolving an
+        // attribute value can bind attributed subexpressions, which adds more
+        // entries to the attribute map, so collect the work first to avoid
+        // invalidating the map iterator.
+        SmallSet<const AttributeSymbol*, 8> visitedAttrs;
```
MikePopoloski: Nested attributes are illegal, so we don't need to iterate here. We can just make a copy of the map and operate on that; any additional entries that get added can just be ignored.

PR #1844, opened 2026-05-21
Analysis: Fix multidriven with --allow-dup-initial-drivers triggering on initial drivers specified in initializer syntax

AndrewNolte:


AndrewNolte:
An alternative solution is not adding initial drivers to the map altogether. I don't think that's ideal since we'd still want to warn on overlaps in initial drivers, but this is not a current behavior for `--allow-dup-initial-drivers`. So to implement those checks, I think it's best to add another `initialDrivers` map to DriverTracker that is used with `--allow-dup-initial-drivers`

PR #1845, opened 2026-05-22
Include conda installation instructions for pyslang

timkpaine:


MikePopoloski:
The conda package isn't something I officially maintain, so there probably needs to be some disclaimer. It looks like that package isn't on 11.0, for example, and I haven't actually verified its contents in any way.

MikePopoloski:
Closing for now since there's been no followup.

PR #1847, opened 2026-05-23
Allow legacy cross_auto_bin_max coverage option

### LINKED ISSUE #1843: Add compatibility-mode support for legacy cross_auto_bin_max coverage option

C++ slang

I ran into some legacy SystemVerilog coverage code that uses the old `cross_auto_bin_max` option on a cross:

```systemverilog
x : cross cp_a, cp_b {
    option.cross_auto_bin_max = 0;
}
```

slang currently rejects this with an unknown member error because `cross_auto_bin_max` is no longer part of the IEEE 1800 coverage option set:

```text
error: no member named 'cross_auto_bin_max' in 'struct{int weight, int goal, string comment, int at_least, ...}'
```

I understand that this is correct behavior for strict IEEE 1800 mode. However, `cross_auto_bin_max` was part of the older [SystemVerilog 3.1a](https://courses.cs.washington.edu/courses/cse371/references/SystemVerilog_Reference_Manual.pdf) coverage option set, where it was allowed at covergroup scope as the default for crosses, and also directly at cross scope.

This option also still appears to be accepted by the major commercial simulators for compatibility with older code. I checked the following on EDA Playground, and they all accept `option.cross_auto_bin_max = 0;` inside a cross:

* Synopsys VCS
* Cadence Xcelium
* Siemens Questa

So the practical issue is that some legacy coverage code still compiles with the commercial tools, but fails with slang. This seems like the kind of gap that `--compat` mode could reasonably cover.

It would be helpful if slang could accept `option.cross_auto_bin_max = <int>` in compatibility mode, for example under `--compat vcs` / `--compat all`, or under a dedicated compatibility flag if that fits better.

The option could be accepted at the same levels where SystemVerilog 3.1a allowed it: covergroup scope and cross scope. Since slang does not compute runtime coverage bin sets anyway, accepting and ignoring the option should be enough for compatibility.

The source can be rewritten to avoid `cross_auto_bin_max`, for example by using explicit bins / `ignore_bins`, or by using `option.cross_retain_auto_bins = 0` in newer IEEE 1800-2023 code.

However, those alternatives require editing legacy code that already builds cleanly with the major commercial simulators. In practice, this makes slang the only tool that needs a source change for this case.

I also looked at whether this could be handled by downgrading the diagnostic from the command line, but that does not seem possible because the error comes from the missing field in the option struct rather than from a compatibility warning that can be remapped.

**Additional context**

Here is a small reproducer:

```systemverilog
module tb;
    bit clk;
    logic [1:0] a, b;

    covergroup cg @(posedge clk);
        cp_a : coverpoint a;
        cp_b : coverpoint b;

        x : cross cp_a, cp_b {
            option.cross_auto_bin_max = 0;
        }
    endgroup

    cg c = new;

    initial #1 $finish;
endmodule
```

hankhsu1996:
## Summary

Fixes #1843. Adds `--allow-cross-auto-bin-max` so legacy code that uses the SystemVerilog 3.1a `cross_auto_bin_max` coverage option compiles under slang. The issue has the full background (3.1a Table 20-2, commercial tool behavior, reproducer); this PR is just the implementation.

## Design

- New `CompilationFlags::AllowCrossAutoBinMax` flag. When set, adds `cross_auto_bin_max` to the implicit `option` struct on covergroup body and cross, marked `ImmutableCoverageOption` to match the existing `auto_bin_max`. Coverpoint scope is left as-is per 3.1a Table 20-2.
- `--allow-cross-auto-bin-max` is also added to `VCS_COMP_FLAGS`, so `--compat=vcs` and `--compat=all` enable it automatically. Same wiring as `--allow-lib-module-redef` and friends.
- slang doesn't compute coverage bins at elaboration, so the field is accepted and parsed but otherwise inert. That's enough to make the legacy syntax accepted, which is what the issue asked for.

## Testing

Three unit tests in `CoverTests.cpp`:
- without the flag, the issue's reproducer still errors with `UnknownMember`
- with the flag, the option works at both covergroup and cross scope
- with the flag, coverpoint scope still rejects

## AI disclosure

### DIFF
--- a/bindings/python/CompBindings.cpp
+++ b/bindings/python/CompBindings.cpp
@@ -69,6 +69,7 @@ void registerCompilation(py::module_& m, py::module_& ast, py::module_& driver) 
         .value("AllowUnnamedGenerate", CompilationFlags::AllowUnnamedGenerate)
         .value("AllowVirtualIfaceWithOverride", CompilationFlags::AllowVirtualIfaceWithOverride)
         .value("AllowArrayConcatAssignPattern", CompilationFlags::AllowArrayConcatAssignPattern)
+        .value("AllowCrossAutoBinMax", CompilationFlags::AllowCrossAutoBinMax)
         .finalize();
 
     py::classh<CompilationOptions>(ast, "CompilationOptions")
--- a/docs/command-line-ref.dox
+++ b/docs/command-line-ref.dox
@@ -662,6 +662,16 @@ redefinition produces a `duplicate-definition` warning and a different-kind rede
 
 This option is enabled automatically when `--compat=vcs` or `--compat=all` is used.
 
+`--allow-cross-auto-bin-max`
+
+Allow the legacy `cross_auto_bin_max` coverage option to be set on covergroups and crosses.
+This option was part of the SystemVerilog 3.1a coverage option set but was removed from
+IEEE 1800; some tools still accept it for compatibility with older code. When this flag is
+set, the option is accepted and parsed but has no effect (slang does not compute coverage
+bins at elaboration).
+
+This option is enabled automatically when `--compat=vcs` or `--compat=all` is used.
+
 `--cmd-ignore <vendor_cmd>,<N>`
 
 Define rule to ignore vendor command &lt;vendor_cmd> with its following &lt;N> parameters.
--- a/include/slang/ast/Compilation.h
+++ b/include/slang/ast/Compilation.h
@@ -130,9 +130,14 @@ enum class SLANG_EXPORT CompilationFlags {
     AllowLibModuleRedefinition = 1 << 16,
 
     /// Don't instantiate unreferenced modules to perform semantic checking on them.
-    IgnoreUninstantiatedModules = 1 << 17
+    IgnoreUninstantiatedModules = 1 << 17,
+
+    /// Allow the legacy `cross_auto_bin_max` coverage option to be set on covergroups
+    /// and crosses. This option was part of SystemVerilog 3.1a but is not in IEEE 1800;
+    /// some tools still accept it for compatibility with older code.
+    AllowCrossAutoBinMax = 1 << 18
 };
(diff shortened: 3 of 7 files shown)


AndrewNolte:
I disagree with the implementation here; This doesn't change compilation behavior in a meaningful enough way to justify a new compilation option. I think this should instead be a suppressible warning for legacy/deprecated options, and should be utilize a new `VariableFlags::Deprecated` flag.

hankhsu1996:
Yeah, I think I can see the argument for making this a suppressible warning instead of a separate compilation option. If we go that route, I'd probably frame it as a non-standard legacy coverage option rather than a deprecated option though. From what I found, `cross_auto_bin_max` was defined in the old Accellera SystemVerilog 3.1a coverage option set, but didn't make it into the IEEE SystemVerilog standards; Sigasi seems to model it as a non-standard covergroup option warning too. So maybe something like `NonstandardCoverageOption` under the pedantic group would be a better fit than `Deprecated`. I'm happy to rework this in that direction if @MikePopoloski prefers that model.

MikePopoloski:
Unfortunately I think that approach is not as straightforward; I'm not sure there's a great place to put such a check without checking every symbol of every assignment to see if it's such a coverage option. 

Also I think this change is actually detectable at the language level and so code could (very unlikely but still) be determining behavior based on it. Like for example:
```
if ($bits(cg.option) > whatever) begin
  // I'm in SystemVerilog 3.1a mode
end else begin
  // I'm in IEEE compliant mode
end
```

I actually tried to replicate this on edaplayground but the tools are buggy and I can reliably crash VCS this way so again, unlikely anyone is doing this in practice.

You can in theory also do the following and get different output on different simulators:
```
if (stringContains($sformatf("%p", cg.option), "cross_auto_bin_max")) begin
  $display("SystemVerilog 3.1a");
end else begin
  $display("IEEE SystemVerilog");
end
```

PR #1851, opened 2026-05-27
export slang::ast::SFormat funcs

toddstrader:
These are handy.


include/slang/ast/SFormat.h
MikePopoloski: You probably need the options too?

toddstrader: Yeah, I didn't added it because it wasn't necessary to build / make progress.  But I see now that other layout only structs get exported, so I'll follow suit here.

PR #1856, opened 2026-05-30
Parallelize library search parsing with -j

ebrevdo:
Pass the driver thread pool into SourceLoader library search resolution so -y and -Y discovered files can parse in parallel by dependency depth.

Preserve the serial fixed-point merge order by caching speculatively parsed pending trees whenever an earlier merge discovers new missing names, then replaying them later only if still unresolved.

Add unit and driver coverage for parallel library depths and the order-sensitive case where an earlier library load satisfies a later pending name.

### DIFF
--- a/source/driver/SourceLoader.cpp
+++ b/source/driver/SourceLoader.cpp
@@ -8,6 +8,7 @@
 #include "slang/driver/SourceLoader.h"
 
 #include <fmt/core.h>
+#include <iterator>
 
 #include "slang/parsing/Preprocessor.h"
 #include "slang/syntax/AllSyntax.h"
@@ -433,8 +434,8 @@ void SourceLoader::loadTrees(SyntaxTreeList& syntaxTrees,
         findMissingNames(addedTree);
     };
 
-    // Keep the worklist as a LIFO stack. Parallel batches can parse ahead, but trees are
-    // still committed one at a time in the same order as the serial fixed-point search.
+    // The worklist is a LIFO stack. Parsed batches are committed one tree at a time so
+    // newly discovered names can take precedence over older pending loads.
     while (!worklist.empty()) {
         if (!pool || worklist.size() < MinFilesForThreading) {
             auto load = std::move(worklist.back());
@@ -453,10 +454,10 @@ void SourceLoader::loadTrees(SyntaxTreeList& syntaxTrees,
         std::vector<PendingLoad> batch;
         batch.swap(worklist);
 
-        // Buffer lookup stays serial; only parsing the found buffers runs concurrently.
+        // Fill buffers before launching workers so findBufferFunc is only called here.
         std::vector<SourceBuffer> buffers(batch.size());
         for (size_t i = 0; i < batch.size(); i++) {
-            if (!knownNames.contains(batch[i].name) && !batch[i].tree)
+            if (!batch[i].tree && !knownNames.contains(batch[i].name))
                 buffers[i] = findBufferFunc(batch[i].name);
         }
 
@@ -467,7 +468,7 @@ void SourceLoader::loadTrees(SyntaxTreeList& syntaxTrees,
         pool->wait();
 
         for (size_t i = batch.size(); i-- > 0;) {
-            if (knownNames.contains(batch[i].name) || !batch[i].tree)
+            if (!batch[i].tree || knownNames.contains(batch[i].name))
                 continue;
 
             addTree(std::move(batch[i].tree));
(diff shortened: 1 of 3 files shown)


MikePopoloski:
This makes this path much more complicated to understand. Does it actually provide a large performance win? What's the use case / motivation for the change?

ebrevdo:
Thank you for the feedback!  I've updated to what is hopefully a much less complicated intervention.  For me this is a one or two magnitude performance improvement at the very start of processing large IPs; and it's something I also run at scale in batch runs for basic analysis of many IPs of varying sizes.

MikePopoloski:
Ok, that's interesting. I'm a little surprised that the library search feature is used where performance is a big concern, given that you could list out all of the files explicitly and save the whole step, but it's fine to make these improvements too.

source/driver/SourceLoader.cpp
```diff
@@ -415,22 +421,72 @@ void SourceLoader::loadTrees(SyntaxTreeList& syntaxTrees,
...
+        return tree;
+    };
+
+    auto addTree = [&](std::shared_ptr<SyntaxTree> tree) {
+        auto& addedTree = syntaxTrees.emplace_back(std::move(tree));
+        addKnownNames(addedTree);
+        findMissingNames(addedTree);
+    };
+
+    // Keep the worklist as a LIFO stack. Parallel batches can parse ahead, but trees are
```
MikePopoloski: Comments should be written to describe the behavior as it exists in the source, not in reference to previous logic that no longer exists, which future readers can't see. I notice that LLMs make this mistake quite a lot.

source/driver/SourceLoader.cpp
```diff
@@ -415,22 +421,72 @@ void SourceLoader::loadTrees(SyntaxTreeList& syntaxTrees,
...
-            auto tree = syntax::SyntaxTree::fromBuffer(buffer, sourceManager, optionBag,
-                                                       inheritedMacros);
-            tree->isLibraryUnit = true;
-            syntaxTrees.emplace_back(tree);
+        std::vector<PendingLoad> batch;
+        batch.swap(worklist);
 
-            addKnownNames(tree);
-            findMissingNames(tree);
+        // Buffer lookup stays serial; only parsing the found buffers runs concurrently.
```
MikePopoloski: Same here.

source/driver/SourceLoader.cpp
```diff
@@ -415,22 +421,72 @@ void SourceLoader::loadTrees(SyntaxTreeList& syntaxTrees,
...
+        std::vector<SourceBuffer> buffers(batch.size());
+        for (size_t i = 0; i < batch.size(); i++) {
+            if (!knownNames.contains(batch[i].name) && !batch[i].tree)
```
MikePopoloski: Should check the `!batch[i].tree` condition first since it's cheaper than the hash lookup.

source/driver/SourceLoader.cpp
```diff
@@ -415,22 +421,72 @@ void SourceLoader::loadTrees(SyntaxTreeList& syntaxTrees,
...
+        }
+
+        pool->detach_loop(size_t(0), batch.size(), [&](size_t i) {
+            if (buffers[i])
+                batch[i].tree = parseBuffer(buffers[i]);
+        });
+        pool->wait();
+
+        for (size_t i = batch.size(); i-- > 0;) {
+            if (knownNames.contains(batch[i].name) || !batch[i].tree)
```
MikePopoloski: Same here.

source/driver/SourceLoader.cpp
```diff
@@ -415,22 +421,72 @@ void SourceLoader::loadTrees(SyntaxTreeList& syntaxTrees,
...
+
+                // New names discovered by this tree must be processed before older parsed
+                // pending loads. Push the older loads first so the new loads end up on top
+                // of the LIFO stack.
+                for (size_t j = 0; j < i; j++) {
+                    if (batch[j].tree && !knownNames.contains(batch[j].name))
+                        worklist.push_back(std::move(batch[j]));
+                }
+
+                for (auto& load : newLoads)
```
MikePopoloski: I think this should be a single `insert` call for efficiency.

source/driver/SourceLoader.cpp
ebrevdo: yah this is annoying.  no theory of mind :P

ebrevdo:
thanks for bearing with the slop.  i try to keep it minimal; but it was overly eager to explain how it tries to keep serial invariants.  hopefully the remaining comments are useful to future readers.

MikePopoloski:
Looks like the build is failing, please fix that up, then this should be good to go.

PR #1865, opened 2026-06-09
Use path string for name if not proximate.

jpienaar:
This is a bit of a RFC: currently it seems one can either do proximate or just filename for the buffers in SourceManager (e.g., one can't just populate the SourceManager using path names desired and it gets propagated it does get changed one way or another). Except via chdir which is undesirable in multithreaded context. While we wanted to use absolute paths here.

Would a change like this to enable dictating name via SourceManager be desirable?

### DIFF
--- a/source/text/SourceManager.cpp
+++ b/source/text/SourceManager.cpp
@@ -703,7 +703,7 @@ SourceBuffer SourceManager::cacheBuffer(fs::path&& path, std::string&& pathStr,
     }
 
     if (name.empty())
-        name = getU8Str(path.filename());
+        name = getU8Str(path);
 
     std::unique_lock<std::shared_mutex> lock(mutex);
 


MikePopoloski:
I'm not too concerned with the behavior when proximate paths are disabled either way, since in-tree it's only used for tests. That being said, the `name` field in the FileData is intended to be just the name of the file itself; the full path is also stored in the `fullPath` field, and can be retrieved via `SourceManager::getFullPath`, so I'm not sure what your intended use case is? If you just want to force absolute paths in the diagnostic output you can use `DiagnosticClient::showAbsPaths`, which is hooked up to the CLI flag `--diag-abs-paths`

jpienaar:
Yes fair point, its basically a switch needs to be somewhere (to recover the name), and I had seen name being set to `fs::proximate(path, ec))` so figured name can be a path and gives option to do this beyond CWD relative or filename or absolute. We have a flow that imports multiple modules in parallel during one pass, and then processes some information on them later in parallel again, today it is all on same machine and can just use absolute paths as nothing changes. But I wanted to make it robust to moving around (each module can remain self-contained, but need not be at exactly the same absolute path between two tools). Definitely something that can be done in another way. But this seemed appealing given delta & existing flag exists.

MikePopoloski:
Ok sure, this basically doesn't impact slang upstream so I'm fine with merging it. If someone downstream elsewhere is relying on this and complains we can figure out something more involved.

PR #1868, opened 2026-06-11
fix data race in getState() that corrupts the heap during parallel analysis

likeamahoney:
### What

Under parallel analysis multiple worker threads could share the **same** `WorkerState`, racing on its `Diagnostics` vector. Result: heap corruption (use-after-free / overflow in `SmallVector<Diagnostic>::emplaceRealloc` via `addUnusedDiag` to `Diagnostics::add`), showing up as random SIGSEGV/SIGABRT or allocator assertions. First seen on `gcc-debug-shared` -  it's the source of the  intermittent failures on that preset (for example here - https://github.com/MikePopoloski/slang/actions/runs/27310618389/job/80679614776). Needs ≥ 2 cores and is timing/layout  sensitive (≈100% in the `unittests` binary, ~never in `slang`), hence the  flakiness.

  ### Why

`getState()` picked the per-thread state with `BS::this_thread::get_index()`.  That index lives in an `inline static thread_local` in the vendored   `BS_thread_pool.hpp`. Under `-fvisibility=hidden` (shared build) it isn't exported, so each module gets its **own copy**: the pool's workers (instantiated  in the executable) set one copy, while `getState()` in `libsvlang.so` reads another that's never written then `get_index()` returns `nullopt` then every worker  falls into the same fallback slot and shares one `WorkerState`.

  ### Fix

Don't rely on the cross-module thread index. `AnalysisManager` hands out a  distinct `workerStates` slot per thread via an atomic counter, cached in a `thread_local` inside the non-inline `getState()` (a single symbol in the .so, so it can't duplicate like `my_index` did).

  ### Also: `gcc-asan` preset

The bug is a heap UAF, which **TSan doesn't report** - so the existing TSan-only `clang-sanitize` preset stayed clean. ASan pinpointed it immediately. Adding a `gcc-asan` preset (inherits `gcc-debug-shared`, `-fsanitize=address`,  mimalloc off) gives an in-tree heap-error build for this class of bug.

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -171,6 +171,18 @@
       }
     },
     {
+      "name": "gcc-asan",
+      "displayName": "GCC ASan",
+      "description": "GCC debug shared build with AddressSanitizer + UBSan",
+      "inherits": "gcc-debug-shared",
+      "cacheVariables": {
+        "CMAKE_CXX_FLAGS": "-DSLANG_ASSERT_ENABLED -fsanitize=address,undefined -fno-sanitize-recover=undefined -fno-omit-frame-pointer -g",
+        "CMAKE_EXE_LINKER_FLAGS": "-fsanitize=address,undefined",
+        "CMAKE_SHARED_LINKER_FLAGS": "-fsanitize=address,undefined",
+        "SLANG_USE_MIMALLOC": "OFF"
+      }
+    },
+    {
       "name": "gcc-debug-noexcept",
       "displayName": "GCC Debug No Exceptions",
       "inherits": [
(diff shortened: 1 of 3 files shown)


likeamahoney:
very weird ci failure cause, i guess it wasn't me

MikePopoloski:
Hmm yeah, weird that it didn't start showing up until your PR. I don't think this is the right fix though. I'll look at it some more.

likeamahoney:
> Hmm yeah, weird that it didn't start showing up until your PR. I don't think this is the right fix though. I'll look at it some more.

Yeah but I also reproduced it on clean master without my changes. This can be flaky.

MikePopoloski:
Fixed this separately. Build is reliably passing now.

PR #1887, opened 2026-06-22
Add -Wnonstandard-real-case-eq warning for case equality on reals

### LINKED ISSUE #1880: Case equality and inequality on real and shortreal are accepted with no diagnostic

C++ slang

**Describe the bug**

slang accepts the case equality operators `===` and `!==` on real and shortreal operands with no diagnostic, even under `-Weverything`. Table 11-1 in IEEE 1800-2023 lists the operand data types for `===` and `!==` as "Any except real and shortreal", so this construct is outside what the standard allows. The same should hold for realtime, which the LRM treats as a synonym for real.

I could not find documentation or a diagnostic indicating that this is an intentional compatibility extension. There is also a problem beyond the missing check. The standard never defines what `===` means on a real, so accepting it means picking a meaning that the source does not specify. I explain why that matters below.

**To Reproduce**

Scalar:

```sv
module top;
  real x;
  real y;
  bit s;
  initial s = (x === y);
endmodule
```

Unpacked array with a real element type:

```sv
module top;
  real a[];
  real b[];
  bit c;
  initial c = (a === b);
endmodule
```

Both compile with no error and no warning.

**Additional context**

I first assumed this was harmless, on the grounds that a real has no x or z bits, so `===` could only behave like `==`. I no longer think that reasoning holds, and it may be the reason Table 11-1 excludes real in the first place.

If `===` is accepted on real operands, the implementation has to choose an extension semantics for it, and at least two are plausible. One is numeric equality, the ordinary IEEE-754 comparison. The other is representation identity, comparing the underlying bits. Nothing in the source expression selects one over the other, and the two differ observably for signed zero and NaN.

|                              | numeric equality | representation identity |
| ---------------------------- | ---------------- | ----------------------- |
| `+0.0 === -0.0`              | 1                | 0                       |
| NaN vs NaN, identical bits   | 0                | 1                       |

This is the same split C++ has on double. `x == y` is numeric equality, so `+0.0 == -0.0` is true and `NaN == NaN` is false. Comparing the raw bits instead, for example with `memcmp` or `std::bit_cast<uint64_t>`, gives the representation identity column, where the two zeros differ and two NaNs with the same bits match. Both are normal things to want, they just answer different questions.

This is not just an internal detail. SystemVerilog exposes a real's representation through `$realtobits`, `$bitstoreal`, and the shortreal versions, so signed zero is distinguishable from source code. NaN makes the choice sharper. Under numeric equality a NaN is not equal to anything, including itself, while under representation identity two NaNs with the same encoding would match. And if the chosen meaning is representation identity, the result starts to depend on NaN payload preservation across constant folding, the host compiler, and the target. That payload preservation is itself one of the open questions, not something the standard guarantees, which is a shaky thing for a portable language semantics to rest on. None of this comes up for integral types, where case equality has a single well defined meaning.

There is also related evidence that the LRM is conservative about real equality where stable identity is required. Associative array index types must have equality defined for the type, and real, shortreal, and any type containing them are explicitly illegal as index types. I do not read that as proof of the reasoning behind Table 11-1, but it is consistent with the standard avoiding an implicit choice between the two meanings above.

One wrinkle worth being fair about. Section 11.4.5 introduces all four equality operators together and says:

> If either operand is a real operand, then the other operand shall be converted to an equivalent real value, and the expression shall be interpreted as a comparison between real values.

Read on its own that sounds like it covers `===` too. But the same subsection describes `===` and `!==` purely in terms of bits:

> Bits that are x or z shall be included in the comparison and shall match for the result to be considered equal.

which says nothing for an operand that has no x or z bits. My reading is that the real conversion sentence is meant for `==` and `!=`, which Table 11-1 does allow on real, and that Table 11-1 is what governs legality for `===` and `!==`. The associative array index rule fits that reading.

What slang does today is silently accept this and evaluate it as numeric equality. That is probably the least problematic behavior and it matches Verilator, but it is still an extension that picks one of the two meanings, not specified standard behavior, and right now it happens with no diagnostic at all. At a minimum I think this is worth diagnosing rather than accepting silently. My own lean is that rejecting by default is the conformance-correct choice given Table 11-1, ideally as a downgradeable error so existing code can opt back in. But the more useful first question is probably whether this is intended as a compatibility extension, and if so whether it could be surfaced under the pedantic or strict-LRM policy.

Glad to take a look at a fix if this direction sounds reasonable.

hankhsu1996:
## Summary

Fixes #1880. Adds an opt-in `-Wnonstandard-real-case-eq` warning that fires when the case equality operators `===` and `!==` are applied to a real operand. Table 11-1 in the LRM lists the operand types for these operators as "Any except real and shortreal", but slang (like every commercial tool) accepts them and evaluates them just like `==`/`!=`. The issue has the full background — Table 11-1, the surrounding LRM wording, and why it is ambiguous; this PR is just the implementation.

## Design

- Emitted during elaboration from `BinaryExpression::fromComponents`, in the case-equality branch of the numeric path, guarded by `!bothIntegral` so a mixed real/integer comparison is caught too.
- Placed in the `pedantic` group, off by default. It is non-standard per Table 11-1 but every tool accepts it deterministically, so it belongs with the other strict-LRM opt-ins rather than a default-on bundle.
- Scope is the scalar operators. Arrays of reals compared with `===`, and `case` statements on reals, go through other paths and are left as-is.

## Testing

One test in `ExpressionTests.cpp`: `===` and `!==` on `real` and `shortreal` each warn once. The existing "More operator eval tests", which incidentally uses `===` on a `shortreal`, filters the new code at its assertion.

## AI disclosure

### DIFF
--- a/scripts/diagnostics.txt
+++ b/scripts/diagnostics.txt
@@ -815,6 +815,7 @@ warning consecutive-comparison ConsecutiveComparison "comparisons like 'x < y < 
 note NoteForPortConn "for connection to port '{}'"
 warning dynamic-non-procedural DynamicNotProcedural "cannot refer to an element or member of a dynamic type outside of a procedural context"
 warning nonstandard-string-concat NonstandardStringConcat "string concatenation using '+' operator is non-standard"
+warning nonstandard-real-case-eq NonstandardRealCaseEq "case equality operator '{}' on a real operand is non-standard"
 warning inc-dec-bit IncDecBit "increment/decrement of a 1-bit operand"
 warning string-int-concat ConcatWithStringInt "cannot mix strings and integers in a concatenation (use a cast if this is desired)"
 warning range-select-reversed RangeSelectReversed "range of selection [{}:{}] from {} is reversed"
@@ -1329,7 +1330,8 @@ group extra = { empty-member empty-stmt dup-import pointless-void-cast case-gen-
                 bits-of-integer-constant missing-else-clause }
 
 group pedantic = { empty-pattern implicit-net-port nonstandard-escape-code nonstandard-generate format-multibit-strength
-                   nonstandard-sys-func nonstandard-foreach nonstandard-dist nonstandard-randomize isunbounded-param-arg udp-coverage }
+                   nonstandard-sys-func nonstandard-foreach nonstandard-dist nonstandard-randomize isunbounded-param-arg udp-coverage
+                   nonstandard-real-case-eq }
 
 group conversion = { width-trunc width-expand port-width-trunc port-width-expand implicit-conv
                      constant-conversion sign-conversion float-bool-conv int-bool-conv float-int-conv
--- a/scripts/diagnostics.txt
+++ b/scripts/diagnostics.txt
@@ -798,6 +798,7 @@ warning arith-op-mismatch ArithOpMismatch "arithmetic between operands of differ
 warning bitwise-op-mismatch BitwiseOpMismatch "bitwise operation between operands of different types ({} and {})"
 warning comparison-mismatch ComparisonMismatch "comparison between operands of different types ({} and {})"
 warning sign-compare SignCompare "comparison of differently signed types ({} and {})"
+warning real-case-eq RealCaseEq "a real operand has no x or z bits so '{}' acts as ordinary {}"
 warning case-type CaseTypeMismatch "comparison of different types in '{}' statement ({} and {})"
 warning case-outside-range CaseOutsideRange "'{}' item with {} bits can never match the {} bit case expression"
 warning bitwise-rel-precedence BitwiseRelPrecedence "'{}' has lower precedence than '{}'; '{}' will be evaluated first"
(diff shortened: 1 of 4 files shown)


MikePopoloski:
I'm still not convinced this is non-standard. Table 11-1 is incomplete in many ways, so I think including the "nonstandard" as part of the warning name is too forceful. I'd just call this Wreal-case-eq and allow people to opt-in if they want it without implying that they're violating the LRM, which is just under-specified for these cases.

hankhsu1996:
Thanks for pushing back on this. I've pushed a follow-up commit that drops the "nonstandard" framing entirely:

- Renamed to `-Wreal-case-eq` and removed it from `-Wpedantic`; it's now a standalone, off-by-default opt-in warning, with no implication that the code violates the LRM.
- Reworded the message to just describe the behavior: "a real operand has no x or z bits so '===' acts as ordinary equality" (and "inequality" for !==).

PR #1891, opened 2026-06-25
Add skip capture override option

mmicko:
Output callbacks are useful for testing but doing something like 
```
auto guard = slang::OS::captureOutput([&](std::string_view text, bool) { log("%s", text); });
```
could use main application logging system, but due to `skipCapture` parameter of print functions would also skip some important parts. With this change we can override this and make sure that always all is propagated to our capture funciton.

### DIFF
--- a/include/slang/util/OS.h
+++ b/include/slang/util/OS.h
@@ -38,6 +38,9 @@ public:
     static void setStdoutColorsEnabled(bool enabled) { showColorsStdout = enabled; }
     static void setStderrColorsEnabled(bool enabled) { showColorsStderr = enabled; }
 
+    /// Sets whether skip capture override should be enabled for the print() functions.
+    static void setSkipCaptureOverride(bool enabled) { captureOverride = enabled; }
+
     /// Reads a file from @a path into memory. If successful, the bytes are placed
     /// into @a buffer -- otherwise, returns false.
     /// Note that the buffer will be null-terminated.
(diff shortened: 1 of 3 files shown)


MikePopoloski:
Hmm, I don't love this. Stacking a new flag to undo the behavior of the old flag makes me think the shape of this API is just not correct to begin with. There probably needs to be some new structure to allow downstream consumers to do what they want with Driver text output.

mmicko:
Agree. Reason for making it like this was to have minimal impact on existing users, but to be honest, I am not aware if there are any or it is actually used only by tests. Other point is that style info (coloring) is also lost, but that info is anyway fmt specific right now.

MikePopoloski:
I know at least one downstream project is using this in an LSP server; it collects the output and sends them in some structured format (JSON I think) to the editor client, so it wants one callback invocation per line.

mmicko:
One way would be to directly export skipCapture  (maybe give it different name) and not check inside OS.cpp at all, but let client do that. Do you have proposal how to proceed ?

MikePopoloski:
I think I need to rework the API. The Driver shouldn't treat it like it's calling printf(), it should issue structured logs (message + severity, like other diagnostics) and let downstream consumers do what they want with it, with the default being to print to stdout/stderr like it does now. Then I'd get rid of this whole skipCapture stuff from OS::print.

MikePopoloski:
1d2b0c76b690fdfb843e2b87c875ba9bc10d778b gets rid of the skipCapture thing. Driver calls to printError / printWarning now go through the normal diagnostic path and can be intercepted at various points.

@AndrewNolte FYI, I don't think this will disrupt your usage but the timing of output will be slightly different now.

mmicko:
Thanks a lot @MikePopoloski

PR #1900, opened 2026-07-14
Compilation: Add AllowInvalidTop flag

AndrewNolte:
Stacked PRs:
 * #1902
 * #1901
 * __->__#1900
 * #1903

--- --- ---

[View individual changes](https://github.com/AndrewNolte/slang/commit/3c571d4ea8adc31de0619de35cfa0cf06aea4365)
### Compilation: Add AllowInvalidTop flag

This flag sets error types for unset top level params, allowing for better diagnostics when not in a full design context.

### DIFF
--- a/bindings/python/CompBindings.cpp
+++ b/bindings/python/CompBindings.cpp
@@ -70,6 +70,7 @@ void registerCompilation(py::module_& m, py::module_& ast, py::module_& driver) 
         .value("AllowVirtualIfaceWithOverride", CompilationFlags::AllowVirtualIfaceWithOverride)
         .value("AllowArrayConcatAssignPattern", CompilationFlags::AllowArrayConcatAssignPattern)
         .value("AllowCrossAutoBinMax", CompilationFlags::AllowCrossAutoBinMax)
+        .value("AllowInvalidTop", CompilationFlags::AllowInvalidTop)
         .finalize();
 
     py::classh<CompilationOptions>(ast, "CompilationOptions")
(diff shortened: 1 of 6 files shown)


source/ast/symbols/ParameterBuilder.h
```diff
@@ -31,7 +31,12 @@ class ParameterBuilder {
 
     void setAssignments(const syntax::ParameterValueAssignmentSyntax& syntax, bool isFromConfig);
     void setOverrides(const HierarchyOverrideNode* newVal) { overrideNode = newVal; }
+    /// Force invalid values (error type) for all parameters.
     void setForceInvalidValues(bool set) { forceInvalidValues = set; }
+    /// Set invalid values (error type) for parameters that are missing values, rather than
+    /// reporting errors.
+    void setUseInvalidForMissing(bool set) { useInvalidForMissing = set; }
```
MikePopoloski: Why do we need this new flag instead of just using the suppressErrors flag?

tests/unittests/ast/InterfaceTests.cpp
```diff
@@ -580,9 +580,8 @@ endmodule
     compilation.addSyntaxTree(tree);
 
     auto& diags = compilation.getAllDiagnostics();
-    REQUIRE(diags.size() == 2);
+    REQUIRE(diags.size() == 1);
     CHECK(diags[0].code == diag::InvalidModportAccess);
-    CHECK(diags[1].code == diag::ParamHasNoValue);
```
MikePopoloski, line 585: Why did this error go away?

MikePopoloski:
Also need to update Python bindings for the new flag.

source/ast/symbols/ParameterBuilder.h
AndrewNolte: I think at one point it was to avoid setting anyErrors to true, and ensuring the invalid types were set. Although I don't see where that affects things down stream anymore. I'll change it to use that.

tests/unittests/ast/InterfaceTests.cpp
AndrewNolte: Fixed now

PR #1901, opened 2026-07-14
Compilation: Add CheckUninstantiated flag

AndrewNolte:
Stacked PRs:
 * #1910
 * #1909
 * #1902
 * __->__#1901

--- --- ---

[View individual changes](https://github.com/AndrewNolte/slang/commit/c2c4ae42e55554fb30bfaa00d421832fe8ba9279)
### Compilation: Add CheckUninstantiated flag

This compilation flag enables more checks to be done in untaken generate branches, or gen loops with invalid loop params.

### DIFF
--- a/source/ast/Compilation.cpp
+++ b/source/ast/Compilation.cpp
@@ -16,6 +16,7 @@
 #include "slang/ast/SystemSubroutine.h"
 #include "slang/ast/types/TypePrinter.h"
 #include "slang/diagnostics/DiagnosticEngine.h"
+#include "slang/diagnostics/Diagnostics.h"
 #include "slang/diagnostics/LookupDiags.h"
 #include "slang/parsing/Parser.h"
 #include "slang/parsing/Preprocessor.h"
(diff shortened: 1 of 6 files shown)


include/slang/ast/Symbol.h
```diff
@@ -173,6 +173,9 @@ class SLANG_EXPORT Symbol {
     /// Determines whether this symbol represents a value.
     bool isValue() const;
 
+    /// Determines whether this symbol is instantiated.
+    bool isInstantiated() const;
```
MikePopoloski: This is a confusing method to add since Scopes have an isUninstantiated method already. I think it should stay a helper in the Compilation file.

source/ast/symbols/BlockSymbols.cpp
```diff
@@ -913,12 +913,15 @@ GenerateBlockArraySymbol& GenerateBlockArraySymbol::fromSyntax(Compilation& comp
 
     result->entries = entries.copy(comp);
     if (entries.empty()) {
+        // Keep result->entries empty so indexed lookup (g[0], etc.) still reports
```
MikePopoloski: This whole comment is pretty hard to understand. It's also unclear to me why the else block was removed, this should be the same behavior right?

source/ast/Compilation.cpp
```diff
@@ -1686,31 +1689,47 @@ void Compilation::addDiagnostics(const Diagnostics& diagnostics) {
         addDiag(diag);
 }
 
+bool shouldReportUninstantiatedDiag(const DiagCode& code) {
+    switch (code.getSubsystem()) {
+        case DiagSubsystem::Declarations:
+            return true;
+        case DiagSubsystem::Lookup:
+            switch (code.key()) {
+                case diag::ScopeIndexOutOfRange.key():
```
MikePopoloski: I would just put these explicit cases in a hash set and then you don't need the new key() stuff.

source/ast/symbols/BlockSymbols.cpp
AndrewNolte: createBlock() adds to entries, so the loop actually does get traversed. This makes it such that AST Visitors can visit this uninstantiated scope, but it's still not accessible by a gen_arr[0] ref bc it's an uninstantiated block

AndrewNolte: But it doesn't add to result->entries. I could change it to just `result->addMember(entries[0])` to avoid the loop.

source/ast/Compilation.cpp
AndrewNolte: I made this bc before I had a different system that blocked diags from different subsystems in a single switch. So if we expanded this we'd need either multiple hash sets, or again this key thing.

PR #1902, opened 2026-07-15
Analysis: Diag ranges instead of locs for unused vars

AndrewNolte:
Stacked PRs:
 * #1910
 * __->__#1902

--- --- ---

[View individual changes](https://github.com/AndrewNolte/slang/commit/3f0a7c51430d465642946bef5071ff3ee71db50e)
### Analysis: Diag ranges instead of locs for unused vars

For slang-server we want these to cover the whole id.
Related issue: https://github.com/hudson-trading/slang-server/issues/187

### DIFF
--- a/source/analysis/AnalysisManager.cpp
+++ b/source/analysis/AnalysisManager.cpp
@@ -14,6 +14,7 @@
 #include "slang/analysis/DataFlowAnalysis.h"
 #include "slang/ast/ASTDiagMap.h"
 #include "slang/ast/Compilation.h"
+#include "slang/text/SourceManager.h"
 
 namespace slang::analysis {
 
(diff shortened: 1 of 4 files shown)


source/analysis/AnalysisManager.cpp
```diff
@@ -87,8 +98,7 @@ void AnalysisManager::analyze(const Compilation& compilation) {
     if (hasFlag(AnalysisFlags::CheckUnused)) {
         for (auto def : compilation.getUnreferencedDefinitions()) {
             if (!def->name.empty() && def->name != "_"sv && !hasUnusedAttrib(compilation, *def)) {
-                state.context.addDiag(*def, diag::UnusedDefinition, def->location)
-                    << def->getKindString();
+                state.context.addDiag(*def, diag::UnusedDefinition) << def->name;
```
MikePopoloski: This changed getKindString to def->name, which seems wrong.

AndrewNolte: Initially I was going to derive the range from the diag arg, in which case these would be wrong. But yeah this change is unnecessary now.

PR #1904, opened 2026-07-16
Fix Indirect Package exports with refactor

AndrewNolte:
Stacked PRs:
 * #1902
 * #1901
 * __->__#1904

--- --- ---

[View individual changes](https://github.com/AndrewNolte/slang/commit/c3c241e07525712fa8840a028c9c58e90befb4ee)
### Fix Indirect Package exports with refactor

- Removes exports from AST/Pkg scope, removing ability to easily access symbol attributes
- Cleans up other spurious export diagnostics, adds more correct diags for some cases, and adds more test cases

### DIFF
--- a/source/ast/symbols/CompilationUnitSymbols.cpp
+++ b/source/ast/symbols/CompilationUnitSymbols.cpp
@@ -15,6 +15,7 @@
 #include "slang/ast/symbols/MemberSymbols.h"
 #include "slang/ast/types/NetType.h"
 #include "slang/diagnostics/DeclarationsDiags.h"
+#include "slang/diagnostics/LookupDiags.h"
 #include "slang/syntax/AllSyntax.h"
 #include "slang/syntax/SyntaxTree.h"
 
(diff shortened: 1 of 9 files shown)


source/ast/symbols/CompilationUnitSymbols.cpp
```diff
@@ -101,26 +104,193 @@ PackageSymbol& PackageSymbol::fromSyntax(const Scope& scope, const ModuleDeclara
...
+            continue;
+
+        if (auto package = findPackageForExport(item->package.valueText(), *this,
+                                                item->package.location())) {
+            data->starMap->emplace(item->package.valueText(), package);
+        }
+    }
+}
+
+bool PackageSymbol::exportsPackage(std::string_view packageName) const {
```
MikePopoloski: This is only called in one place, probably not worth being its own helper method.

source/ast/symbols/CompilationUnitSymbols.cpp
```diff
@@ -144,8 +323,16 @@ const Symbol* PackageSymbol::findForImport(std::string_view lookupName) const {
         }
     }
 
+    resolveExplicitExports();
```
MikePopoloski: Rather than sprinkling calls to the resolve* methods everywhere, they should probably just be called once at the start of the method? Also I don't see why they need to be separate for explicit and star exports. I think you can merge both into one method, resolveExports, and make it public so you can just get rid of the checkExplicitExports method.

AndrewNolte: This was sort of to keep things as lazy as possible, but package loops aren't allowed anyway. Slang does allow cycles atm though (package ref cycles - not import chain cycles). Maybe in a follow up pr we'd want to refactor to do a topo sort to catch cycles, then eval packages in that order. This would still allow for specifying packages in any order though. I included some cycle tests to maintain that there's no crash.

PR #1905, opened 2026-07-19
Required fmt version 12.2 -> 12

evanwporter:
`12.2` is literally the cutting edge of `fmt` release less than a month ago. Imo it seems a bit excessive, since I want to use my system library of `fmt` (provided by nix) but its only at `12.1.0`

Are there specific features from 12.2 that slang requires?

### DIFF
--- a/external/CMakeLists.txt
+++ b/external/CMakeLists.txt
@@ -5,7 +5,7 @@
 
 # --- fmt lib ---
 if(SLANG_USE_SYSTEM_FMT)
-  find_package(fmt 12.2 REQUIRED)
+  find_package(fmt 12 REQUIRED)
 
   get_target_property(FMT_CFG fmt::fmt IMPORTED_CONFIGURATIONS)
   list(GET FMT_CFG 0 FMT_CFG_FIRST)


evanwporter:
> Enabled the full Dragonbox lookup cache by default for floating-point formatting unless optimizing for binary size (__OPTIMIZE_SIZE__), giving a ~10–25% speedup. Thanks Matthias Kretz for the suggestion. Average time per double on Apple M1 Pro (clang 17, random digits, smaller is better) measured with [dtoa-benchmark](https://github.com/fmtlib/dtoa-benchmark):

- From fmt release page

Update: it seems like you might be using the above features? If so I guess 12.2.0 is necessary :(

MikePopoloski:
fmt releases aren't always perfectly forwards / backwards compatible, for example see https://github.com/MikePopoloski/slang/pull/1878

This change seems fine with me though, as long as it works for you. I expect the option to use the system package will be rarely used in practice.

evanwporter:
> fmt releases aren't always perfectly forwards / backwards compatible, for example see https://github.com/MikePopoloski/slang/pull/1878

Doesn't this go against the whole idea of semver though lol. If fmt is going to break, then it should be a new major release no?

> This change seems fine with me though, as long as it works for you. 

Doesn't this file require 12.2 though?: https://github.com/MikePopoloski/slang/blob/3998d90dc814c01b9c88a3c66b117172adc89ee5/source/util/Fmt.cpp

Note: i have no idea what dragonbox is, I just saw that its defined here and it was mentioned in the release.

> I expect the option to use the system package will be rarely used in practice.

My end goal is to add `slang-server` to nixpkgs; https://search.nixos.org/packages. This way I can just integrate slang-server with the rest of my lsps, without manually having too manage builds;

https://github.com/evanwporter/dotfiles/blob/d87025c8c221e6ef1be0428b354616ab78f4f143/nix/home/modules/neovim.nix#L43-L60

To add it to nixpkg you need to be able to build the binary in a sandbox w/ no internet access. This means no FetchFromGithub.

MikePopoloski:
> Doesn't this go against the whole idea of semver though lol. If fmt is going to break, then it should be a new major release no?

I can't control what other projects do, only my own. 

> Doesn't this file require 12.2 though?: https://github.com/MikePopoloski/slang/blob/3998d90dc814c01b9c88a3c66b117172adc89ee5/source/util/Fmt.cpp
> 
> Note: i have no idea what dragonbox is, I just saw that its defined here and it was mentioned in the release.

I don't think any of that is 12.2 specific. Dragonbox is an algorithm for formatting floating point numbers, and fmt has used it for a long time.

> My end goal is to add `slang-server` to nixpkgs; https://search.nixos.org/packages. This way I can just integrate slang-server with the rest of my lsps, without manually having too manage builds;

Using system packages *should* work, that's supposed to be supported by the slang build, but not necessarily downgrading to earlier packages. Alternatively, it seems like it should be possible to build with local copies of everything anyway. FetchContent can disable network access with `FETCHCONTENT_FULLY_DISCONNECTED`, so if you pull local copies of everything first you can use cmake options to point to those copies and FetchContent will just work with them. Also this seems like the whole reason CMake has this [dependency provider](https://cmake.org/cmake/help/latest/guide/using-dependencies/index.html#dependency-providers) stuff, which lets you intercept the FetchContent and find_package calls, though I haven't used it personally.

evanwporter:
> > Doesn't this go against the whole idea of semver though lol. If fmt is going to break, then it should be a new major release no?
> 
> I can't control what other projects do, only my own.

Yeah I didn't mean to suggest it was on you, its just funny to me that the author of fmt is pushing breaking changes then only incrementing the version by a minor release

> FetchContent can disable network access with `FETCHCONTENT_FULLY_DISCONNECTED`, so if you pull local copies of everything first you can use cmake options to point to those copies and FetchContent will just work with them. Also this seems like the whole reason CMake has this [dependency provider](https://cmake.org/cmake/help/latest/guide/using-dependencies/index.html#dependency-providers) stuff, which lets you intercept the FetchContent and find_package calls, though I haven't used it personally.

This is way more complicated than I think I'm willing to do right now :/ . For me it would be easiest if I could just tweak the upstream a bit.

> > Doesn't this file require 12.2 though?: https://github.com/MikePopoloski/slang/blob/3998d90dc814c01b9c88a3c66b117172adc89ee5/source/util/Fmt.cpp
> > Note: i have no idea what dragonbox is, I just saw that its defined here and it was mentioned in the release.
> 
> I don't think any of that is 12.2 specific. Dragonbox is an algorithm for formatting floating point numbers, and fmt has used it for a long time.

I see. I thought it was a new feature or something that slang utilized, because the aforementioned file was created in the same commit that pinned the fmt version to 12.2

PR #1907, opened 2026-07-23
Migrate pyslang bindings from pybind11 to nanobind

jpienaar:
This is a mostly mechanical change with the exception of multiple inheritance shim, and changes related to life times with cycles being detected.

Changes:
- bindings: replace pybind11 (py::) with nanobind (nb::) across all binding translation units and pyslang.h; reformat to the repo .clang-format (4-space, 100 col).
- CMake: detect nanobind from the active Python env (falling back to FetchContent), build via nanobind_add_module, and generate the type stub with nanobind_add_stub.
- pyproject.toml: switch build-system requirement to nanobind.
- scripts/syntax_gen.py: emit nanobind-based generated syntax bindings.
- tests: construct EvalContext from the module instance (which has a parent scope) instead of the compilation root, matching the binding guard that turns a null-parent-scope deref into a clean error.
- tests: add a TODO documenting the nanobind reference-cycle leak in test_custom_systasks (Compilation <-> Python-defined subroutine).

All 111 pyslang tests pass.

My formatting on change did result in a few more changes and also I have a reasonably small python usage, so let me know if there are specific tests to look at.

Assisted-by: Gemini


pyslang/tests/test_analysis.py
```diff
@@ -196,7 +196,10 @@ def test_value_path_stringify():
     arr = root.lookupName("m.arr")
     drivers = am.getDrivers(arr)
 
-    ec = EvalContext(root)
+    # Note: EvalContext must be constructed from a symbol that has a parent
+    # scope. Passing the compilation root (which has no parent scope) triggers
+    # a null-pointer dereference (SIGILL) inside slang's EvalContext ctor.
+    ec = EvalContext(m)
```
MikePopoloski: I think the better fix for this is to make EvalContext handle a symbol without a parent scope more gracefully. If the provided symbol is itself a scope, it can use that directly. Otherwise it should throw an exception rather than crash.

pyslang/tests/test_analysis.py
```diff
@@ -455,20 +460,25 @@ def _get_proc_block(code):
 
 
 def _analyze(code):
-    """Compile *code* and return (compilation, AnalyzedProcedure) for module m."""
+    """Compile *code* and return (AnalysisManager, AnalyzedProcedure) for module m.
+
+    The returned AnalyzedProcedure is a non-owning reference into the
```
MikePopoloski: I think this is too much of a footgun for Python users, who never really want to think about lifetime stuff like this. We need a workaround for it to get the lifetimes right.

pyslang/tests/test_custom_systasks.py
```diff
@@ -24,6 +24,22 @@
 
 
 def test_custom_systasks():
+    # NOTE: This test reports nanobind instance leaks at interpreter shutdown
```
MikePopoloski: This also seems to be no good. It looks like the docs spell out how to fix cases like this: https://nanobind.readthedocs.io/en/latest/refleaks.html#fixing-reference-leaks

pyslang/tests/test_analysis.py
jpienaar: SG, switch here to analysis context.

pyslang/tests/test_analysis.py
jpienaar: Done - just Python side if I understood correctly.

pyslang/tests/test_custom_systasks.py
jpienaar: This ended up a bit more of trial than expected, getting this working for Compilation was rather direct, for the subfunction parts too, but then the closure over remained tricky. I went with the stashing approach here which does result in making these visible edges to GC (I tried to avoid that as mentioned in the commit but couldn't find a way that wasn't as ugly here - I mean easiest is to just disable leak warnings :), but that's not desirable).

bindings/python/AnalysisBindings.cpp
```diff
@@ -197,160 +195,175 @@ void registerAnalysis(py::module_& m, py::module_& ast) {
...
+        .def_prop_ro("isInSingleDriverProcedure", &ValueDriver::isInSingleDriverProcedure);
+
+    nb::class_<AnalysisManager>(m, "AnalysisManager")
+        .def(
+            "__init__",
+            [](AnalysisManager* self, AnalysisOptions options) {
+                new (self) AnalysisManager(std::move(options));
+            },
+            "options"_a = AnalysisOptions())
+        // The analyzed objects handed to listeners are passed as non-owning
```
MikePopoloski: I think this is still no good. You can see from the failing pre commit job that any basic Python linter is going to tell users to get rid of the extra object reference, leading to hard to debug crashes later on. Is there really no way to make this work with the tp_traverse / tp_clear slots as described in the docs?

jpienaar: Sorry yes, I was focusing so much on the other one that I forgot to get back to this one. Created another wrapper class for this one.

MikePopoloski:
Looks like some of the tests are still failing.

pyslang/pyslang/pyslang.cpython-310-x86_64-linux-gnu.so
MikePopoloski, line 1: This shouldn't be checked in.

pyslang/tests/conftest.py
```diff
@@ -1,7 +1,18 @@
...
 import sys
 from pathlib import Path
 
+import pytest
+
 sys.path.insert(0, str(Path(__file__).resolve().parent.parent / "examples"))
+
+
+@pytest.fixture(autouse=True)
+def _auto_gc():
```
MikePopoloski: Hmm, what's going on with this file?

pyslang/pyslang/pyslang.cpython-310-x86_64-linux-gnu.so
jpienaar: Apologies, I pushed the wrong branch (well I had rebased incorrectly locally and then pushed). Let me flip this to draft while fixing.

pyslang/tests/conftest.py
jpienaar: This was from testing and I forgot it, removed.

PR #1909, opened 2026-07-27
Instances: Split default interface port connections

AndrewNolte:
Stacked PRs:
 * #1910
 * __->__#1909
 * #1902
 * #1901

--- --- ---

[View individual changes](https://github.com/AndrewNolte/slang/commit/bcf676003e68c456e687a436d48b1ce34d93dafc)
### Instances: Split default interface port connections

Splits up behavior for instantiating default interface port connections into a separate file.
This is so it's more clear which code is used in the abnormal / slang extensions flow.
I'm planning on adding $static_assert iface param pinning in a followup that requires additional code.


MikePopoloski:
Not sure this is really necessary. It's only a hundred lines of code, and you're not adding that much more in the followup diff. The new file name is also pretty awkward.

PR #1910, opened 2026-07-27
--allow-toplevel-iface-ports: Apply $static_assert iface param overrides

AndrewNolte:
Stacked PRs:
 * #1927
 * #1926
 * #1925
 * __->__#1910

--- --- ---

[View individual changes](https://github.com/AndrewNolte/slang/commit/6c9bf2328a60d346f795253310c2d950ef3aff41)
### --allow-toplevel-iface-ports: Apply $static_assert iface param overrides

This allows users to pin iface port params, allowing for better LSP features outside of a design context.
The body will be inspected for $static_asserts, and will use those to fill in the expected parameter values for iface ports.
Related issue: https://github.com/hudson-trading/slang-server/issues/401

### DIFF
--- a/source/ast/symbols/InstanceSymbols.cpp
+++ b/source/ast/symbols/InstanceSymbols.cpp
@@ -856,8 +856,236 @@ void InstanceSymbol::resolvePortConnections() const {
     connections = conns.copy(comp);
 }
 
+struct IfaceParamAccess {
+    std::string_view portName;
+    std::string_view paramName;
+    SmallVector<const syntax::ExpressionSyntax*, 4> elementSelectors;
+};
+
+static bool parseIfacePortBase(const syntax::ExpressionSyntax& expr, std::string_view& portName,
+                               SmallVectorBase<const syntax::ExpressionSyntax*>& selectors) {
+    if (expr.kind == syntax::SyntaxKind::IdentifierName) {
+        portName = expr.as<syntax::IdentifierNameSyntax>().identifier.valueText();
+        return true;
+    }
+
+    if (expr.kind == syntax::SyntaxKind::IdentifierSelectName) {
+        auto& selected = expr.as<syntax::IdentifierSelectNameSyntax>();
+        portName = selected.identifier.valueText();
+        for (auto select : selected.selectors) {
+            if (!select->selector || select->selector->kind != syntax::SyntaxKind::BitSelect)
+                return false;
+            selectors.push_back(select->selector->as<syntax::BitSelectSyntax>().expr);
+        }
+        return true;
+    }
+
+    if (expr.kind == syntax::SyntaxKind::ElementSelectExpression) {
+        auto& selected = expr.as<syntax::ElementSelectExpressionSyntax>();
+        if (!selected.select->selector ||
+            selected.select->selector->kind != syntax::SyntaxKind::BitSelect) {
+            return false;
+        }
+
+        if (!parseIfacePortBase(*selected.left, portName, selectors))
+            return false;
+        selectors.push_back(selected.select->selector->as<syntax::BitSelectSyntax>().expr);
+        return true;
+    }
+
+    return false;
+}
+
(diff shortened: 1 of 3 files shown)


source/ast/symbols/ParameterBuilder.h
```diff
@@ -30,6 +34,21 @@ class ParameterBuilder {
     bool hasErrors() const { return anyErrors; }
 
     void setAssignments(const syntax::ParameterValueAssignmentSyntax& syntax, bool isFromConfig);
+
+    /// Adds a pre-resolved override for a value parameter. Unlike a syntactic assignment, the
+    /// value is already evaluated, so no instance context is needed and it is applied directly.
+    /// Useful for propagating an already-elaborated value (e.g. copied from another instance, or
+    /// derived from a constraint) onto the parameter of a freshly built instance.
+    void addValueOverride(std::string_view name, ConstantValue value) {
+        resolvedOverrides.emplace(name, std::move(value));
```
MikePopoloski: These are stored as a variant and the caller already has a variant, so you may as well have one method that takes a variant and not force the caller to unwrap it.

source/ast/symbols/ParameterBuilder.cpp
```diff
@@ -177,6 +177,18 @@ const ParameterSymbolBase& ParameterBuilder::createParam(
         }
 
         auto& tt = param->targetType;
+
+        // A pre-resolved type override takes precedence and is applied directly, since it has
+        // already been resolved and doesn't depend on any instance context. Localparams can't be
+        // overridden, so fall through to normal resolution for those.
+        if (auto it = resolvedOverrides.find(decl.name);
```
MikePopoloski: I don't love adding another map lookup for every param we create, when this will almost never be used. Can this extra value be stored alongside other stuff in the `assignments` map so we get it as part of the main lookup?

source/ast/symbols/InstanceSymbols.cpp
```diff
@@ -900,19 +1073,33 @@ void InstanceSymbol::connectDefaultIfacePorts() const {
...
 
     SmallVector<const PortConnection*> conns;
     for (auto port : body.getPortList()) {
         if (port->kind == SymbolKind::InterfacePort) {
             auto& ifacePort = port->as<InterfacePortSymbol>();
             if (ifacePort.interfaceDef) {
+                SmallVector<IfaceParamConstraint> constraints;
+                SmallVector<IfaceParamOverride> paramOverrides;
+                if (bodySyntax) {
+                    collectIfaceParamConstraints(ifacePort, *bodySyntax, constraints);
```
MikePopoloski: These two methods are only ever called back to back, so you can just make them one method instead, and avoid exposing the intermediary IfaceParamConstraint type.

source/ast/symbols/InstanceSymbols.cpp
```diff
@@ -856,7 +856,170 @@ void InstanceSymbol::resolvePortConnections() const {
...
+// Only asserts written directly in the module body are considered. Asserts nested in generate
+// blocks are skipped: their constraints are conditional, but the override has to be applied
+// before elaboration (so before we know which branches are taken), and pinning a parameter from
+// a possibly-untaken branch would be wrong.
+static void collectIfaceParamConstraints(const InterfacePortSymbol& port,
+                                         const syntax::ModuleDeclarationSyntax& topBody,
+                                         SmallVectorBase<IfaceParamConstraint>& constraints) {
+    auto& def = *port.interfaceDef;
+
+    for (auto member : topBody.members) {
```
MikePopoloski: For every iface port we loop over every body member. Probably this should be one loop to find all static asserts and then reusing them for each iface port.

source/ast/symbols/ParameterBuilder.h
AndrewNolte: I had added these as values/types for a focused-instance selection feature, but I'll switch to one method here. I think we should be able to do that feature with cloning/recreating syntax nodes.

PR #1921, opened 2026-08-03
Bump mimalloc to 3.4.4

AndrewNolte:
This picks up the fix for this mimalloc issue: https://github.com/microsoft/mimalloc/issues/1341

Related slang-server issue: https://github.com/hudson-trading/slang-server/issues/434


MikePopoloski:
Seems like this update breaks the noexcept build. Sigh.

PR #1925, opened 2026-08-05
Diagnostics: Put notes under associated diag for clients and json output

AndrewNolte:
Stacked PRs:
 * #1927
 * #1926
 * __->__#1925

--- --- ---

[View individual changes](https://github.com/AndrewNolte/slang/commit/897322eeeaff84eb3fc88729f9216625dae5be9a)
### Diagnostics: Put notes under associated diag for clients and json output

This puts some more restrictions on notes, many that were already being followed:
- Notes cannot have subnotes
- Notes cannot be waived, only diags
- Note severities can only be remapped to ignored, not errors

### DIFF
--- a/source/diagnostics/DiagnosticEngine.cpp
+++ b/source/diagnostics/DiagnosticEngine.cpp
@@ -213,13 +213,44 @@ void DiagnosticEngine::issue(const Diagnostic& diagnostic) {
                          severity == DiagnosticSeverity::Fatal;
     if (isError && errorLimit && numErrors >= errorLimit) {
         Diagnostic diag(diag::TooManyErrors, SourceLocation::NoLocation);
-        issueImpl(diag, DiagnosticSeverity::Fatal);
+        ReportedDiagnostic report(diag, DiagnosticSeverity::Fatal);
+        report.location = diag.location;
+        report.formattedMessage = formatMessage(diag);
+        for (auto& client : clients)
+            client->report(report);
         issuedOverLimitErr = true;
         return;
     }
 
-    if (!issueImpl(diagnostic, severity))
+    auto info = getReportedDiag(diagnostic);
+    if (!info)
         return;
+
+    ReportedDiagnostic report(std::move(*info), severity);
+
+    for (const Diagnostic& note : diagnostic.notes) {
+        // Notes shouldn't have subnotes. We should change this shape too eventually, unless we
+        // want to support that.
+        SLANG_ASSERT(note.notes.empty());
+
+        // Notes are ignored if location is "NoLocation" since they frequently make no
+        // sense without location information.
+        if (note.location == SourceLocation::NoLocation && !note.code.showNoteWithNoLocation())
+            continue;
+
+        auto noteSeverity = getSeverity(note.code, note.location);
+        if (noteSeverity == DiagnosticSeverity::Ignored)
+            continue;
+
+        SLANG_ASSERT(noteSeverity == DiagnosticSeverity::Note);
+
+        auto noteInfo = getReportedDiag(note);
+        SLANG_ASSERT(noteInfo);
+        report.notes.emplace_back(std::move(*noteInfo));
+    }
+
+    for (auto& client : clients)
+        client->report(report);
 
     if (severity == DiagnosticSeverity::Warning)
         numWarnings++;
@@ -232,11 +263,11 @@ void DiagnosticEngine::issue(const Diagnostics& diagnostics) {
         issue(diag);
 }
 
-bool DiagnosticEngine::issueImpl(const Diagnostic& diagnostic, DiagnosticSeverity severity) {
+std::optional<ReportedDiagnosticInfo> DiagnosticEngine::getReportedDiag(
+    const Diagnostic& diagnostic) {
     // Walk out until we find a location for this diagnostic that isn't inside a macro.
     SmallVector<SourceLocation, 8> expansionLocs;
     SourceLocation loc = diagnostic.location;
-    size_t ignoreExpansionsUntil = 0;
     bool showIncludeStack = false;
 
     if (loc != SourceLocation::NoLocation) {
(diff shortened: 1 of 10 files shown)


MikePopoloski:
The inheritance model here is weird and a bit clunky. How about instead of trying to render the notes into the top level diagnostic, the report() method takes the original Diagnostic& and then it's up to the derived class to decide what to do with the original notes? I think there's no reason to try to store them all in the ReportedDiagnostic, the derived class can just call back into the engine for each note to render it individually.

source/diagnostics/DiagnosticEngine.cpp
```diff
@@ -213,13 +213,43 @@ void DiagnosticEngine::issue(const Diagnostic& diagnostic) {
...
+    ReportedDiagnostic report(std::move(*info), severity);
+
+    for (const Diagnostic& note : diagnostic.notes) {
+        // Notes shouldn't have subnotes. We should change this shape too eventually, unless we
+        // want to support that.
+        SLANG_ASSERT(note.notes.empty());
+
+        // Notes are ignored if location is "NoLocation" since they frequently make no
+        // sense without location information.
+        if (note.location == SourceLocation::NoLocation && !note.code.showNoteWithNoLocation())
```
MikePopoloski, line 238: This logic could go into getReportedDiag

source/diagnostics/DiagnosticEngine.cpp
```diff
@@ -256,12 +287,22 @@ bool DiagnosticEngine::issueImpl(const Diagnostic& diagnostic, DiagnosticSeverit
         }
 
         showIncludeStack = reportedIncludeStack.emplace(loc.buffer()).second;
+    }
+
+    ReportedDiagnosticInfo report(diagnostic);
```
MikePopoloski: Why did this get moved around?

source/diagnostics/JsonDiagnosticClient.cpp
```diff
@@ -34,6 +54,33 @@ void JsonDiagnosticClient::report(const ReportedDiagnostic& diag) {
                                       sourceManager->getLineNumber(loc), getColumnNumber(loc)));
     }
 
+    SmallVector<SourceRange> mappedRanges;
```
MikePopoloski: So this is unrelated to the stated purpose of the diff right? Like we're adding new output to the JSON diagnostics.

source/diagnostics/JsonDiagnosticClient.cpp
```diff
@@ -81,7 +128,6 @@ void JsonDiagnosticClient::report(const ReportedDiagnostic& diag) {
         }
         writer.endArray();
     }
-    writer.endObject();
```
MikePopoloski, line 84: Huh, was this always broken?

tests/unittests/parsing/DiagnosticTests.cpp
```diff
@@ -577,8 +642,16 @@ endmodule
 
     auto client = std::make_shared<TextDiagnosticClient>();
     engine.addClient(client);
-    for (auto& diag : compilation.getAllDiagnostics())
-        engine.issue(diag);
+    for (auto& diag : compilation.getAllDiagnostics()) {
+        if (diag.code == diag::UndeclaredIdentifier) {
```
MikePopoloski: What's going on here?

tests/unittests/parsing/DiagnosticTests.cpp
```diff
@@ -657,8 +730,16 @@ endmodule
 
     auto client = std::make_shared<JsonDiagnosticClient>(writer);
     engine.addClient(client);
-    for (auto& diag : compilation.getAllDiagnostics())
-        engine.issue(diag);
+    for (auto& diagnostic : compilation.getAllDiagnostics()) {
```
MikePopoloski: Same?

source/diagnostics/JsonDiagnosticClient.cpp
AndrewNolte: This is for the next pr- https://github.com/MikePopoloski/slang/pull/1926

AndrewNolte: The motivation for all of this was for the lit diags- like I wanted the json to reflect the diag->notes relationship, which it currently did not. This bit exposes the diag ranges

tests/unittests/parsing/DiagnosticTests.cpp
AndrewNolte: Sorry didn't look too closely at these test edits - removing. This one was pointless

tests/unittests/parsing/DiagnosticTests.cpp
AndrewNolte: Yeah I need to instead change the sv to make a diag with notes show up

source/diagnostics/DiagnosticEngine.cpp
AndrewNolte: I can try and make the structure match the previous one a bit more

source/diagnostics/DiagnosticEngine.cpp
AndrewNolte: I think it's better here bc we don't have to re-check that they are notes?

source/diagnostics/JsonDiagnosticClient.cpp
AndrewNolte: I can put this on the next diff

source/diagnostics/JsonDiagnosticClient.cpp
AndrewNolte: No this function was split into two, so the endObject() stayed in the calling function, on line 35.

source/diagnostics/DiagnosticEngine.cpp
MikePopoloski: getReportedDiag is already checking the severity, and already has a branch for notes, so it may as well also check the location rule, to keep all the filtering logic in one place.

MikePopoloski (commented):
Looks like tests are still failing btw.

PR #1927, opened 2026-08-05
Structured Assigns: better diags

AndrewNolte:
[View individual changes](https://github.com/AndrewNolte/slang/commit/e477b641c300575cf6978312122caac30b1c7b36)
- No more generic 'missing fields' - elaborate which fields
- For arrays, specify the type of the array so the user can see which indexes to add, or which type to add
- Highlight only first part of struct assign so that the diag doesn't obscure more important diags like syntax errors.


MikePopoloski:
Hmm, looks like this one does not rebase cleanly either.

AndrewNolte:
I'll be using github stacks for future stacked PRs, which should avoid failed rebases like this.

PR #1928, opened 2026-08-06
Preprocessor: Add synthetic token for trailing disabled branch trivia

AndrewNolte:
[View individual changes](https://github.com/AndrewNolte/slang/commit/ae9d3c53222db3e66f4c26b2ce47568e8374fa62)

This is for the purpose of putting all the disabled content back under disabledTokens, which is generally easier to work with.

While testing it was noticed that cst json round trip was brokekn when ifdefs were present.
The 'fromExpansion' attribute was added to the cst json to address this, and additional round trip tests were added.

Also while testing, cmake was found to accept older python versions that won't work for the bindings; 3.11 is set as the floor as specified in bindings/CMakeLists.txt

### DIFF
--- a/bindings/python/SyntaxBindings.cpp
+++ b/bindings/python/SyntaxBindings.cpp
@@ -392,15 +392,16 @@ void registerSyntax(py::module_& syntax, py::module_& parsing) {
         .def("__str__", &SyntaxNode::toString)
         .def(
             "to_json",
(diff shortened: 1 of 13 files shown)


AndrewNolte:
Going to clean this up a bit more

MikePopoloski:
This needs to be fixed up after the change to use nanobind for bindings. Otherwise looks ok to me.

MikePopoloski:
Bump

MikePopoloski:
Looks like another rebase needed.

PR #1933, opened 2026-08-15
CMake: Check for system `mimalloc` before fetching

evanwporter:
When `SLANG_USE_MIMALLOC` is true, this PR first checks for a system mimalloc, and if one is not found then it fetches it from github as before.

I also considered adding a `SLANG_USE_SYSTEM_MIMALLOC` but I thought that might create a confusion about which flags should be on, and whether they are conflicting, etc, etc.

An alternative is to set SLANG_USE_SYSTEM_MIMALLOC to have three possible values, `TRUE`, `FALSE` and `SYSTEM`. But again this might be more complicated than its worth.

### DIFF
--- a/external/CMakeLists.txt
+++ b/external/CMakeLists.txt
@@ -96,26 +96,32 @@ if(SLANG_USE_MIMALLOC)
       FATAL_ERROR "mimalloc cannot be used with Windows shared lib builds")
   endif()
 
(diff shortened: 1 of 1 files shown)


evanwporter:
just realized this is redundant due to find package args :facepalm: 

sorry for creating pr spam

PR #1949, opened 2026-09-03
Fix instance cache key ignoring arrayed interface port connections

### LINKED ISSUE #1947: Arrayed interface port loses errors depending on instantiation order

C++ slang, master at `cbd5d62d8` (11.0.470).

**Describe the bug**

A design with an obvious error compiles clean, and whether the error shows up at all depends on which order two instances are written in.

It takes a module whose interface port carries a range, instantiated twice against arrays of different interfaces. Only the one written first gets checked; errors in the other one are just gone. The same design with a scalar interface port is fine.

**To Reproduce**

`NoAck` has no `ack`, and `Reader` reads `a[0].ack`:

```sv
interface HasAck;
  logic req;
  logic ack;
endinterface

interface NoAck;
  logic req;
endinterface

module Reader (interface a[2]);
  initial $display("%b", a[0].ack);
endmodule

module top;
  HasAck has_ack [2] ();
  NoAck  no_ack [2] ();

  Reader on_has_ack (.a(has_ack));
  Reader on_no_ack  (.a(no_ack));
endmodule
```

```
$ slang --quiet repro.sv
$ echo $?
0
```

Swap the two instantiation lines and it shows up:

```
$ slang --quiet repro.sv
  in instance: top.on_no_ack
repro.sv:11:30: error: could not resolve hierarchical path name 'ack'
  initial $display("%b", a[0].ack);
                             ^~~~
$ echo $?
1
```

It is the range that matters rather than the generic port: a named `Bus a[2]` connected to two parameterizations of `Bus` does the same thing, and both scalar forms report correctly.

**A fix**

The two instances end up sharing a body, which `--ast-json` shows as the second one's body being a link rather than its own object.

`InstanceCacheKey` descends an arrayed interface connection to `elements[0]`, and the check on the way is `arr.empty()`, which is `Scope::empty()` -- whether the symbol has scope members, not whether it has elements. The array built for an interface port connection has elements and no members, so the descent bails out to null, the interface never enters the key, and both instances hash to just the definition pointer. Everywhere else that tests an instance array for emptiness uses `array.elements.empty()`; `InstanceCacheKey.cpp:49` is the only `arr.empty()`.

I'll open a PR for this.

**Additional context**

The dropped diagnostic is the visible half. Anything reading `getCanonicalBody()` as "these two instances compile alike", which is what it is there for, also gets one body where two are needed -- which is how I ran into it.

hankhsu1996:
## Summary

Fixes #1947. Two instances of a module whose interface port carries a range were cached together even when connected to arrays of different interfaces, so the second one's body was never visited and errors in it were dropped. Which instance got checked depended on the order they were written in. The issue has the full background; this PR is the one-line fix plus a test.

## Design

- `InstanceCacheKey` guarded its descent with `arr.empty()`, which resolves to `Scope::empty()` -- whether the symbol has members, not whether it has elements. The array a port connection carries comes from `rewireIfaceArrayIndices`, which fills `elements` and adds no members, since that array is only handed to `PortConnection` and never placed in a scope for anything to traverse. So the guard fired, the descent bailed out to null, the interface contributed nothing to the key, and both instances hashed to just the definition pointer and compared equal.
- Every other place that asks this question already tests `elements`. `Lookup.cpp` (`selectChild`, `selectChildRange`), `Expression.cpp` and `MiscStatements.cpp` all use `array.elements.empty()`, and this was the only `arr.empty()`.

## Testing

New test in `InterfaceTests.cpp`, with one module taking an arrayed generic interface port and instantiated against two interfaces where only one declares the member the module reads. The error is reported with the offending instance written second, where before the change that case produced no diagnostics at all. The full suite passes.

## AI disclosure

### DIFF
--- a/source/ast/InstanceCacheKey.cpp
+++ b/source/ast/InstanceCacheKey.cpp
@@ -38,7 +38,7 @@ InstanceCacheKey::InstanceCacheKey(const InstanceSymbol& symbol, bool& valid,
             auto [iface, modport] = conn->getIfaceConn();
             while (iface && iface->kind == SymbolKind::InstanceArray) {
                 auto& arr = iface->as<InstanceArraySymbol>();
-                if (arr.empty())
+                if (arr.elements.empty())
                     iface = nullptr;
                 else
                     iface = arr.elements[0];
--- a/tests/unittests/ast/InterfaceTests.cpp
+++ b/tests/unittests/ast/InterfaceTests.cpp
@@ -1326,3 +1326,37 @@ endclass
     compilation.addSyntaxTree(tree);
     NO_COMPILATION_ERRORS;
 }
+
+TEST_CASE("Arrayed interface port cache regress -- GH #1947") {
+    // NoAck has no 'ack', so on_no_ack is in error. It is written second because that is
+    // the instance a shared body would hide.
+    auto tree = SyntaxTree::fromText(R"(
+interface HasAck;
+    logic req;
+    logic ack;
+endinterface
+
+interface NoAck;
+    logic req;
+endinterface
+
+module Reader (interface a[2]);
+    initial $display("%b", a[0].ack);
+endmodule
+
+module top;
+    HasAck has_ack [2] ();
+    NoAck no_ack [2] ();
+
+    Reader on_has_ack (.a(has_ack));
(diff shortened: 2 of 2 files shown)


MikePopoloski:
Nice, thanks for finding and fixing.

Issue #1953, opened 2026-09-04
PySlang Stub Generation


allRisc:
**Is your bug report related to the C++ slang project or the pyslang Python bindings?**
Pyslang Python bindings

**Describe the bug**
The PySlang stubs generated by nanobind do no include the child modules of the pyslang ('pyslang.ast', 'pyslang.syntax', etc.). This leaves the current type stubs broken for most real work.

**To Reproduce**
```bash
git clone https://github.com/allrisc/slang
cd slang
git  checkout 546cce943

python3 -m venv .venv
source .venv/bin/activate

pip install -e .
pip install pyright pytest

pytest ./pyslang/tests/test_submodule_stubs.py
```

**AI Disclosure**
AI was used to write the above test, however the problem was found and diagnosed by hand.

allRisc:
Rebased off the latest master which includes worktree fixes and a switch to nanobind split-mode, so the demo commit hash is now: 546cce943

allRisc:
Following the initial fix of this I have discovered the following additional issues from the new comprehensive testing suite:

**Syntax Errors**
- [x] Some methods use Python keywords (such as `with`) as argument names
    - 169c70125 
- [x] Some methods (especially in SyntaxFactory) have default arguments before required arguments
    - 3ec13d5cb
- [x] Some enum members have a '$' at the start. (need feedback on this)
    - 1d837cc07
- [x] Some enums still have 'None' (DriverFlags and AnalysisFlags)
    - 654f1598d

**Proposals**
- Methods which use Python keywords have '_' appended to the argument name. This is consistent with other Keyword collisions such as the Enums
- The generation of these methods should be changes so that any optional arguments which have required arguments after have their '= None' dropped from them and are required, but nullable.
- Enum members which start with '$' should be changed to start with 's_' and if they have '$' within the enum then they should be directly replaced with '_' (so `$test$plusargs` would become `s_test_plusargs)
- Update the enums which still have None to use `None_` instead

allRisc:
@MikePopoloski are they any other invested contributors who should be looped in to the above discussion?

MikePopoloski:
Thanks for investigating. Plenty of people use the pyslang bindings but I don't know of any who would be particularly invested in this change. For example, I doubt anyone will notice or care about the breaking changes introduced here, but I'm sure many will be thankful for the fixed stubs.

PR #1956, opened 2026-09-08
Bugfix/pyslang stubs

allRisc:
This PR addresses and closes the issues presented in #1953 

**Summary**
The generated stubs for Pyslang have the following key deficiencies:
- Does not generate type-hints for child modules
- Does not properly type hint the modules imported into the top-level pyslang file (i.e. `import pyslang.ast` causes all underlying `pyslang.ast.Member` to be typed as 'Unknown')
- The generated stubs do not produce valid Python AST:
    - Several method use reserved python keywords (such as `with`)
    - Methods in `SyntaxFactory` have default arguments before required arguments
    - `DriverFlags` and `Analysis` flags have direct `None` member (caused by the need to bypass the enum macro which normally sanitizes there values

All of the above deficiencies are addressed in the following ways respectively:
- Use the nanobind recursive stub generation
- Duplicate the generated stubs into both locations during install (directly under `pyslang/` and `pyslang/pyslang/`)
- Fixed Generation issues present
    - Added keyword checking and sanitization (via a trailing underscore) to method generation
    - Dropped the `= None` from the typing of method arguments which have required arguments after them (this still permits None to be directly passed)
    - Updated the direct bindings of the `DriverFlags` and `AnalysisFlags`

**Verification**
A new set of tests were added dedicated to the stub generation:
- Check that the module stubs are added in the correct location
- Check that the generated stubs can be parsed into valid Python AST
- _OPTIONAL:_ If pyright is installed will check that pyright can correctly resolve the typing from the two import locations
- _OPTIONAL:_ If mypy is installed will check that mypy can correctly resolve the typing from the two import locations

**AI Disclosure**
AI was used to accelerate debugging the initial issues presented in this PR, review changes prior to each commit, and create simple tests (which were then expanded) to validate the issues and resolutions. AI was _not_ used to generate solutions to any of the issues.


bindings/python/pyslang.h
```diff
@@ -48,19 +49,50 @@ using namespace slang::parsing;
...
+    "class", "continue", "def",   "del",    "elif", "else",   "except", "finally",  "for",
+    "from",  "global",   "if",    "import", "in",   "is",     "lambda", "nonlocal", "not",
+    "or",    "pass",     "raise", "return", "try",  "while",  "with",   "yield",
+};
+
+// For each name in the enum we need to sanatize it.
+// For simple Python Keywords we need to simply add '_' to the end.
+// For names which start with '$' prepend 's_' inplace of '$'.
+// For names which contain '$' replace '$' with '_'.
+inline std::string sanatize_name(const std::string& name) {
```
MikePopoloski, line 66: "sanitize", and camelCase: "sanitizeName"

MikePopoloski (commented):
Looks like there is a failing pyslang test to fix.

allRisc:
My apologies. It appears I missed updating a different test that was impacted by the changes. I will have a fix tonight.

allRisc:
I have updated the test to correspond with the updated behavior. I also believe I have fixed the Windows PySlang build problem that was present prior to this PR.

PR #1966, opened 2026-09-18
Adding the --ignore-untaken-generate-refs option for #1602

### LINKED ISSUE #1602: undriven-port warning is not flagged when the signal is driven in one of the if condition
C++ slang

**Describe the bug**
In below case, when COND_A parameter is assigned to 0, output "a" is not assigned to any value. But slang is able to build successfully without any warning. 
This behaviour is consistent on slang version 9.1.194+343611be4 (from slang explorer).
It will happen when one of the if condition has output "a" assigned with value.

**To Reproduce**
Run with -Weverything

```systemverilog
module testing #(
    parameter COND_A = 0
) (
    output logic a
);

generate
    if (COND_A) begin : cond_a
        assign a = 0;
    end
    else begin : default_case

    end
endgenerate

endmodule
```

AlexThePotter:
# Add `--ignore-untaken-generate-refs` for unused / undriven checking

Developed with AI, but I have reviewed changes and looks reasonable. Works well in my testing with no observed performance cost. Happy to change things around if there are any preferred styles etc

Closes #1602

## Summary

Adds an opt-in compilation flag, `CompilationFlags::IgnoreUntakenGenerateRefs`
(CLI: `--ignore-untaken-generate-refs`), that makes the unused / undriven warnings
(`-Wundriven-port`, `-Wunused-net`, `-Wunused-variable`, etc.) ignore references that
occur inside untaken generate branches.

Default behaviour is unchanged: references from untaken branches still count, so the
warnings continue to mean "never used under any parameterization". This follows the
suggestion in #1602 to add a flag controlling whether untaken generate blocks are
examined for the purposes of these warnings.

## Motivation

```systemverilog
module m #(parameter COND_A = 0) (output logic a);
  if (COND_A) begin : cond_a
    assign a = 0;   // only driver, in the untaken branch
  end
endmodule
```

With `COND_A = 0` nothing drives `a`, but slang is silent under `-Weverything` because
the reference inside `cond_a` is recorded like any other. Users who override parameters
to match their configuration want the warnings to reflect the design as actually
elaborated; users linting shared parameterized IP want the current behaviour. The flag
lets each pick.

## Design

References are tracked in `Compilation::referenceStatusMap`, keyed on the declaration's
syntax node and populated during binding via `ASTContext::noteReference`. Untaken
generate blocks are deliberately created and bound (their diagnostics are filtered in
`Compilation::addDiag`, but their `noteReference` calls are not).

The change gates the write, not the read: when the flag is set and the referencing
`ASTContext::scope->isUninstantiated()`, the reference is not recorded. Two sites:

- `ASTContext::noteReference` (all net / variable / port references, including port
  connection expressions)
- `ArbitrarySymbolExpression::fromSyntax`, which calls `Compilation::noteReference`
  directly

This has to be a `CompilationFlags` flag rather than an `AnalysisFlags` flag because the
map is fully populated before `AnalysisManager` runs; the analysis side cannot tell after
the fact which bits came from untaken code.

`Scope::isUninstantiated()` also returns true for uninstantiated instance bodies and in
`LintMode`. Both are harmless here: symbols declared in uninstantiated scopes are already
excluded from unused checking by `shouldWarnUnused`, and in lint mode nothing is ever
warned about. The check is ordered so the default path only pays a single `hasFlag` test.

Because the map is keyed on the declaration syntax, a branch that is taken in *any*
instance of a module still counts for all instances of that module. This is documented in
the command-line reference and covered by a test.

## Changes

- `include/slang/ast/Compilation.h`: new `IgnoreUntakenGenerateRefs` flag; `SLANG_BITMASK`
  last-value updated
- `source/ast/ASTContext.cpp`, `source/ast/expressions/MiscExpressions.cpp`: gate the
  reference writes
- `source/driver/Driver.cpp`: `--ignore-untaken-generate-refs` in the Diagnostic Control
  group
- `bindings/python/CompBindings.cpp`: expose the enum value
- `docs/command-line-ref.dox`: flag documentation with example
- `CHANGELOG.md`: entry under New Features
- `tests/unittests/analysis/UnusedTests.cpp`: new test covering if / case / loop
  generate, an unused local variable in an untaken block, the shared-syntax two-instance
  case, and a negative control asserting no diagnostics without the flag
- `tests/unittests/DriverTests.cpp`: CLI plumbing test

## Testing

- Full `unittests` suite passes, including the existing
  `"Unused nets and vars false positives regress"` test that locks in the default.
- Reproducer from #1602: clean by default, `undriven-port 'a'` with the flag.
- A real parameterized AHB stub module with nine outputs driven only in an untaken
  `case` arm: clean by default, nine `undriven-port` warnings with the flag.

## Notes

- While testing I noticed that port connections of *instances* inside untaken generate
  blocks are already not recorded as references today (e.g. `sub s(.o(c[i]))` inside an
  untaken `if` leaves `c` reported as undriven even without the flag), whereas plain
  `assign` statements in the same position are recorded. This PR does not change that;
  mentioning it in case it is unintended.
- The bit-level case raised later in #1602 (`a[0]` undriven while `a[3:1]` are driven
  in taken loop iterations) is not addressed. `checkValueUnused` operates on whole
  symbols; per-bit reporting would need `DriverTracker` data and is a separate feature.
- References to typedefs, imports and parameters that go through
  `Compilation::noteReference(const Symbol&)` directly from `Lookup.cpp` (without an
  `ASTContext`) are not gated. The flag targets nets, variables and ports, which is what
  the issue is about. Easy to extend if desired.


MikePopoloski:
As you mention in your diff, this will still mark a variable used if any instance uses it, which I think does not address the intent of the issue request, because they are looking to find undriven output ports to catch actual wires / connections with no value driven. The approach to handling that correctly is completely different from the current unused symbol system, and this PR doesn't really get us closer to it, though it does find a subset of possible issues. I'm not sure it's worth the complexity of adding this new flag when we'll likely need a new system that would supersede it anyway.

